PR #118, opened 2022-04-18
Add optional `aria-label` prop to `DropdownInput`.

tmeyer2115:
This prop allows users of `DropdownInput` to optionally add an `aria-label`
to the `input` element. This is especially useful when using the Component
with buttons such as the clear button or the `DropdownSubmitButton`.

J=SLAP-2058
TEST=auto

Verified that Storybook no longer reported WCAG violations for the `SearchBar`.

### DIFF
--- a/src/components/Dropdown/DropdownInput.tsx
+++ b/src/components/Dropdown/DropdownInput.tsx
@@ -10,6 +10,7 @@ import { useInputContext } from './InputContext';
 export function DropdownInput(props: {
   className?: string,
   placeholder?: string,
+  ariaLabel?: string,
   onSubmit?: (value: string, index: number, focusedItemData: FocusedItemData | undefined ) => void,
   onFocus?: (value: string) => void,
   onChange?: (value: string) => void,
@@ -18,6 +19,7 @@ export function DropdownInput(props: {
   const {
     className,
     placeholder,
+    ariaLabel,
     onSubmit,
     onFocus,
     onChange,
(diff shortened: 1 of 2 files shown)


src/components/Dropdown/DropdownInput.tsx
```diff
@@ -79,6 +81,7 @@ export function DropdownInput(props: {
       id={screenReaderUUID && generateDropdownId(screenReaderUUID, -1)}
       aria-describedby={screenReaderUUID}
       aria-activedescendant={screenReaderUUID && generateDropdownId(screenReaderUUID, focusedIndex)}
+      aria-label={ariaLabel}
```
nmanu1, line 84: Is a label always required on the input for WCAG purposes? If so, we should probably use a default value if `ariaLabel` isn't specified

src/components/SearchBar.tsx
```diff
@@ -280,6 +280,7 @@ export function SearchBar({
         onSubmit={handleSubmit}
         onFocus={handleInputFocus}
         onChange={handleInputChange}
+        ariaLabel={'Conduct a search'}
```
oshi97, line 283: extra braces!

src/components/Dropdown/DropdownInput.tsx
tmeyer2115: No, it's not. An `aria-label` is only required if the `input` is labeled visually (i.e. with a search icon).

src/components/SearchBar.tsx
tmeyer2115: D'oh. Fixed!

oshi97 (approved):

PR #119, opened 2022-04-18
Perform a single CSS composition

nmanu1:
This PR ensures that components are only performing a single CSS composition. A child component either relies solely on the parent components to perform the composition and now accepts only `cssClasses`, or if it might be used without a parent, it performs it's own composition, but parent components pass in 'replace' for the `cssCompositionMethod`.

J=SLAP-2011
TEST=manual

Spin up the test-site and click through to make sure the styling is still applied as expected.


src/components/DirectAnswer.tsx
```diff
@@ -39,11 +40,7 @@ export interface DirectAnswerCssClasses extends ThumbsFeedbackCssClasses {
   viewDetailsLink?: string,
   viewDetailsLinkContainer?: string,
   highlighted?: string,
-  answerContainer?: string,
-  feedbackButtonsContainer?: string,
-  feedbackButton?: string,
-  thumbsUpIcon?: string,
-  thumbsDownIcon?: string
```
cea2aj, line 46: I believe these were here because its a little nicer for the docs to display the full list of classes rather than having to click through to see the interface which is being extended. However 'feedbackButton' doesn't look like it's being used anywhere so we can remove that one regardless

src/components/UniversalResults.tsx
```diff
@@ -10,12 +14,13 @@ import { VerticalConfigMap } from '../models/verticalConfig';
...
  * @public
  */
-export interface UniversalResultsCssClasses {
+export interface UniversalResultsCssClasses extends SectionHeaderCssClasses {
   container?: string,
   results___loading?: string
 }
 
 const builtInCssClasses: UniversalResultsCssClasses = {
+  ...sectionHeaderCssClasses,
```
cea2aj, line 23: Can we be consistent about whether we spread the css classes or specify them individually?

src/components/sections/SectionHeader.tsx
```diff
@@ -1,15 +1,18 @@
...
 import { AppliedFiltersDisplay } from '../AppliedFiltersDisplay';
-import { useComposedCssClasses, CompositionMethod } from '../../hooks/useComposedCssClasses';
 import { CollectionIcon } from '../../icons/CollectionIcon';
 import { AppliedQueryFilter, SelectableFilter as DisplayableFilter, useAnswersState } from '@yext/answers-headless-react';
 import classNames from 'classnames';
 import { useAnalytics } from '../../hooks/useAnalytics';
 import { VerticalLink } from '../../models/verticalLink';
 import { useCallback } from 'react';
 
+const DEFAULT_CSS_CLASSES: SectionHeaderCssClasses = {};
```
cea2aj, line 10: What do you think of "CSS_CLASSES_FALLBACK"?  since 'default' is similar in meaning to 'builtIn'

src/components/sections/StandardSection.tsx
```diff
@@ -2,13 +2,14 @@ import { VerticalResultsDisplay } from '../VerticalResultsDisplay';
...
 
 /**
  * The CSS class interface used for {@link StandardSection}.
  *
  * @public
  */
-export interface StandardSectionCssClasses {
+export interface StandardSectionCssClasses extends VerticalResultsCssClasses {
   section?: string
 }
```
cea2aj, line 14: Should we define `results___loading` in the StandardSection builtInCssClasses?

nmanu1: 50% opacity while loading universal results is already set in the `UniversalResults` component, which is why it doesn't need to be defined again with `results___loading` in the `StandardSection`.

I think having it at the higher-level component makes a little more sense, but, I don't have a strong preference. if it's confusing, I can remove the one from `UniversalResults` instead and define it in `StandardSection`

src/components/DirectAnswer.tsx
nmanu1: I see. I think it might still be better to remove these for consistency though, because the other interfaces don't include the full list of classes when they extend other classes.
If we want to always display the full list of classes, that might be better to do in a separate item

src/components/UniversalResults.tsx
nmanu1: Sure, I updated to specify them individually

src/components/sections/StandardSection.tsx
cea2aj: Got it. In that case, should `results` or `results___loading` be in the VerticalResultsCssClasses then? I'm not seeing where they are being used

nmanu1: Yes, both `results` and `results___loading` are part of `VerticalResultsCssClasses`.

cea2aj: I see. I didn't notice that VerticalResults passes those classes to VerticalResultsDisplay through the `otherProps` variable

cea2aj (approved):

PR #124, opened 2022-04-21
Refactor the `StandardCard` and add `PropTypes`.

tmeyer2115:
This PR splits the `StandardCard` into two Components: `StandardCard` and
`StandardCardDisplay`. This was done to enhance readability and organization.
There was a lot going on in the original `StandardCard`, making the file hard
to parse. We want readability so that users of the library easily know how to
make their own cards.

I removed our own `validateData` run-time type checking from the card. Instead,
I used the native `React.PropTypes` in the display Component. This will be more
natural for React developers. `PropTypes` is the idiomatic way to do type
checking at run-time in React.

TEST=auto

Ensured all the unit tests passed.


src/components/cards/standard/StandardCardDisplay.tsx
```diff
@@ -0,0 +1,176 @@
...
+  container: 'flex flex-col justify-between border rounded-lg mb-4 p-4 shadow-sm',
+  header: 'flex text-neutral-dark',
+  body: 'flex justify-end pt-2.5 text-base',
+  cta1: 'whitespace-nowrap bg-primary text-white font-medium rounded-lg py-2 px-5 shadow',
+  cta2: 'whitespace-nowrap bg-white text-primary font-medium rounded-lg py-2 px-5 mt-2 shadow',
+  title: 'text-lg font-medium',
+  feedbackButtonsContainer: 'flex justify-end mt-4 text-sm text-gray-400 font-medium',
+  thumbsUpIcon: thumbsFeedbackCssClasses.thumbsUpIcon,
+  thumbsDownIcon: thumbsFeedbackCssClasses.thumbsDownIcon
+};
```
cea2aj, line 42: Is the removal of many of the classes from the interfaces intentional? We have generally followed the convention that all styling of the component should be customizable from the CSS class interface. For example, titleLink and ctaContainer are no longer customizable

tmeyer2115: I forgot to mention that in the commit message, but yes, this was intentional. Davish and I went through and took out a number of styling config options. The purpose of this Component is to show someone how to write a Card. We should show them how to support CSS customization. But, we don't need to make every part of the DOM customizable to do that.

src/components/cards/standard/StandardCard.tsx
```diff
@@ -0,0 +1,76 @@
...
+    dynamic data, returned from LiveAPI. This data must be type-checked at run-time. This is
+    accomplished by using the StandardCardDisplay's PropTypes.
+  */
+  return (
+    <StandardCardDisplay
+      customCssClasses={customCssClasses}
+      cssCompositionMethod={cssCompositionMethod}
+      showFeedbackButtons={showFeedbackButtons}
+      clickHandlers={clickHandlers}
+      // @ts-ignore
```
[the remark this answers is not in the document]
tmeyer2115: Yep! Done.

PR #131, opened 2022-05-02
add commonjs support

oshi97:
This PR adds commonjs support by separating the tsconfig into esm and commonjs versions,
updating our api extractor config, and using conditional exports to provide the right entry point.

J=SLAP-2074
TEST=manual

test that the dev command works - I can hook up a local answers-react-components to a local 
project, and changes in answers-react-components will automatically be reflected in my project

test that answers-react-components works with yext sites SSR in both prod and dev mode

### DIFF
--- a/api-extractor.json
+++ b/api-extractor.json
@@ -32,7 +32,7 @@
    * SUPPORTED TOKENS: <lookup>
    * DEFAULT VALUE: "<lookup>"
    */
-  // "projectFolder": "..",
+  "projectFolder": ".",
 
   /**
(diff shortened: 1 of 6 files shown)


.github/workflows/third_party_notices_check.yml
```diff
@@ -5,3 +5,5 @@ on: pull_request
 jobs:
   call_notices_check:
     uses: yext/slapshot-reusable-workflows/.github/workflows/third_party_notices_check.yml@v1
+    with:
+      environment: macos-latest
```
tmeyer2115, line 9: That's nifty.

tmeyer2115 (approved):

PR #141, opened 2022-05-04
i18next Prototype

cea2aj:
Example of adding i18next

This example shows how to use i18next for interpolation, pluralization, and intermixed markdown. By default, the components use an i18n instance which only supports english. I propose we distribute json files for the translations for the other languages, which can be added to the i18n instance as needed. So long as tree shaking is working, this makes it easy for sites to include only the languages they need. We can generate the json files from PO files using i18next-conv.

Languages can also be toggled at runtime. You would just need to make a button which calls i18n.changeLanguage on both i18n instances.

The components are all wrapped around a shared i18n instance for the components and the sample app gets access to its own i18n instance which glues to react context with the `.use(initReactI18next)` call.

J=SLAP-2082, SLAP-2083
TEST=manual

Load sample app run various queries and see the translations appear on the results component and the alternative verticals component


cea2aj:
Closing since we haven't moved forward with this for almost a year now

PR #149, opened 2022-05-11
Add stories for Pagination component

yen-tt:
Add 6 stories for Pagination component:
- Primary: pagination UI for a results count of a search
- PaginateAllOnNoResults: pagination UI using total results count of a vertical when there's no results from a search
- OnMidPageWithEllipses:  UI when user is on the middle page number of the result set, with many more pages on each side of it which would collapse into ellipses.
- OnMidPage:  UI when user is on the middle page number of the result set, no ellipses.
- OnLastPage: UI when user is on the last page number of the result set
- Loading: when a search is in loading state

J=SLAP
TEST=auto

Screen Shot 2022-05-11 at 3 39 55 PMScreen Shot 2022-05-11 at 3 40 07 PMScreen Shot 2022-05-11 at 3 40 15 PM

see that percy build captured all the new snapshots

### DIFF
--- a/tests/components/Pagination.stories.tsx
+++ b/tests/components/Pagination.stories.tsx
@@ -1,0 +1,108 @@
+import React from 'react';
+import { ComponentMeta } from '@storybook/react';
+import { AnswersHeadlessContext } from '@yext/answers-headless-react';
+
+import { Pagination } from '../../src/components/Pagination';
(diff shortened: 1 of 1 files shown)


oshi97 (approved):
💯 💯 💯

PR #150, opened 2022-05-11
update answers-headless-react to 1.1.1

oshi97:


nmanu1 (commented):
does the component lib version need to be updated too?

oshi97:
> does the component lib version need to be updated too?

I don't think there's much reason to publish a new beta version. People who install the current version will automatically get the newest answers-headless-react even without us changing anything.

PR #153, opened 2022-05-16
Add Storybook controls

nmanu1:
This PR adds Storybook controls for each component's stories. This allows almost all component props to be configured from the Storybook UI. Props that are functions or objects containing functions (e.g. `onSearch`, `onClick`, `renderEntityPreviews`, etc.) cannot be directly configured via the controls (see https://storybook.js.org/docs/react/essentials/controls#dealing-with-complex-values).

Additionally, there are a few props that can be set, but their values don't affect the components because we are hardcoding the returned response results. I disabled the controls for these, but if we want to update the response based on those props in the future, they can easily be added back:
- `hierarchicalFacetsDelimiter` in `AppliedFiltersDisplay`
- `sectioned` and `searchFields` in `FilterSearch`
- `geolocationOptions` in `LocationBias` and `SearchBar`

The `expanded: true` config option in `.storybook/preview.js` ensures that the prop descriptions remain visible in the table on the Storybook site even after controls are enabled.

Lastly, I noticed that the applied filters styling was not being applied for universal results, so the `builtInCssClasses` in `UniversalResults` was updated to fix that and the corresponding stories were updated to display the applied filters by default.

J=SLAP-2094
TEST=manual

Went through each component's stories and checked that each prop is configurable using the controls and changes the display as expected except for those exceptions listed above.

### DIFF
--- a/tests/__compound-components__/Filters/Facets.tsx
+++ b/tests/__compound-components__/Filters/Facets.tsx
@@ -1,8 +1,9 @@
 import { Filters } from '../../../src/components';
 
-export function Facets({ searchOnChange = true }: { searchOnChange?: boolean }): JSX.Element {
+export function Facets(args: Filters.FacetsProps): JSX.Element {
+  const searchOnChange = args.searchOnChange === false ? false : true;
   return (
-    <Filters.Facets searchOnChange={searchOnChange}>
+    <Filters.Facets {...args}>
       {facets => {
         const filteredFacets = facets.filter(f => f.options.length > 0);
         return (
--- a/tests/__compound-components__/Filters/Facets.tsx
+++ b/tests/__compound-components__/Filters/Facets.tsx
@@ -1,9 +1,12 @@
 import { Filters } from '../../../src/components';
 
 export function Facets(args: Filters.FacetsProps): JSX.Element {
-  const searchOnChange = args.searchOnChange === false ? false : true;
+  const config = {
+    searchOnChange: true,
+    ...args
+  };
   return (
-    <Filters.Facets {...args}>
+    <Filters.Facets {...config}>
       {facets => {
         const filteredFacets = facets.filter(f => f.options.length > 0);
         return (
@@ -22,7 +25,7 @@ export function Facets(args: Filters.FacetsProps): JSX.Element {
                       />
                     )}
                   </Filters.CollapsibleSection>
-                  {filteredFacets.length > 0 && !searchOnChange && <Filters.ApplyFiltersButton />}
+                  {filteredFacets.length > 0 && !config.searchOnChange && <Filters.ApplyFiltersButton />}
                 </Filters.FilterGroup>
               );
             })}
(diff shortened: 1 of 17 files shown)


tests/__compound-components__/Filters/Facets.tsx
```diff
@@ -1,8 +1,9 @@
 import { Filters } from '../../../src/components';
 
-export function Facets({ searchOnChange = true }: { searchOnChange?: boolean }): JSX.Element {
+export function Facets(args: Filters.FacetsProps): JSX.Element {
+  const searchOnChange = args.searchOnChange === false ? false : true;
```
tmeyer2115, line 4: Do we need to merge this with `args`?

tests/__compound-components__/Filters/Facets.tsx
```diff
@@ -1,8 +1,9 @@
 import { Filters } from '../../../src/components';
 
-export function Facets({ searchOnChange = true }: { searchOnChange?: boolean }): JSX.Element {
+export function Facets(args: Filters.FacetsProps): JSX.Element {
```
tmeyer2115, line 3: For the pure Jest tests, do we need to modify the args of the the top-level function? I understand why we need to do this for the Stories, that's how we get controls working properly.

tests/components/AppliedFilters.stories.tsx
```diff
@@ -5,15 +5,15 @@ import { AnswersHeadlessContext, SearchTypeEnum } from '@yext/answers-headless-r
 import { generateMockedHeadless } from '../__fixtures__/answers-headless';
 import { AppliedFiltersDisplay } from '../../src/components/AppliedFiltersDisplay';
 import { DisplayableFilters, DisplayableHierarchicalFacets } from '../__fixtures__/data/filters';
-import { builtInCssClasses } from '../../src/components/AppliedFilters';
+import { AppliedFiltersProps, builtInCssClasses } from '../../src/components/AppliedFilters';
```
yen-tt, line 8: why can't we make `hierarchicalFacetsDelimiter` configurable from the pr description? Also, I think it may be better to avoid confusion if we disable control configuration for `geolocationOptions`, `sectioned`, and `searchFields` since the response is mocked a specific way at the moment.

tests/__compound-components__/Filters/Facets.tsx
yen-tt, line 4: nit: maybe we can have a defaultConfig object so it's clearer? (`{...defaultConfig, ...args}`)

tests/__compound-components__/Filters/Facets.tsx
nmanu1: These top-level function components are used in their corresponding Stories, so in order to set the controls for those Stories, the args need to be passed in

tests/components/AppliedFilters.stories.tsx
nmanu1: `hierarchicalFacetsDelimiter` in `AppliedFiltersDisplay` is only used when removing an applied filter and updating the hierarchical filters accordingly. But, since we are passing in hard-coded filters directly into `AppliedFiltersDisplay`, the re-computed applied filters are never used or displayed so changing the delimiter has no effect. I'll update to disable all controls that don't have an effect

tests/__compound-components__/Filters/Facets.tsx
nmanu1: It's not necessary, because `searchOnChange` is true by default in the `Facets` component. But, I updated to merge it in for the sake of clarity

nmanu1: Since there's only one default option (`searchOnChange`), I updated to merge it with `args` directly

yen-tt (approved):

PR #154, opened 2022-05-17
Add jest tests for AlternativeVerticals component

yen-tt:
Added jest tests for AlternativeVerticals component that checks if the href and label of the vertical and universal links are as expected, using the default and custom `getSuggestionUrl` function. Also, a jest test for when there's no vertical suggestions from the response.

Since there's already UI check for `verticalConfigMap` and `displayAllOnNoResults` config option from storybook for AlternativeVerticals, there's no jest tests added for it.

J=SLAP-2107
TEST=auto

see that jest tests passed

### DIFF
--- a/tests/components/AlternativeVerticals.test.tsx
+++ b/tests/components/AlternativeVerticals.test.tsx
@@ -1,0 +1,94 @@
+import { render, screen, within } from '@testing-library/react';
+import { State } from '@yext/answers-headless-react';
+
+import { AlternativeVerticals } from '../../src/components/AlternativeVerticals';
+import { mockAnswersHooks } from '../__utils__/mocks';
+import { verticalNoResults } from '../__fixtures__/data/verticalnoresults';
+import { VerticalLink } from '../../src/models/verticalLink';
+import { UniversalLink } from '../../src/models/universalLink';
+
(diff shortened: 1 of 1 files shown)


tests/components/AlternativeVerticals.test.tsx
```diff
@@ -0,0 +1,105 @@
...
+    );
+    const expectedLinkHrefs = [
+      'http://localhost/sometext+events',
+      'http://localhost/sometext+faqs',
+      'http://localhost/sometext+universal'
+    ];
+    const actualLinkHrefs = screen.getAllByRole('link')
+      .map(link => (link as HTMLAnchorElement).href);
+    expect(actualLinkHrefs).toEqual(expectedLinkHrefs);
+    expect(mockedGetSuggestionUrl).toBeCalledTimes(3);
```
tmeyer2115, line 92: Small nit: is this something we need to test? Can we infer the number of times it's been called by the previous assertion and the assertions in the test above? It feels like the only thing we're testing with this assertion is that we don't call the function too many times.

oshi97 (approved):

tmeyer2115 (approved):

PR #156, opened 2022-05-19
Add `StaticFilters` component with baked-in layout

nmanu1:
This PR adds a higher-level `StaticFilters` component that uses the existing compound components and has a baked-in layout. Only basic static filters (with an `$eq` matcher) are supported. Each filter option in this component will have the same `fieldId`, so a page might have multiple `StaticFilters` components to filter on multiple `fieldId`s.

While testing, I found a bug with our current implementation of selecting a filter by default. The Clear All button of applied filters would clear out the static filters entirely, so each time the static filters that were selected by default would be applied again even after clearing them. This was fixed by updating the Clear All button to unselect static filters instead of clearing them out of state.

The `CollapsibleSection` styling was updated to ensure the apply filters button styling remained as is even when included in a `CollapsibleSection`. Removing the flex styling didn't have any other effects on desktop or mobile.

Lastly, since we eventually plan to not export any of the `Filters` compound components, I commented out the `StaticFilters` import in `Filters/index.ts` but left it in the file. Once we can remove all exports, I think it would be easier to remove the `Filters` export from `components/index.ts` and uncomment the exports in `Filters/index.ts` once more. This way, internally we'll be able to import all of the compound components from the same `Filters` module path even if we no longer export them for users.

J=SLAP-2105
TEST=auto, manual

Check that Jest tests pass and test each prop for `StaticFilters` with the test-site to ensure correct behavior.


src/components/StaticFilters.tsx
```diff
@@ -0,0 +1,133 @@
...
+  searchable?: boolean,
+  /**
+   * Whether or not a search is automatically run when a filter is selected.
+   * Defaults to true.
+   */
+  searchOnChange?: boolean,
+  /** CSS classes for customizing the component styling. */
+  customCssClasses?: StaticFiltersCssClasses,
+  /** {@inheritDoc CompositionMethod} */
+  cssCompositionMethod?: CompositionMethod
```
oshi97, line 71: Is it intentional that `cssCompositionMethod` only applies to CheckboxOption here? I feel like that's a little confusing.

src/components/StaticFilters.tsx
```diff
@@ -0,0 +1,133 @@
...
+  collapsible = true,
+  defaultExpanded = true,
+  searchable,
+  searchOnChange = true,
+  customCssClasses = {},
+  cssCompositionMethod
+}: StaticFiltersProps): JSX.Element {
+  const cssClasses = useMemo(() => {
+    return {
+      ...customCssClasses,
```
oshi97, line 96: Could we remove these additional spreads and have this be:
```ts
return {
  ...customCssClasses,
  optionContainer: customCssClasses.option,
  label: customCssClasses.optionLabel,
  input: customCssClasses.optionInput
}
```
Also, is it intentional that this memoized `cssClasses` has both the original keys from customCssClasses and also the renamed keys of optionContainer, label, and input? That is that it has the shape of
```ts
{
  // custom css classes shape
  container?: string,
  searchInput?: string,
  optionsContainer?: string,
  option?: string,
  optionInput?: string,
  optionLabel?: string,
  applyFiltersButton?: string,
  
  // renamed keys
  optionContainer: string,
  label: string,
  input: string
}
```

oshi97: This results in some extra keys that I think could lead to confusing behavior in the future.

tests/components/AppliedFilters.test.tsx
```diff
@@ -135,15 +135,15 @@ describe('AppliedFilters', () => {
     expect(isSelected).toBe(false);
   });
 
-  it('The clear all button clears all filters', () => {
+  it('The clear all button unselect all filters', () => {
```
oshi97, line 138: unselects

src/components/StaticFilters.tsx
```diff
@@ -0,0 +1,133 @@
...
+      : (title &&
+        <div className='text-neutral-dark text-sm font-medium text-left mb-4'>
+          {title}
+        </div>);
+  }
+
+  return (
+    <StaticFiltersCompoundComponent searchOnChange={searchOnChange} className={cssClasses.container}>
+      <FilterGroup fieldId={fieldId} defaultExpanded={!collapsible || defaultExpanded}>
+        {renderTitle()}
```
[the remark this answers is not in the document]
nmanu1: `renderTitle()` uses `CollapsibleLabel` when `collapsible = true`

src/components/StaticFilters.tsx
nmanu1: It is intentional because the other compounds components, besides `CheckboxOption`, set their styling with a single `className` prop, so they don't accept a composition method. I think the only case where the behavior might differ from what's expected is for the `merge` composition method because the `className` would still be replaced instead of merged. `assign` and `replace` would match current behavior even without passing along the composition method.

Since that could be confusing, I could update the docs and/or name it something like `optionsCssCompositionMethod` to make it clear that it's only for the checkbox options. Alternatively, since it still might be a little weird to have a composition method for only some of the CSS classes, maybe we could compose all of the classes before passing them to the different components. The downside of that would be some additional complexity and we'd need to export the default styling for the compound components with the single `className` prop (they're currently just set inline as a default for the prop).

src/components/StaticFilters.tsx
nmanu1: I had tried it without the additional spreads initially, but I noticed that it would pass in 
```ts
{
  optionContainer: undefined,
  label: undefined,
  input: undefined
}
```
as the `customCssClasses` to `CheckboxOption`. Since it explicitly sets those classes as undefined, all the styling for those classes was removed, regardless of composition method. So, I added the additional spreads to make sure the classes weren't set (even as undefined) when the classes aren't defined in the props.

I can update to explicitly list the CSS classes so the extra keys aren't included.

src/components/StaticFilters.tsx
oshi97: I like the idea floating around in slack of getting rid of cssCompositionMethod and always having it be merge

src/components/StaticFilters.tsx
oshi97: Ah I see, that's unfortunate but also makes sense

PR #158, opened 2022-05-23
Add Jest tests for StaticFilters

nmanu1:
This PR adds Jest tests for the higher-level `StaticFilters` component.

J=SLAP-2111
TEST=auto

Check that the added Jest tests pass.

### DIFF
--- a/src/components/Filters/CheckboxOption.tsx
+++ b/src/components/Filters/CheckboxOption.tsx
@@ -105,7 +105,7 @@ export function CheckboxOption(props: CheckboxOptionProps): JSX.Element | null {
       return false;
     }
 
-    if (!answersUtilities.isCloseMatch(label.toString(), searchValue)) {
+    if (!answersUtilities.isCloseMatch(label, searchValue)) {
       return false;
     }
 
--- a/tests/components/StaticFilters.test.tsx
+++ b/tests/components/StaticFilters.test.tsx
@@ -1,0 +1,189 @@
+import { render, screen } from '@testing-library/react';
+import { AnswersHeadless, State } from '@yext/answers-headless-react';
+import { mockAnswersHooks, spyOnActions } from '../__utils__/mocks';
+import userEvent from '@testing-library/user-event';
+import { FilterOptionConfig, StaticFilters } from '../../src/components';
+import { staticFilters } from '../__fixtures__/data/filters';
+
+const mockedState: Partial<State> = {
+  filters: {
+    static: staticFilters
+  },
+  vertical: {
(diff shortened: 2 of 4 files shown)


src/components/Filters/CheckboxOption.tsx
```diff
@@ -105,7 +105,7 @@ export function CheckboxOption(props: CheckboxOptionProps): JSX.Element | null {
       return false;
     }
 
-    if (!answersUtilities.isCloseMatch(label.toString(), searchValue)) {
```
oshi97, line 108: 🔥 🔥 🔥

tests/components/StaticFilters.test.tsx
```diff
@@ -0,0 +1,197 @@
...
+  it('Clicking a filter option executes a search when searchOnChange is true', () => {
+    const actions = spyOnActions();
+    render(<StaticFilters {...staticFiltersProps} />);
+
+    const martyCheckbox: HTMLInputElement = screen.getByLabelText('MARTY!');
+    userEvent.click(martyCheckbox);
+    expect(actions.executeVerticalQuery).toBeCalled();
+  });
+
+  it('Only clicking the apply button executes a search when searchOnChange is false', () => {
```
oshi97, line 166: "only clicking" might be a little unclear, I read it as "I only need to click the apply button to search" as opposed to "the only way I am able to search is by clicking the apply button"

oshi97 (commented):
nice tests!

tests/components/StaticFilters.test.tsx
nmanu1: updated!

yen-tt (commented):
Looks like changes from the baked in StaticFilters pr caused some potential errors with AppliedFilters. Specifically, when you applied an option from StaticFilters, that option is displayed twice in AppliedFilters.

nmanu1:
> Looks like changes from the baked in StaticFilters pr caused some potential errors with AppliedFilters. Specifically, when you applied an option from StaticFilters, that option is displayed twice in AppliedFilters.

This is just a bad example of using `StaticFilters` in the test-site. We don't support filtering on the same `fieldId` across multiple components (between multiple `StaticFilters` components or between `Facets` and `StaticFilters`). So the behavior is a little weird when we filter on puppy preference because it's applying the static filter and then also applying the corresponding facet option filter, so 2 identical applied filters are added. I updated the test-site `StaticFilters` to filter on a different `fieldId` to only test behavior we support

oshi97 (approved):

PR #160, opened 2022-05-25
Add Facets component with baked-in layout and export ApplyFiltersButton

yen-tt:
this pr adds 3 higher level `Facets` components, `StandardFacets`, `HierarchicalFacets`, and `NumericalFacets`, that uses the Facets compound component structure internally. The props and interfaces follow closely with the `StaticFilters` component.

- Since `Facets` and `StaticFilters` share similar setup for the FilterGroup part, this pr extracted that out into its own component to use internally.
- To avoid duplicate `ApplyFiltersButton` on page when `searchOnChange` is false in Facet/StaticFilters component,  `ApplyFiltersButton` is exported as a stand alone component and will NOT be render as part of Facet/StaticFilters component
- update storybook and jest tests accordingly

~Note: this item excludes Hierarchical & Numerical Facets.~ After talking with product and the team, we will be supporting it in the same higher level component  

J=SLAP-2104
TEST=manual and auto

see that jest tests passed
see that People page in test-site, which uses Facets, StaticFilters, and ApplyFiltersButton, works as expected


etc/answers-react-components.api.md
```diff
@@ -357,8 +379,8 @@ declare namespace Filters {
         CollapsibleLabelProps,
         CollapsibleSection,
         CollapsibleSectionProps,
-        Facets,
-        FacetsProps,
+        Facets_2 as Facets,
+        FacetsProps_2 as FacetsProps,
```
nmanu1, line 383: can we remove the Facets exports from the Filters namespace so we're not exposing both to the user?

src/components/ApplyFiltersButton.tsx
```diff
@@ -1,29 +1,44 @@
...
 }
 
+
+/**
+ * The CSS class interface for {@link ApplyFiltersButtonProps}.
+ *
+ * @public
+ */
+export interface ApplyFiltersButtonCssClasses {
+  button?: string
```
nmanu1, line 24: did we change this to an interface instead of a single className because we anticipate adding more CSS classes to the component?

src/components/Facets.tsx
```diff
@@ -0,0 +1,88 @@
...
+  optionLabel?: string
+}
+
+/**
+ * Props for the {@link Facets} component.
+ *
+ * @public
+ */
+export interface FacetsProps {
+  /** Whether or not the filter is collapsible. Defaults to true. */
```
nmanu1, line 27: can we inherit most of these docs from the new FilterGroup component to avoid duplication (and similarly for StaticFiltersProps)?

src/components/Facets.tsx
```diff
@@ -0,0 +1,88 @@
...
+import { FilterGroup } from './FilterGroup';
+import { CompositionMethod } from '../hooks/useComposedCssClasses';
+import { Fragment } from 'react';
+
+/**
+ * The CSS class interface for {@link Facets}.
+ *
+ * @public
+ */
+export interface FacetsCssClasses {
```
nmanu1, line 11: can this be an extension of the FilterGroupCssClasses interface (and similarly for StaticFiltersCssClasses)?

src/components/Facets.tsx
```diff
@@ -0,0 +1,88 @@
...
+ * A component that displays all facets applicable to the current vertical search.
+ *
+ * @param props - {@link FacetsProps}
+ * @returns A React component for facets
+ *
+ * @public
+ */
+export function Facets(props: FacetsProps) {
+  const { searchOnChange, customCssClasses = {}, ...filterGroupProps } = props;
+  const { container:containerCssClasses, divider, ...filterGroupCssClasses } = customCssClasses;
```
nmanu1, line 57: what's the reason for renaming the container class?

src/components/StaticFilters.tsx
```diff
@@ -21,23 +13,15 @@ export interface StaticFiltersCssClasses {
...
  */
-export interface FilterOptionConfig {
-  /** {@inheritDoc Filters.CheckboxOptionProps.value} */
-  value: string | number | boolean,
-  /** {@inheritDoc Filters.CheckboxOptionProps.label} */
-  label?: string,
-  /** {@inheritDoc Filters.CheckboxOptionProps.selectedByDefault} */
-  selectedByDefault?: boolean
-}
+export type StaticFilterOptionConfig = Omit<FilterOptionConfig, 'matcher'>;
```
nmanu1, line 24: since we're only accepting `$eq` matcher, then we also need to limit the `value` to not accept `NumberRangeValue`.

but, it might also just be easier to accept any matcher and then we can probably just use `FilterOptionConfig` here. I'm not sure if we've tested that for static filters though

src/components/StaticFilters.tsx
```diff
@@ -48,9 +32,9 @@ export interface StaticFiltersProps {
   /** The fieldId corresponding to the static filter group. */
   fieldId: string,
   /** {@inheritDoc FilterOptionConfig} */
-  filterOptions: FilterOptionConfig[],
+  filterOptions: StaticFilterOptionConfig[],
   /** The displayed label for the static filter group. */
-  title?: string,
+  title: string,
```
nmanu1, line 37: it seems odd that the title is required here when it's optional in FilterGroup. I think we should keep it consistent

tests/components/Filters/Facets.test.tsx
```diff
@@ -130,7 +130,7 @@ describe('Facets', () => {
     expect(actions.executeVerticalQuery).toBeCalled();
   });
 
-  it('Only clicking the apply button executes a search when searchOnChange is false', () => {
+  it('Clicking a facet option does not execute a search when searchOnChange is true', () => {
```
nmanu1, line 133: should this be when searchOnChange is false?

tests/components/StaticFilters.test.tsx
```diff
@@ -158,12 +156,16 @@ describe('Static Filters', () => {
...
+    expect(martyCheckbox.checked).toBeFalsy();
+
     userEvent.click(martyCheckbox);
-    expect(actions.executeVerticalQuery).toBeCalled();
+    expectFilterOptionSet(actions, staticFiltersProps.fieldId, martyFilter, true);
+    expect(actions.executeVerticalQuery).toBeCalledTimes(1);
   });
 
-  it('When searchOnChange is false, the apply button must be clicked to execute a search', () => {
+  it('Clicking a filter option does not execute a search when searchOnChange is fa;se', () => {
```
nmanu1, line 168: fa;se

tests/components/StaticFilters.test.tsx
```diff
@@ -158,12 +156,16 @@ describe('Static Filters', () => {
...
     render(<StaticFilters {...staticFiltersProps} />);
 
-    const martyCheckbox: HTMLInputElement = screen.getByLabelText('MARTY!');
+    const martyFilter = staticFiltersProps.filterOptions[0];
+    const martyCheckbox: HTMLInputElement = screen.getByLabelText(martyFilter.label);
+    expect(martyCheckbox.checked).toBeFalsy();
+
     userEvent.click(martyCheckbox);
-    expect(actions.executeVerticalQuery).toBeCalled();
+    expectFilterOptionSet(actions, staticFiltersProps.fieldId, martyFilter, true);
```
nmanu1, line 164: I think we already check that setFilterOption is called correctly when clicking a checkbox option in other tests above, so not sure it's needed here as well

src/components/Facets.tsx
yen-tt: just to be more specific that this is cssClasses

etc/answers-react-components.api.md
yen-tt: HierarchicalFacet, at least in our tests, still depends on it

src/components/ApplyFiltersButton.tsx
yen-tt: It's to be more consistent with the rest of the components

src/components/Facets.tsx
yen-tt: sure thing

src/components/StaticFilters.tsx
yen-tt: oh I think it should be required in FilterGroup actually

src/components/StaticFilters.tsx
yen-tt: The discussion thread on StaticFilters with product seems to settle on accepting $eq only so we can keep it as is. I will limit `value` to omit numberRangeValue

src/components/Facets.tsx
yen-tt: sure thing

etc/answers-react-components.api.md
nmanu1: even in that case I think we can remove the export in Filters/index.ts? because in the tests we can just change to import it directly from src/components/Filters/Facets.tsx instead of src/components/Filters

tests/components/Filters/Facets.stories.tsx
```diff
@@ -5,13 +5,12 @@ import { AnswersHeadlessContext, State } from '@yext/answers-headless-react';
 import { generateMockedHeadless } from '../../__fixtures__/answers-headless';
 import { RecursivePartial } from '../../__utils__/mocks';
 import { DisplayableFacets } from '../../__fixtures__/data/filters';
-import { Filters } from '../../../src/components';
-import { Facets } from '../../__compound-components__/Filters/Facets';
+import { Facets, FacetsProps } from '../../../src';
```
nmanu1, line 8: I think now we can move the Facets stories and jest test files out of the Filters directory to be at the same level as StaticFilters

etc/answers-react-components.api.md
yen-tt: sure

nmanu1: actually, thinking about it more, if we use the Facets compound component in HierarchicalFacets, then other people would need it too. so until we make a baked-in component for that we should probably export it to users even if it means our docs are a little messed up in the meantime

src/components/Facets.tsx
nmanu1: I thought we used cssClasses when there was more than one class, and here it's just container?

yen-tt: we usually have `blahCssClasses.container` or `containerCssClasses`, so the rename is so it's a bit more clear that this isn't a container element or anything but a string of css classes. If you feel strongly about it, I can update it to just `container`

etc/answers-react-components.api.md
yen-tt: I believe we will have changes for a higher level HierarchicalFacets but I can add this back for now. Our docs wouldn't be messed up if you look at the documentation pages on repo since this is exported under a namespace 'Filters' - it just looks like it from the files here

nmanu1: ah, got it!

src/components/Facets.tsx
nmanu1: I think maybe it would be better to name it `containerClassName` so it still makes it clear that it's a string for styling and not an element, but also make it clear that it's a single class name and not an object with multiple classes.

in the same way, we should probably change the divider CSS class name also. and maybe in the Divider component, we can change the prop name to `className` instead of `cssClasses` to indicate it's a single class and not an object

test-site/src/pages/PeoplePage.tsx
```diff
@@ -7,10 +8,12 @@ import {
...
   LocationBias,
-  StaticFilters
+  StaticFilters,
+  Facets,
+  ApplyFiltersButton
 } from '@yext/answers-react-components';
-import { useLayoutEffect } from 'react';
-import { Facets, hierarchicalFacetFieldIds } from '../components/Facets';
+
+export const hierarchicalFacetFieldIds = ['c_hierarchicalFacet'];
```
oshi97, line 16: why is this exported?

src/components/Facets.tsx
```diff
@@ -0,0 +1,80 @@
...
+  return (
+    <FacetsCompoundComponent searchOnChange={searchOnChange} className={containerCssClasses}>
+      {facets => {
+        const filteredFacets = facets.filter(f => f.options.length > 0);
```
oshi97, line 53: Is the plan right now to push HierarchicalFacets off to a later date, and for now hierarchical facets will just get rendered as regular facets?

src/components/StaticFilters.tsx
```diff
@@ -1,43 +1,25 @@
...
  */
-export interface FilterOptionConfig {
-  /** {@inheritDoc Filters.CheckboxOptionProps.value} */
-  value: string | number | boolean,
-  /** {@inheritDoc Filters.CheckboxOptionProps.label} */
-  label?: string,
-  /** {@inheritDoc Filters.CheckboxOptionProps.selectedByDefault} */
-  selectedByDefault?: boolean
-}
+export type StaticFilterOptionConfig = Omit<FilterOptionConfig, 'matcher' | 'value'> & {
```
oshi97, line 19: I think sharing types across different values can be nice, but maybe we need to think about this one a little more? Currently `StaticFilterOptionConfig` is `Omit<FilterOptionConfig, 'matcher' | 'value'>` and then override `value` with a different type, and `FilterOptionConfig` is `Omit<CheckboxOptionProps, 'customCssClasses' | 'cssCompositionMethod'>`. I feel like that's a lot of mental gymnastics and isn't really a clear mental model

src/components/StaticFilters.tsx
```diff
@@ -80,55 +59,16 @@ export interface StaticFiltersProps {
...
-      ? <CollapsibleLabel label={title} />
-      : (title &&
-        <div className='text-neutral-dark text-sm font-medium text-left mb-4'>
-          {title}
-        </div>);
-  }
-
+export function StaticFilters(props: StaticFiltersProps): JSX.Element {
+  const { searchOnChange, customCssClasses = {}, ...filterGroupProps } = props;
+  const { container:containerCssClasses, ...filterGroupCssClasses } = customCssClasses;
```
oshi97, line 64: space after colon

test-site/src/pages/PeoplePage.tsx
```diff
@@ -23,16 +26,20 @@ export function PeoplePage() {
       <SearchBar />
       <div className='flex'>
         <div className='min-w-fit pr-4'>
-          <Facets />
+          <Facets
+            searchOnChange={false}
```
oshi97, line 30: nit let's make this one line

src/components/Facets.tsx
```diff
@@ -0,0 +1,80 @@
...
+ * @returns A React component for facets
+ *
+ * @public
+ */
+export function Facets(props: FacetsProps) {
+  const { searchOnChange, customCssClasses = {}, ...filterGroupProps } = props;
+  const { container:containerCssClasses, divider, ...filterGroupCssClasses } = customCssClasses;
+  return (
+    <FacetsCompoundComponent searchOnChange={searchOnChange} className={containerCssClasses}>
+      {facets => {
```
oshi97, line 52: I think this can just be 
```tsx
{facets.filter(blah).map((f,i) => {
```

src/components/FilterGroup.tsx
```diff
@@ -0,0 +1,111 @@
...
+  /** CSS classes for customizing the component styling. */
+  customCssClasses?: FilterGroupCssClasses,
+  /** {@inheritDoc CompositionMethod} */
+  cssCompositionMethod?: CompositionMethod
+}
+
+/**
+ * Renders a group of selectable filters with support for searching and collapsing.
+ */
+export function FilterGroup({
```
oshi97, line 62: I don't think we should have two components with the same name

src/components/Facets.tsx
```diff
@@ -0,0 +1,80 @@
...
+          {filteredFacets.map((f, i) => {
+            const filterOptions = f.options.map(option => ({ ...option, label: option.displayName }));
+            return (
+              <Fragment key={f.fieldId}>
+                <FilterGroup
+                  fieldId={f.fieldId}
+                  filterOptions={filterOptions}
+                  title={f.displayName}
+                  customCssClasses={filterGroupCssClasses}
+                  {...filterGroupProps}
```
oshi97, line 64: We're intentionally passing cssCompositionMethod through right?

oshi97: actually, where is the cssCompositionMethod prop used in this component?

src/components/Facets.tsx
```diff
@@ -0,0 +1,80 @@
...
+            );
+          })}
+        </>);
+      }}
+    </FacetsCompoundComponent>
+  );
+}
+
+function Divider({
+  cssClasses = 'w-full h-px bg-gray-200 my-4'
```
oshi97, line 77: Normally cssClasses means an object in our library right? Should this just be className instead?

oshi97: Also I don't think we need this component? All it's doing is providing a default className for the div, and it's only called one time?

src/components/Facets.tsx
```diff
@@ -0,0 +1,80 @@
...
+export function Facets(props: FacetsProps) {
+  const { searchOnChange, customCssClasses = {}, ...filterGroupProps } = props;
+  const { container:containerCssClasses, divider, ...filterGroupCssClasses } = customCssClasses;
+  return (
+    <FacetsCompoundComponent searchOnChange={searchOnChange} className={containerCssClasses}>
+      {facets => {
+        const filteredFacets = facets.filter(f => f.options.length > 0);
+        return (<>
+          {filteredFacets.map((f, i) => {
+            const filterOptions = f.options.map(option => ({ ...option, label: option.displayName }));
```
oshi97, line 56: All this map does is copy `displayName` into `label` right? Can we just, use displayName instead?

src/components/Facets.tsx
yen-tt: the item mentioned to excludes Hierarchical & Numerical Facets. I imagined we will schedule work for it later

src/components/Facets.tsx
yen-tt: yes, it's pass through to FilterGroup and into CheckboxOption

src/components/Facets.tsx
yen-tt: this is because CheckboxOption, use in FilterGroup, expects a label prop in its interface. StaticFilters also have label prop as agreed by product. Since StaticFilters and Facets both use FilterGroup, this mapping is needed in Facets for it to be consistent too with `label`

src/components/Facets.tsx
yen-tt: Just thought it would be easier to recognize the div as a divider for the UI, but I can update to include it directly in the main component!

src/components/Facets.tsx
yen-tt: because of how the FacetsCompoundComponent is structured, providing `facets` as children, I have to follow the format `{facets => blah}`. But I can update to combine filter and map part

src/components/Facets.tsx
yen-tt: sure

test-site/src/pages/PeoplePage.tsx
yen-tt: oops, copied over from the all Facets component in test-site and forgot to remove export

src/components/Facets.tsx
nmanu1: if this component should exclude Numerical Facets, should we only support primitive value types and basic matchers?

src/components/Facets.tsx
oshi97: Oh I see I think that's a good point!

src/components/Facets.tsx
```diff
@@ -0,0 +1,80 @@
+import { Facets as FacetsCompoundComponent } from './Filters/Facets';
```
oshi97, line 1: I didn't catch this for Nidhi's PR but could we rename the compound components for Facets and Filters to something else? I'd rather not have confusion when talking about components in the future. Maybe `FacetsProvider` since all it does is provide some behavior and Context setup? I'd avoid something like `CompoundFacets` or `FacetsCompoundComponent` because those names are imo more historical than descriptive

src/components/Facets.tsx
oshi97: Oh I see, did we want to wrap filterOptions in a useMemo then? Otherwise we're never getting reference equality since each time it renders it'll be a new array. Alternatively I don't mind if we accept `displayName` as an alias for `label` but I think Tom would be against that (cc: @tmeyer2115 )

src/components/FilterGroup.tsx
oshi97: This component basically provides the common layout we use for facets and filters so maybe something along those lines...? Honestly not sure

src/components/Facets.tsx
yen-tt: can't have useMemo inside callback. I will check in with product if we can use displayName instead in StaticFilters, I think switching `label` in Facets/StaticFilters/CheckboxOptionProps to `displayName` would be better align with the `Filter` interface we use in answers-headless and core, so we don't have to do additional mapping

src/components/StaticFilters.tsx
yen-tt: Since the components build on top of each other, I think it's kind of nice to have the configs depends on each other. This way, it's easier to make changes and/or keep the props align for all these components. I can reorganize it a bit so it's more clear.

src/components/Facets.tsx
yen-tt: product is good with switching to displayName for all so I will update the pr!

src/components/Facets.tsx
yen-tt: I see. Sounds good to me! Will update

src/components/FilterGroup.tsx
yen-tt: I think it makes sense to have this as` FiltersGroup`, product seems to use similar term to describe this group of selectable filters. I will update the components in `Filters` to `FiltersGroupContextProvider` `FacetsContextProvider` and ``StaticFiltersContextProvider``

src/components/Facets.tsx
yen-tt: per all hands discussion and confirmation with product, we will be including support for numerical and hierarchical facets in the same component so the interface should remain the same

src/components/NumericalFacets.tsx
```diff
@@ -0,0 +1,100 @@
...
+  getFilterDisplayName,
+  inputPrefix = <>$</>,
+  customCssClasses = {},
+  ...filterGroupProps
+}: NumericalFacetsProps) {
+  const rangeInputCssClasses = useMemo(() => {
+    return {
+      ...customCssClasses,
+      container: customCssClasses.inputContainer,
+      divider: customCssClasses.divider
```
nmanu1, line 67: is this needed if the name is staying the same?

src/components/NumericalFacets.tsx
```diff
@@ -0,0 +1,100 @@
...
+          return (
+            <Fragment key={f.fieldId}>
+              <FilterGroup
+                fieldId={f.fieldId}
+                filterOptions={f.options}
+                title={f.displayName}
+                customCssClasses={customCssClasses}
+                {...filterGroupProps}
+              >
+                <RangeInput
```
nmanu1, line 85: should this be configurable, or do most people always want the range input with their numerical facets?

src/components/NumericalFacets.tsx
```diff
@@ -0,0 +1,100 @@
...
+      }
+    </FacetsProvider>
+  );
+}
+
+export function isNumericalFacet(facet: DisplayableFacet): boolean {
+  return facet.options.length > 0 && isNumberRangeFilter(facet.options[0]);
+}
+
+function Divider({ className ='w-full h-px bg-gray-200 my-4' }: { className?: string }) {
```
nmanu1, line 104: it looks like we repeat this code in each facets component. can we abstract it out?

src/components/NumericalFacets.tsx
yen-tt: this should be:
`container: customCssClasses.rangeInputContainer,
 divider: customCssClasses.rangeInputDivider`
 since there's css name conflict between this component and the RangeInput one

src/components/NumericalFacets.tsx
yen-tt: product didn't mention anything on that. I could ask them

src/components/NumericalFacets.tsx
yen-tt: sure

yen-tt: I'm not sure if it's worth making this as its own thing since it's a simple div with some styling.

src/components/NumericalFacets.tsx
```diff
@@ -0,0 +1,105 @@
...
+ * A component that displays numerical facets applicable to the current vertical search.
+ *
+ * @param props - {@link NumericalFacetsProps}
+ * @returns A React component for facets
+ *
+ * @public
+ */
+export function NumericalFacets({
+  searchOnChange,
+  excludeFieldIds = [],
```
oshi97, line 56: would we want this to be `includedFIeldIds` instead? I imagine most facets will not be numerical (or hierarchical) and it could be annoying for people to type out all but one fieldId that they have facets for. 

I feel like product may end up just asking for both excluded and included props, but maybe for now we can just have `includedFieldIds` for both hierarchical and numerical facets?

src/components/HierarchicalFacets.tsx
```diff
@@ -0,0 +1,96 @@
...
+ * @param props - {@link HierarchicalFacetsProps}
+ * @returns A React component for facets
+ *
+ * @public
+ */
+export function HierarchicalFacets({
+  searchOnChange,
+  collapsible,
+  defaultExpanded,
+  excludeFieldIds = [],
```
nmanu1, line 54: why do we need both `excludeFieldIds` and `fieldIds`?

src/components/NumericalFacets.tsx
```diff
@@ -0,0 +1,105 @@
...
+  getFilterDisplayName,
+  inputPrefix = <>$</>,
```
oshi97, line 58: Could we define this JSX outside of the function instead of here? Otherwise we break the https://github.com/cvazac/eslint-plugin-react-perf/blob/master/docs/rules/jsx-no-jsx-as-prop.md rule
I'm not sure how strict we want to be for this stuff, but we've kind of committed to it (though maybe that's a sunk cost fallacy)

src/components/NumericalFacets.tsx
yen-tt: numerical facets will only display filters with a value shape of `{start: blah, end: blah}` so it wouldn't overlap with Hierarchical/StandardFacets ones. `excludeFieldIds` is inherited as the prop extends `StandardFacetsProps`. hierarchicalFacets already have `fieldIds` (I can rename) since there's no good way to distinguish it from the normal ones without assuming with delimiters. I can see why it would be good to have for numerical ones too, I can add in `includedFieldIds` for numerical facets

src/components/NumericalFacets.tsx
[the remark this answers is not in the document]
yen-tt: I believe product was fine with the defaults when discussed offline

src/components/HierarchicalFacets.tsx
yen-tt: Per comment above with oshi, I will remove`excludeFieldIds` (inherited from extending `StandardFacetsProps`) and have includedFieldIds for hierarchical and numerical

src/components/NumericalFacets.tsx
```diff
@@ -0,0 +1,105 @@
...
+  customCssClasses = {},
+  ...filterGroupProps
+}: NumericalFacetsProps) {
+  const rangeInputCssClasses = useMemo(() => {
+    return {
+      ...customCssClasses,
+      container: customCssClasses.rangeInputContainer,
+      divider: customCssClasses.rangeInputDivider
+    };
+  }, [customCssClasses]);
```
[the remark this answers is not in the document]
yen-tt: once we pass the cssClasses to `RangeInput`, it would limit it to the fields in `RangeInputCssClasses` so typescript wouldn't let you access rangeInputContainer and rangeInputDivider. But in the case we may make changes to the css interfaces, I will update to remove the original keys

oshi97 (approved):

PR #163, opened 2022-05-26
Add testing for LocationBias component

alextaing:
Added automated tests for the LocationBias component that address:
1. all elements of the component are rendered properly (displayName, attribution message, "Update your location" button)
2. setUserLocation uses coordinates returned by getUserLocation to set the updated location
3. clicking on "Update your location" properly renders new displayName if it is changed
4. correct attribution messages are rendered
5. absence of displayName correctly renders nothing

J=SLAP-2109
TEST=auto

### DIFF
--- a/tests/components/LocationBias.test.tsx
+++ b/tests/components/LocationBias.test.tsx
@@ -1,4 +1,5 @@
-import { fireEvent, render, screen, waitFor } from '@testing-library/react';
+import { render, screen, waitFor } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
 import { LocationBias } from '../../src/components/LocationBias';
 import { State, LocationBiasMethod } from '@yext/answers-headless-react';
 import * as locationOperations from '../../src/utils/location-operations';
@@ -75,11 +76,11 @@ it('renders the proper text (location DisplayName, method, and update btn)', () 
   const basedOnMethodElement = screen.getByText(expectedMethodMessage);
   expect(basedOnMethodElement).toBeDefined();
 
-  const updateLocationButton = screen.getByText('Update your location');
+  const updateLocationButton = screen.getByRole("button", {name: 'Update your location'});
   expect(updateLocationButton).toBeDefined();
 });
 
-it('calls setUserLocation with coordinates returned by getUserLocation as params when update location clicked', async () => {
+it('calls setUserLocation with coordinates returned by getUserLocation as params when update location is clicked', async () => {
   const actions = spyOnActions();
   render(<LocationBias />);
   clickUpdateLocation();
@@ -131,13 +132,12 @@ it('renders correct attribution message, IP', () => {
 
 it('renders nothing if there is no display name', () => {
   mockAnswersState(mockedStateNoDisplayName);
-  render(<LocationBias />);
-
-  const updateLocationButton = screen.queryByText('Update your location');
-  expect(updateLocationButton).toBeNull();
+  const { container }  = render(<LocationBias />);
+  console.log(container);
+  expect(container).toBeEmptyDOMElement();
 });
 
 function clickUpdateLocation() {
-  const updateLocationButton = screen.getByText('Update your location');
-  fireEvent.click(updateLocationButton);
+  const updateLocationButton = screen.getByRole('button', {name: 'Update your location'});
+  userEvent.click(updateLocationButton);
 }
--- a/tests/components/LocationBias.test.tsx
+++ b/tests/components/LocationBias.test.tsx
@@ -1,0 +1,108 @@
+import {fireEvent, render, screen} from "@testing-library/react";
+import { LocationBias } from "../../src/components/LocationBias";
+import { State, LocationBiasMethod } from "@yext/answers-headless-react";
+import { getUserLocation } from "../../src/utils/location-operations";
+
+import { mockAnswersState } from "../__utils__/mocks";
+
+const mockedState_VaFromDevice: Partial<State> = {
+  location: {
+    locationBias: {
+      latitude: 38.89552025579547,
+      longitude: -77.06991520330985,
+      displayName: 'Arlington, VA', 
+      method: LocationBiasMethod.Device
+    }
+  },
+  meta: {
+    searchType: 'universal'
+  },
+}
+
+const mockedState_NycFromIp: Partial<State> = {
+  location: {
+    locationBias: {
+      latitude: 40.741591687843005,
+      longitude: -74.00530254443494,
+      displayName: 'New York City, NY', 
+      method: LocationBiasMethod.Ip
+    }
+  },
+  meta: {
+    searchType: 'universal'
+  },
+}
+
+jest.mock("@yext/answers-headless-react");
+jest.mock("../../src/utils/location-operations", () => {
+  // const original = jest.requireActual("../../src/utils/location-operations");
+  return {
+      // ...original,
+      getUserLocation: jest.fn()
+  };
+});
+
+
+
+/// SIMPLE LOG LOCATION TEST
+
+beforeEach(() => {
+  const data = {
+    coords: {
+      accuracy: null,
+      altitude: null,
+      altitudeAccuracy: null,
+      heading: null,
+      latitude: 40.741591687843005,
+      longitude: -74.00530254443494,
+      speed: null,
+    },
+  timestamp: null
+  };
+  (getUserLocation as jest.Mock).mockResolvedValue(data);
+})
+
+it("log location", async () => {
+
+  const userLocation = await getUserLocation();
+  console.log(userLocation);
+  console.log(getUserLocation);
+});
+
+/////// TESTS
+
+// describe("LocationBias", () => {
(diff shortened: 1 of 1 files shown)


tests/components/LocationBias.test.tsx
```diff
@@ -0,0 +1,148 @@
...
+      mockedActions: {
+        state: mockedStateVaDevice,
+        setUserLocation: jest.fn(),
+      } 
+    });
+    jest.spyOn(locationOperations, "getUserLocation").mockImplementation(async ()=> {
+      return Promise.resolve(newGeoPosition);
+    });
+    jest.spyOn(searchOperations, "executeSearch").mockImplementation();
+    // jest.spyOn(searchOperations, "executeSearch").mockImplementation(()=> new Promise(()=>null));
```
oshi97, line 69: remove comment

tests/components/LocationBias.test.tsx
```diff
@@ -0,0 +1,148 @@
...
+
+    const expectedMethodMessage = "(based on your device) -";
+    const basedOnMethodElement = screen.getByText(expectedMethodMessage);
+    expect(basedOnMethodElement).toBeDefined();
+
+    const updateLocationButton = screen.getByText("Update your location");
+    expect(updateLocationButton).toBeDefined();
+  })
+
+  it("On update location click, setUserLocation is called with params being coordinates returned by getUserLocation", async () => {
```
oshi97, line 86: Wording a little weird here, specifically the "with params being".  Naming is hard.

We don't follow this rule a lot of the time, but one way to name your tests is as a readable sentence USING the "it" call as the first word. For example, if I had the test
`it('throws an error when I eat a cookie')`, you could read it as "It throws an error when I eat a cookie."
Again we probably follow this less than 50% of the time

tests/components/LocationBias.test.tsx
```diff
@@ -0,0 +1,148 @@
...
+    clickUpdateLocation();
+
+    const expectedCoordinates = {
+      latitude: newGeoPosition.coords.latitude,
+      longitude: newGeoPosition.coords.longitude
+    }
+    
+    await waitFor(() => {
+      expect(locationOperations.getUserLocation).toBeCalled();
+      expect(actions.setUserLocation).toBeCalledWith(expectedCoordinates);
```
oshi97, line 98: I think we can move the await for setUserLocation outside of the waitFor? If we can I slightly prefer that, because, well, we're not waiting on anything in setUserLocation.

tests/components/LocationBias.test.tsx
```diff
@@ -0,0 +1,148 @@
+import {fireEvent, render, screen, waitFor} from "@testing-library/react";
```
oshi97, line 1: Could you run eslint and fix any errors? Probably all of them can be autofixed with the `--autofix` flag. 
https://eslint.org/docs/user-guide/command-line-interface

tests/components/LocationBias.test.tsx
```diff
@@ -0,0 +1,148 @@
...
+      expect(searchOperations.executeSearch).toBeCalled();
+    })
+
+    const newExpectedLocationName = mockedStateNyIP.location.locationBias.displayName; 
+    const newLocationNameElement = screen.getByText(newExpectedLocationName);
+    expect(newLocationNameElement).toBeDefined();
+  });
+});
+
+it('Correct device attribution message rendered, device', () => {
```
oshi97, line 121: I would either include these tests in the big describe('LocationBias') block or just remove the describe block. I prefer to not have the describe block, especially for a component test file which is pretty explanatory, but most of our other component tests have this block that somehow I've missed my chance to get on a soapbox for so up to you which you prefer

tests/components/LocationBias.test.tsx
alextaing: done!

tests/components/LocationBias.test.tsx
alextaing: renamed

tests/components/LocationBias.test.tsx
```diff
@@ -0,0 +1,148 @@
...
+describe("LocationBias", () => {
+  beforeEach(() => {
+    mockAnswersHooks({
+      mockedState: mockedStateVaDevice,
+      mockedActions: {
+        state: mockedStateVaDevice,
+        setUserLocation: jest.fn(),
+      } 
+    });
+    jest.spyOn(locationOperations, "getUserLocation").mockImplementation(async ()=> {
```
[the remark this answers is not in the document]
alextaing: Yes, it does! I think we tried mockResolvedValue before moving it into the beforeEach block and thought it was a part of the problem, but it's not.  Fixed.

tests/components/LocationBias.test.tsx
alextaing: removed describe block

tests/components/LocationBias.test.tsx
```diff
@@ -0,0 +1,143 @@
...
+  render(<LocationBias />);
+  const expectedLocationName = mockedStateVaDevice.location.locationBias.displayName;
+  const locationNameElement = screen.getByText(expectedLocationName);
+  expect(locationNameElement).toBeDefined();
+
+  const expectedMethodMessage = '(based on your device) -';
+  const basedOnMethodElement = screen.getByText(expectedMethodMessage);
+  expect(basedOnMethodElement).toBeDefined();
+
+  const updateLocationButton = screen.getByText('Update your location');
```
nmanu1, line 78: I have a slight preference for accessing buttons using something like `screen.getByRole('button', {name: 'Update your location'})` so that we can be sure that it's a button and not just text being rendered

tests/components/LocationBias.test.tsx
```diff
@@ -0,0 +1,143 @@
...
+  expect(updateLocationButton).toBeDefined();
+});
+
+it('calls setUserLocation with coordinates returned by getUserLocation as params when update location clicked', async () => {
```
nmanu1, line 82: nit: "is clicked"

tests/components/LocationBias.test.tsx
```diff
@@ -0,0 +1,143 @@
...
+    expect(searchOperations.executeSearch).toBeCalled();
+  });
+
+  const newExpectedLocationName = mockedStateNyIP.location.locationBias.displayName;
+  const newLocationNameElement = screen.getByText(newExpectedLocationName);
+  expect(newLocationNameElement).toBeDefined();
+});
+
+it('renders correct attribution message, device', () => {
+  mockAnswersState(mockedStateVaDevice);
```
nmanu1, line 117: is this mock statement needed? I thought by default it would have the VA location in state?

tests/components/LocationBias.test.tsx
```diff
@@ -0,0 +1,143 @@
...
+  const basedOnMethodElement = screen.getByText(expectedMethodMessage);
+  expect(basedOnMethodElement).toBeDefined();
+});
+
+it('renders nothing if there is no display name', () => {
+  mockAnswersState(mockedStateNoDisplayName);
+  render(<LocationBias />);
+
+  const updateLocationButton = screen.queryByText('Update your location');
+  expect(updateLocationButton).toBeNull();
```
nmanu1, line 137: is there a way to check that nothing is rendered on the screen (not just that the update location button isn't there)?

tests/components/LocationBias.test.tsx
```diff
@@ -0,0 +1,143 @@
...
+});
+
+function clickUpdateLocation() {
+  const updateLocationButton = screen.getByText('Update your location');
+  fireEvent.click(updateLocationButton);
```
nmanu1, line 142: I think it's generally recommended to use `userEvent` rather than `fireEvent` because the former simulates actual user behavior instead of only firing the exact event specified

tests/components/LocationBias.test.tsx
alextaing: changed all button accesses from getByText to getByRole('button' {name: ___ })

tests/components/LocationBias.test.tsx
alextaing: corrected!

tests/components/LocationBias.test.tsx
alextaing: No, I don't think it is needed--  when I was deciding on whether or not to include that line, I was considering that it might be more clear when reading it later since there's also a test for IP that requires a change from default.  Good to know that I should not do that, though!

tests/components/LocationBias.test.tsx
alextaing: Thanks!  I was looking into this, and I also saw that some people rendered the element, changed it to a JSON and checked if it was null-- I thought that was a little weird, but this one makes sense.

tests/components/LocationBias.test.tsx
alextaing: got it-- changed fireEvent to userEvent.

oshi97:
addresses -> address in the PR description

oshi97 (approved):

PR #166, opened 2022-05-31
support react 16.14

oshi97:
Updates answers-headless-react to the alpha version that
announces react 16.14 support.

Update unit tests to run in react 16 in addition to 17.

I tried to upgrade the lib to use react 18 as as a dev dep,
but we still have a few dependencies that don't allow for 18
like storybook/react.
I opened up an issue for that particular one, since I believe it's the only one.
https://github.com/storybookjs/storybook/issues/18349#issuecomment-1142141171

J=SLAP-2113
TEST=manual,auto

run test-site with react 16, check dev tools to see it is using react-dom 16.14
update unit tests to run in react 16


oshi97:
If this PR is approved I will change the required checks under branch protection rules

PR #168, opened 2022-05-31
VisualAutocomplete: change universalLimit and restrictVerticals to simple props

oshi97:
This PR removes automatic detection of universalLimit and restrictVerticals based
on props passed into EntityPreviews.

J=SLAP-2118
TEST=manual

add VA to the test-site
see it display properly

### DIFF
--- a/src/components/SearchBar.tsx
+++ b/src/components/SearchBar.tsx
@@ -2,6 +2,7 @@ import {
   AnswersHeadless,
   QuerySource,
   SearchTypeEnum,
+  UniversalLimit,
   useAnswersActions,
   useAnswersState,
   useAnswersUtilities,
@@ -27,8 +28,6 @@ import { FocusedItemData } from './Dropdown/FocusContext';
 import { CompositionMethod, useComposedCssClasses } from '../hooks/useComposedCssClasses';
 import {
   calculateEntityPreviewsCount,
-  calculateRestrictVerticals,
-  calculateUniversalLimit,
   transformEntityPreviews
 } from './EntityPreviews';
 import { SearchButton } from './SearchButton';
@@ -118,11 +117,15 @@ export type RenderEntityPreviews = (
  */
 export interface VisualAutocompleteConfig {
   /** The Answers Headless instance used to perform visual autocomplete searches. */
-  entityPreviewSearcher?: AnswersHeadless,
+  entityPreviewSearcher: AnswersHeadless,
+  /** Renders entity previews based on the autocomplete loading state and results. */
+  renderEntityPreviews: RenderEntityPreviews,
+  /** Specify which verticals to return for VisualAutocomplete */
+  restrictVerticals: string[],
+  /** Specify the number of entities to return per vertical **/
+  universalLimit?: UniversalLimit,
   /** The debouncing time, in milliseconds, for making API requests for entity previews. */
-  entityPreviewsDebouncingTime?: number,
-  /** Renders entity previeews based on the autocomplete loading state and results. */
-  renderEntityPreviews?: RenderEntityPreviews
+  entityPreviewsDebouncingTime?: number
 }
 
 /**
@@ -171,7 +174,7 @@ export function SearchBar({
   placeholder,
   geolocationOptions,
   hideRecentSearches,
-  visualAutocompleteConfig = {},
+  visualAutocompleteConfig,
   hideVerticalLinks,
   onSelectVerticalLink,
   verticalKeyToLabel,
@@ -183,8 +186,10 @@ export function SearchBar({
   const {
     entityPreviewSearcher,
     renderEntityPreviews,
+    restrictVerticals,
+    universalLimit,
     entityPreviewsDebouncingTime = 500
-  } = visualAutocompleteConfig;
+  } = visualAutocompleteConfig ?? {};
   const answersActions = useAnswersActions();
   const answersUtilities = useAnswersUtilities();
   const reportAnalyticsEvent = useSearchBarAnalytics();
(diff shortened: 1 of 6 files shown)


src/components/SearchBar.tsx
```diff
@@ -118,11 +117,15 @@ export type RenderEntityPreviews = (
  */
 export interface VisualAutocompleteConfig {
   /** The Answers Headless instance used to perform visual autocomplete searches. */
-  entityPreviewSearcher?: AnswersHeadless,
+  entityPreviewSearcher: AnswersHeadless,
+  /** Renders entity previews based on the autocomplete loading state and results. */
+  renderEntityPreviews: RenderEntityPreviews,
+  /** Specify which verticals to return for VisualAutocomplete */
+  restrictVerticals: string[],
+  /** Specify the number of entities to return per vertical **/
```
nmanu1, line 125: nit: missing period here and above

test-site/src/pages/PeoplePage.tsx
```diff
@@ -7,7 +7,7 @@ import {
   StandardCard,
   VerticalResults,
   LocationBias,
-  StaticFilters
+  StaticFilters,
```
nmanu1, line 10: nit: extra comma

test-site/src/pages/PeoplePage.tsx
```diff
@@ -20,7 +20,8 @@ export function PeoplePage() {
 
   return (
     <div>
-      <SearchBar />
+      <SearchBar
```
nmanu1, line 23: nit: this can stay on one line

test-site/src/pages/UniversalPage.tsx
```diff
@@ -1,12 +1,44 @@
...
+
+const visualAutocompleteConfig: VisualAutocompleteConfig = {
+  entityPreviewSearcher: provideAnswersHeadless({
+    ...config,
+    headlessId: 'visual-autocomplete'
+  }),
+  restrictVerticals: ['people'],
+  renderEntityPreviews: isLoading => (
+    <div className={isLoading ? 'opacity-50' : ''}>
+      <EntityPreviews verticalKey='people'>
```
nmanu1, line 23: is it still necessary to wrap in an `EntityPreviews` component? I thought that was only needed to calculate the universal limit and restricted verticals which are being removed

if it is needed, it might be good to update the `renderEntityPreviews` docs to reflect that

test-site/src/pages/UniversalPage.tsx
```diff
@@ -1,12 +1,44 @@
...
+        {results => (
+          <div className='flex ml-4 mt-1'>
+            {results.map((r, index) =>
+              <div
```
nmanu1, line 27: out of curiosity, does the dropdown and arrow navigation work properly like this without wrapping in `DropdownItem`?

oshi97: Hmm I didn't intend for the visual autocomplete results to be navigable here, they were just supposed to be a preview of the universal results. Is it more common for people to want them as navigable?

test-site/src/pages/UniversalPage.tsx
oshi97: You can use EntityPreviews to receive the correct vertical results. I added a comment in the RenderEntityPreviews type and also the prop since maybe this was unclear.

test-site/src/pages/UniversalPage.tsx
nmanu1: Ah ok, not sure which is more common, so as long as they appear in the dropdown properly, that's fine! I just wasn't sure if that was the case because I thought I remembered someone, maybe Aaron, having some issues when they didn't wrap in `DropdownItem`

test-site/src/pages/UniversalPage.tsx
oshi97: Oh to be more clear, it makes sure the portion of universal results for the right vertical key gets passed in. In that sense it's a little magical and it might make sense to get rid of EntityPreviews? We could pass in the entire universal results as a param to renderEntityPreviews

nmanu1: To make sure I'm understanding correctly, if we got rid of `EntityPreviews`, when users defined `renderEntityPreviews`, they would just need to loop through the universal results to find a matching vertical key and use those vertical results, right? If that's the case, I think I'm in favor of removing `EntityPreviews` so there's no more "magic" performed

oshi97: It's gone!

src/components/SearchBar.tsx
```diff
@@ -378,17 +374,13 @@ export function SearchBar({
     );
   }
 
-  const transformedEntityPreviews = entityPreviews
-    && transformEntityPreviews(entityPreviews, verticalResultsArray);
-  const entityPreviewsCount = calculateEntityPreviewsCount(transformedEntityPreviews);
-  const showEntityPreviewsDivider = entityPreviewsCount > 0
```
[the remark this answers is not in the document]
oshi97: I think checking that entityPreviews is a truthy value performs the same function as checking that entityPreviewsCount is greater than 0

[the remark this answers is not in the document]
nmanu1: ah ok. but I think we would still need the exact number for the screenreader announcement?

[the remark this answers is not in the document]
oshi97: The screenreader announcement is gone now. We don't have any notion for what an entity preview even "is" anymore so it's impossible to provide a count for the screen reader.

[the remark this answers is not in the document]
oshi97: post slack disc added it back!

src/hooks/useEntityPreviews.tsx
```diff
@@ -62,5 +65,17 @@ export function useEntityPreviews(
...
+}
+
+/**
+ * @returns a mapping of vertical key to VerticalResults
+ */
+function getVerticalKeyToResults(
+  verticalResultsArray: VerticalResultsData[]
+): Record<string, VerticalResultsData> {
+  return verticalResultsArray.reduce<Record<string, VerticalResultsData>>((prev, current) => {
+    prev[current.verticalKey] = current;
```
nmanu1, line 78: do we want to just store `current.results` so it's just the `Result[]` instead of the whole `VerticalResultsData`?

nmanu1 (approved):

PR #169, opened 2022-06-06
Upgrade Jest to v28

nmanu1:
This PR upgrades Jest from v27 to v28. Most of the breaking changes don't affect us, except for two:
- To use the JSDOM test environment, `jest-environment-jsdom` now has to be installed separately
- Jest now has full support for package exports, which can cause problems where file imports are not resolved correctly. Specifically, this is a problem for `uuid`. There is a discussion and proposed solution in this [GitHub issue](https://github.com/microsoft/accessibility-insights-web/pull/5421#issuecomment-1109168149), which I used in this PR. To summarize, the fix is to add a custom resolver that forces Jest to use the CommonJS+node version of `uuid`, but leaves all other resolutions the same

J=SLAP-2123
TEST=auto

See that Jest tests pass.

### DIFF
--- a/tests/__setup__/resolver.ts
+++ b/tests/__setup__/resolver.ts
@@ -1,0 +1,33 @@
+// This code is from
+// https://github.com/microsoft/accessibility-insights-web/blob/9ad4e618019298d82732d49d00aafb846fb6bac7/src/tests/common/resolver.js
+// and is licensed under the MIT license.
+
+module.exports = (path, options) => {
+  // Call the defaultResolver, so we leverage its cache, error handling, etc.
+  return options.defaultResolver(path, {
+    ...options,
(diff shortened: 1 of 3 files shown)


tests/__setup__/resolver.ts
```diff
@@ -0,0 +1,31 @@
+// This code is from
```
oshi97, line 1: is it possible to use moduleNameMapper instead like suggested here? https://github.com/uuidjs/uuid/pull/616#issuecomment-1114696186
It might be a little easier to find and remove later on

oshi97: It's ridiculous they haven't closed out https://github.com/uuidjs/uuid/pull/616 for 4 months

nmanu1: it is possible to use `moduleNameMapper`, but it seemed like that might have some consequences, discussed here: https://github.com/uuidjs/uuid/pull/616#issuecomment-1111012599

PR #170, opened 2022-06-07
remove cssCompositionMethod and use tailwind-merge lib 

yen-tt:
This pr remove cssCompositionMethod field from all components. Instead, `useComposedCssClasses` hook will use the `tailwind-merge` library to do a deep merge of the builtin and custom style sets.

This PR adds the react hook testing library as a devDep. Once we upgrade the main react testing library (not the hooks
addon) to v13 in order to support react 18, we can delete the hook testing lib as a dep since the latest ver of react testing
lib has absorbed the hooks testing lib. We currently cannot update our react dev dep to 18 yet due to certain
deps not yet supporting react 18.

J=SLAP-2121
TEST=manual,auto

unit tests using the react hook testing library
add cssComposition usage to the test site
tested that the bundle.css still works by removing lib '../lib/**/*.{js,jsx}' from the test-site's tailwind config,
seeing lots of styling go missing, then adding the bundle.css import


tailwind.config.cjs
```diff
@@ -7,7 +7,7 @@ module.exports = {
       colors: {
         'primary': 'var(--primary-color, #2563eb)',
         'primary-light': 'var(--primary-color-light, #dbeafe)',
-        'primary-dark':  'var(--primary-color-dark, #dbeafe)',
+        'primary-dark':  'var(--primary-color-dark, #1e3a8a)',
```
tmeyer2115, line 10: Was this for testing?

src/hooks/useComposedCssClasses.tsx
```diff
@@ -30,47 +8,22 @@ export type CompositionMethod = 'merge' | 'replace' | 'assign';
  *
  * @param builtInClasses - The component's built-in css classes
  * @param customClasses - The custom classes to combine with the built-in ones
- * @param cssCompositionMethod - The method of combining the built-in classes with the custom classes
  * @returns The composed CSS classes
  */
-export function useComposedCssClasses<ClassInterface>(
+export function useComposedCssClasses<
```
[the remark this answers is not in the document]
oshi97: updated! it was pretty vague before. I think supporting multiple composition methods made it hard to write concise but descriptive doc

src/components/Filters/CheckboxOption.tsx
```diff
@@ -78,7 +76,7 @@ export function CheckboxOption(props: CheckboxOptionProps): JSX.Element | null {
     displayName = props.value,
   } = props;
   const cssClasses = useComposedCssClasses(
-    builtInCssClasses, props.customCssClasses, props.cssCompositionMethod);
+    builtInCssClasses, props.customCssClasses);
```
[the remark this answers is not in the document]
oshi97: checked all the other ones too

THIRD-PARTY-NOTICES
```diff
@@ -436,7 +406,7 @@ Apache License
 
 The following NPM package may be included in this product:
 
- - @reduxjs/toolkit@1.8.2
+ - @reduxjs/toolkit@1.8.1
```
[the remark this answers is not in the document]
oshi97: This came from the merge, but was weirdly hard to fix, I just ended up manually installing `answers-headless@alpha` and removing it. I think our versioning scheme might not work perfectly well with npm's semantic versioning. I checked the answers-headless-react alpha and it has the right versions of headless and core, but here in the component lib even specifically installing the latest headless-react alpha wouldn't get me the right versions for the other two deps.

src/hooks/useComposedCssClasses.tsx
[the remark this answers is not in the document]
oshi97: Ah actually I was working off the WIP which just deleted the remarks and example from before, but looking at it again it was pretty in depth

tailwind.config.cjs
```diff
@@ -7,7 +7,7 @@ module.exports = {
       colors: {
         'primary': 'var(--primary-color, #2563eb)',
         'primary-light': 'var(--primary-color-light, #dbeafe)',
-        'primary-dark':  'var(--primary-color-dark, #1e3a8a)',
+        'primary-dark':  'var(--primary-color-dark, #dbeafe)',
```
yen-tt, line 10: I think I made a change here in the wip commit because this is the same color as `primary-light`. So I updated this to use a darker shade from figma for 'primary-dark`.

src/hooks/useComposedCssClasses.tsx
```diff
@@ -22,7 +26,13 @@ export function useComposedCssClasses<
       return mergedCssClasses;
     }
     Object.keys(customClasses).forEach(key => {
-      mergedCssClasses[key] = twMerge(mergedCssClasses[key], customClasses[key]);
+      const builtIn = builtInClasses[key];
+      const custom = customClasses[key];
+      if (!builtIn || !custom) {
+        mergedCssClasses[key] = builtIn ?? custom;
+      } else {
+        mergedCssClasses[key] = twMerge(builtIn, custom);
```
yen-tt, line 49: I haven't finish looking into the documentation for tailwind-merge at the time of the wip commit. Not sure if you saw this section about usage with custom tailwind config: https://github.com/dcastil/tailwind-merge#usage-with-custom-tailwind-config. Since we do have some custom styling defined in tailwind.config.cjs, I want to double check if that would may be handle differently.

tests/hooks/useComposedCssClasses.test.ts
```diff
@@ -0,0 +1,37 @@
+import { useComposedCssClasses } from '../../src/hooks/useComposedCssClasses';
+import { renderHook } from '@testing-library/react-hooks';
+
+describe('when there are no custom classes', () => {
+  const builtInCssClasses = { container: 'block p-1' };
+
+  it('is a pass through', () => {
+    const spy = jest.spyOn(require('tailwind-merge'), 'twMerge');
```
yen-tt, line 8: nit: can we have a single import at the top `import * as  tailwindMerge from 'tailwind-merge'` and do `jest.spyOn(tailwindMerge, 'twMerge');`?

src/hooks/useComposedCssClasses.tsx
```diff
@@ -1,76 +1,54 @@
...
-    return false;
-  }
-  return true;
+    Object.keys(customClasses).forEach(key => {
+      const builtIn = builtInClasses[key];
+      const custom = customClasses[key];
+      if (!builtIn || !custom) {
+        mergedCssClasses[key] = builtIn ?? custom;
+      } else {
+        mergedCssClasses[key] = twMerge(builtIn, custom);
```
yen-tt, line 49: I believe [twMerge ignores undefined/null values](https://github.com/dcastil/tailwind-merge#ignores-undefined-null-and-false-values) so I'm not sure if this check is needed?

oshi97: Oh so I wasn't sure if this was necessary when I wrote it so I looked at the source code to see if it might be worth it and they do call [mergeClasslist](https://github.com/dcastil/tailwind-merge/blob/main/src/lib/merge-classlist.ts) every time twMerge is called and also check and update a cache so I felt like eh may as well? But I can remove if you feel this is a premature optimization

src/hooks/useComposedCssClasses.tsx
oshi97: Ah yeah I missed that I'll check it out

tests/hooks/useComposedCssClasses.test.ts
```diff
@@ -0,0 +1,37 @@
...
+    container: 'px-4 text-red-200 mb-3'
+  });
+});
+
+it('does not call twMerge when builtInClass is a blank string', () => {
+  const spy = jest.spyOn(require('tailwind-merge'), 'twMerge');
+  const builtInCssClasses = { container: '' };
+  const customClasses = { container: 'p-2' };
+  renderHook(() => useComposedCssClasses(builtInCssClasses, customClasses));
+  expect(spy).toHaveBeenCalledTimes(0);
```
nmanu1, line 36: nit: do we also want to add a check that the composed style matches the custom classes?

oshi97: added! I actually missed a case here...

tests/hooks/useComposedCssClasses.test.ts
oshi97: added!

src/hooks/useComposedCssClasses.tsx
oshi97: I added conflict resolution for the form-* classes. Custom colors and what not are automatically resolved by tailwind-merge, it's only new class groups like form-* we need to worry about. We don't need to worry about conflicts with a user's custom class groups, because we don't use any ones besides forms ourselves so there's no way to have a conflict.

tailwind.config.cjs
oshi97: updated with micho's color! also made a react-theme pr

yen-tt (commented):
+1! (PR author can't approve..)

package.json
```diff
@@ -3,12 +3,12 @@
   "version": "0.1.2-alpha.161",
   "description": "",
   "author": "slapshot@yext.com",
-  "main": "./lib/esm/index.js",
-  "types": "./lib/esm/index.d.ts",
+  "main": "./lib/esm/src/index.js",
```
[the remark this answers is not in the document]
oshi97: the baseUrl for the typescript output was moved down a level to be able to include the tailwind.config.js in the output. This results in all other files being nested in a "src" folder. 
This is like what happened when we included the package.json in our other repos for the SDK Agent work, so that we could access the package json version in source code.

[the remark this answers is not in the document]
tmeyer2115: I see. I find that somewhat confusing. When I see a directory called `src` under `lib`, I expect it to be the raw source code. This is something I have seen in other npm packages. But, that's not really the case here. It's transpiled source code (from TS to JS).

tmeyer2115 (approved):

tmeyer2115 (approved):

tmeyer2115 (approved):

PR #171, opened 2022-06-07
Jest tests for FilterSearch component

alextaing:
This PR adds Jest tests for the FilterSearch component.

J=SLAP-2108
TEST=auto

### DIFF
--- a/tests/__fixtures__/data/filtersearch.ts
+++ b/tests/__fixtures__/data/filtersearch.ts
@@ -1,17 +1,20 @@
-import { FilterSearchResponse } from '@yext/answers-headless-react';
+import { FilterSearchResponse, Matcher } from '@yext/answers-headless-react';
 
-export const unsectionedFilterSearchResponse = {
+export const unsectionedFilterSearchResponse: FilterSearchResponse = {
   sections: [
     {
       results: [
         {
-          value: 'first name 1'
+          value: 'first name 1',
+          filter: { fieldId: 'ce_person', matcher: Matcher.Equals, value: 'first name 1' }
         },
         {
-          value: 'first name 2'
+          value: 'first name 2',
+          filter: { fieldId: 'ce_person', matcher: Matcher.Equals, value: 'first name 2' }
         },
         {
-          value: 'last name 1'
+          value: 'last name 1',
+          filter: { fieldId: 'ce_person', matcher: Matcher.Equals, value: 'last name 1' }
         }
       ],
     }
@@ -19,15 +22,17 @@ export const unsectionedFilterSearchResponse = {
   uuid: ''
 };
 
-export const sectionedFilterSearchResponse = {
+export const sectionedFilterSearchResponse: FilterSearchResponse = {
   sections: [
     {
       results: [
         {
-          value: 'first name 1'
+          value: 'first name 1',
+          filter: { fieldId: 'ce_person', matcher: Matcher.Equals, value: 'first name 1' }
         },
         {
-          value: 'first name 2'
+          value: 'first name 2',
+          filter: { fieldId: 'ce_person', matcher: Matcher.Equals, value: 'first name 2' }
         }
       ],
       label: 'First name'
@@ -35,7 +40,8 @@ export const sectionedFilterSearchResponse = {
     {
       results: [
         {
-          value: 'last name 1'
+          value: 'last name 1',
+          filter: { fieldId: 'ce_person', matcher: Matcher.Equals, value: 'last name 1' }
         }
       ],
       label: 'Last name'
@@ -44,39 +50,7 @@ export const sectionedFilterSearchResponse = {
   uuid: ''
 };
 
-export const mockedFilterSearchResponse: FilterSearchResponse = {
-  sections: [{
-    label: 'People',
-    results: [
-      { value: 'John Doe',
-        filter: { fieldId: null, matcher: null, value: 'test' } },
-      { value: 'Jane Doe',
-        filter: { fieldId: null, matcher: null, value: 'test' } }
-    ]
-  },
-  {
-    label: 'Condiments',
-    results: [
-      { value: 'Jam',
-        filter: { fieldId: null, matcher: null, value: 'test' } }
-    ]
-  }],
-  uuid: null
-};
-
-export const mockedFilterSearchResponseNoLabels: FilterSearchResponse = {
-  sections: [{
-    results: [
-      { value: 'John Doe',
-        filter: { fieldId: null, matcher: null, value: 'test' } },
-      { value: 'Jane Doe',
-        filter: { fieldId: null, matcher: null, value: 'test' } }
-    ]
-  }],
-  uuid: null
-};
-
-export const mockedFilterSearchResponseNoResults: FilterSearchResponse = {
+export const noResultsFilterSearchResponse: FilterSearchResponse = {
   sections: [{
     results:[]
   }],
--- a/tests/__fixtures__/data/filtersearch.ts
+++ b/tests/__fixtures__/data/filtersearch.ts
@@ -1,20 +1,20 @@
 import { FilterSearchResponse, Matcher } from '@yext/answers-headless-react';
 
-export const unsectionedFilterSearchResponse: FilterSearchResponse = {
+export const unlabeledFilterSearchResponse: FilterSearchResponse = {
   sections: [
     {
       results: [
         {
           value: 'first name 1',
-          filter: { fieldId: 'ce_person', matcher: Matcher.Equals, value: 'first name 1' }
+          filter: { fieldId: 'name', matcher: Matcher.Equals, value: 'first name 1' }
         },
         {
           value: 'first name 2',
-          filter: { fieldId: 'ce_person', matcher: Matcher.Equals, value: 'first name 2' }
+          filter: { fieldId: 'name', matcher: Matcher.Equals, value: 'first name 2' }
         },
         {
           value: 'last name 1',
-          filter: { fieldId: 'ce_person', matcher: Matcher.Equals, value: 'last name 1' }
+          filter: { fieldId: 'name', matcher: Matcher.Equals, value: 'last name 1' }
         }
       ],
     }
@@ -22,17 +22,17 @@ export const unsectionedFilterSearchResponse: FilterSearchResponse = {
   uuid: ''
 };
 
-export const sectionedFilterSearchResponse: FilterSearchResponse = {
+export const labeledFilterSearchResponse: FilterSearchResponse = {
   sections: [
     {
       results: [
         {
           value: 'first name 1',
-          filter: { fieldId: 'ce_person', matcher: Matcher.Equals, value: 'first name 1' }
+          filter: { fieldId: 'name', matcher: Matcher.Equals, value: 'first name 1' }
         },
         {
           value: 'first name 2',
-          filter: { fieldId: 'ce_person', matcher: Matcher.Equals, value: 'first name 2' }
+          filter: { fieldId: 'name', matcher: Matcher.Equals, value: 'first name 2' }
         }
       ],
       label: 'First name'
@@ -41,7 +41,7 @@ export const sectionedFilterSearchResponse: FilterSearchResponse = {
       results: [
         {
           value: 'last name 1',
-          filter: { fieldId: 'ce_person', matcher: Matcher.Equals, value: 'last name 1' }
+          filter: { fieldId: 'name', matcher: Matcher.Equals, value: 'last name 1' }
         }
       ],
       label: 'Last name'
--- a/tests/__fixtures__/data/filtersearch.ts
+++ b/tests/__fixtures__/data/filtersearch.ts
@@ -1,3 +1,5 @@
+import { FilterSearchResponse } from '@yext/answers-headless-react';
+
 export const unsectionedFilterSearchResponse = {
   sections: [
     {
@@ -41,3 +43,43 @@ export const sectionedFilterSearchResponse = {
   ],
   uuid: ''
 };
+
+export const mockedFilterSearchResponse: FilterSearchResponse = {
+  sections: [{
+    label: 'People',
+    results: [
+      { value: 'John Doe',
+        filter: { fieldId: null, matcher: null, value: 'test' } },
+      { value: 'Jane Doe',
+        filter: { fieldId: null, matcher: null, value: 'test' } }
+    ]
+  },
+  {
+    label: 'Condiments',
+    results: [
+      { value: 'Jam',
+        filter: { fieldId: null, matcher: null, value: 'test' } }
+    ]
+  }],
+  uuid: null
+};
+
+export const mockedFilterSearchResponseNoLabels: FilterSearchResponse = {
+  sections: [{
+    results: [
+      { value: 'John Doe',
+        filter: { fieldId: null, matcher: null, value: 'test' } },
+      { value: 'Jane Doe',
+        filter: { fieldId: null, matcher: null, value: 'test' } }
+    ]
+  }],
+  uuid: null
+};
+
+export const mockedFilterSearchResponseNoResults: FilterSearchResponse = {
+  sections: [{
+    results:[]
+  }],
+  uuid: null
+};
+
--- a/tests/components/FilterSearch.test.tsx
+++ b/tests/components/FilterSearch.test.tsx
@@ -3,7 +3,7 @@ import { render, screen, waitFor } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import * as searchOperations from '../../src/utils/search-operations';
 import { mockAnswersActions, spyOnActions } from '../__utils__/mocks';
-import { sectionedFilterSearchResponse, unsectionedFilterSearchResponse, noResultsFilterSearchResponse } from '../../tests/__fixtures__/data/filtersearch';
+import { labeledFilterSearchResponse, unlabeledFilterSearchResponse, noResultsFilterSearchResponse } from '../../tests/__fixtures__/data/filtersearch';
 import { Matcher } from '@yext/answers-headless-react';
 
 jest.mock('@yext/answers-headless-react');
@@ -13,11 +13,11 @@ const setFilterOption = jest.fn();
 const setOffset = jest.fn();
 const searchFieldsProp = [{
   fieldApiName: 'name',
-  entityType: 'ce-person'
+  entityType: 'ce_person'
 }];
 
 const expectedFilterSearchFields = [{
-  entityType: 'ce-person',
+  entityType: 'ce_person',
   fetchEntities: false,
   fieldApiName: 'name'
 }];
@@ -27,10 +27,8 @@ describe('search with section labels', () => {
     mockAnswersActions({
       setFilterOption,
       setOffset,
-      executeFilterSearch: jest.fn().mockResolvedValue(sectionedFilterSearchResponse)
+      executeFilterSearch: jest.fn().mockResolvedValue(labeledFilterSearchResponse)
     });
-
-    jest.spyOn(searchOperations, 'executeSearch').mockImplementation();
   });
 
   it('renders the filter search bar and the "Filter" label', () => {
@@ -94,28 +92,8 @@ describe('search with section labels', () => {
     userEvent.type(searchBarElement, 'n');
 
     await waitFor(() => {
-      const firstAutocompleteSection= screen.getByText('First name');
-      expect(firstAutocompleteSection).toBeDefined();
-    });
-
-    await waitFor(() => {
-      const firstAutocompleteSuggestion = screen.getByText('first name 1');
-      expect(firstAutocompleteSuggestion).toBeDefined();
-    });
-
-    await waitFor(() => {
-      const secondAutocompleteSuggestion = screen.getByText('first name 2');
-      expect(secondAutocompleteSuggestion).toBeDefined();
-    });
-
-    await waitFor(() => {
-      const secondAutocompleteSection= screen.getByText('Last name');
-      expect(secondAutocompleteSection).toBeDefined();
-    });
-
-    await waitFor(() => {
-      const thirdAutocompleteSuggestion = screen.getByText('last name 1');
-      expect(thirdAutocompleteSuggestion).toBeDefined();
+      const autocompleteSection= screen.getByText('First name');
+      expect(autocompleteSection).toBeDefined();
     });
   });
 
@@ -125,15 +103,7 @@ describe('search with section labels', () => {
     const searchBarElement = screen.getByRole('textbox');
     userEvent.type(searchBarElement, 'n');
 
-    await waitFor(() => {
-      const firstAutocompleteSuggestion = screen.getByText('first name 1');
-      expect(firstAutocompleteSuggestion).toBeDefined();
-    });
-
-    await waitFor(() => {
-      const secondAutocompleteSuggestion = screen.getByText('first name 2');
-      expect(secondAutocompleteSuggestion).toBeDefined();
-    });
+    await waitFor(() => screen.findByText('first name 1'));
 
     userEvent.type(searchBarElement, '{arrowdown}');
     expect(searchBarElement).toHaveValue('first name 1');
@@ -143,22 +113,21 @@ describe('search with section labels', () => {
   });
 
   it('triggers a search on pressing "enter" when an autocomplete result is selected', async () => {
+    const mockExecuteSearch = jest.spyOn(searchOperations, 'executeSearch').mockImplementation();
+
     render (<FilterSearch searchFields={searchFieldsProp}/>);
 
     const searchBarElement = screen.getByRole('textbox');
     userEvent.type(searchBarElement, 'n');
 
-    await waitFor(() => {
-      const firstAutocompleteSuggestion = screen.getByText('first name 1');
-      expect(firstAutocompleteSuggestion).toBeDefined();
-    });
+    await waitFor(() => screen.findByText('first name 1'));
 
     userEvent.type(searchBarElement, '{arrowdown}');
     expect(searchBarElement).toHaveValue('first name 1');
 
     userEvent.type(searchBarElement, '{enter}');
     const expectedSetFilterOptionParam = {
-      fieldId: 'ce_person',
+      fieldId: 'name',
       matcher: Matcher.Equals,
       value: 'first name 1' ,
       displayName: 'first name 1',
@@ -170,12 +139,13 @@ describe('search with section labels', () => {
 
     const setFilterOptionCallOrder = setFilterOption.mock.invocationCallOrder[0];
     const setOffsetCallOrder = setOffset.mock.invocationCallOrder[0];
+    const mockExecuteSearchCallOrder = mockExecuteSearch.mock.invocationCallOrder[0];
     expect(setFilterOptionCallOrder).toBeLessThan(setOffsetCallOrder);
-
-    expect(searchOperations.executeSearch).toBeCalled();
+    expect(setOffsetCallOrder).toBeLessThan(mockExecuteSearchCallOrder);
   });
 
   it('does not trigger a search on pressing "enter" if no autocomplete result is selected', async () => {
+    const mockExecuteSearch = jest.spyOn(searchOperations, 'executeSearch').mockImplementation();
     render (<FilterSearch searchFields={searchFieldsProp}/>);
 
     const searchBarElement = screen.getByRole('textbox');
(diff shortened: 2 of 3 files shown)


tests/components/FilterSearch.test.tsx
```diff
@@ -0,0 +1,299 @@
...
+    expect(mockedExecuteFilterSearch).toBeCalledWith('J');
+
+    userEvent.type(searchBarElement, 'o');
+    expect(searchBarElement).toHaveValue('Jo');
+    expect(mockedExecuteFilterSearch).toBeCalledWith('Jo');
+
+    expect(mockedExecuteFilterSearch.mock.calls.length).toBe(2);
+  });
+
+  it('calls executeFilterSearch when characters changed by backspace', () => {
```
oshi97, line 99: I think we can change this test to be just "backspace on an empty text box should not trigger a search" and remove the part where it types "J" and then backspaces J. The backspace part can be moved to 'calls executeFilterSearch every time a charachter is entered'

tests/components/FilterSearch.test.tsx
```diff
@@ -0,0 +1,299 @@
...
+
+    const label = 'Filter';
+    const labelElement = screen.getByText(label);
+    const searchBarElement = screen.getAllByRole('textbox');
+
+    expect(labelElement).toBeDefined();
+    expect(searchBarElement.length).toBe(1);
+  });
+
+  it('calls executeFilterSearch every time a charachter is entered', () => {
```
oshi97, line 82: charachter

tests/components/FilterSearch.test.tsx
```diff
@@ -0,0 +1,299 @@
...
+    render(<FilterSearch searchFields={[searchFieldsProp]}/>);
+
+    const searchBarElement = screen.getByRole('textbox');
+    expect(searchBarElement).toBeDefined();
```
oshi97, line 86: We probably don't need an `expect(searchBarElement).toBeDefined();` in tests other than `renders the filter search bar and the "Filter" label`. Ideally we don't have multiple tests that have expect() statements on the same "thing" though that's a bit nebulous, and I personallyy lean towards just spamming expect() statements since I'm more worried about missing test cases than being overly redundant, but it's something to keep in mind!

oshi97: Actually I see that below you started to just drop the expect searchBarElement to beDefined already!

oshi97 (commented):
these look great! it's long but FilterSearch has a lot of functionality quirks

tests/components/FilterSearch.test.tsx
```diff
@@ -0,0 +1,299 @@
...
+
+    userEvent.type(searchBarElement, 'J');
+    expect(searchBarElement).toHaveValue('J');
+    expect(mockedExecuteFilterSearch).toBeCalledWith('J');
+
+    userEvent.type(searchBarElement, 'o');
+    expect(searchBarElement).toHaveValue('Jo');
+    expect(mockedExecuteFilterSearch).toBeCalledWith('Jo');
+
+    expect(mockedExecuteFilterSearch.mock.calls.length).toBe(2);
```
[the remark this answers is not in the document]
alextaing: replaced!

tests/components/FilterSearch.test.tsx
alextaing: Ok!  Also, for the tests where a search is called whenever a character is entered, I also changed "toHaveBeenCalledWith" to "toHaveBeenLastCalledWith" to be more specific, since we are checking if executeFilterSearch is called with "J", "Jo", then "J" again.

tests/components/FilterSearch.test.tsx
alextaing: oops, fixed

tests/components/FilterSearch.test.tsx
alextaing: removed it from the unnecessary functions

tests/components/FilterSearch.test.tsx
```diff
@@ -0,0 +1,295 @@
...
+import { FilterSearchResponse } from '@yext/answers-headless-react';
+import { render, screen } from '@testing-library/react';
+import * as useSynchronizedRequestFunctions from '../../src/hooks/useSynchronizedRequest';
+import userEvent from '@testing-library/user-event';
+import * as searchOperations from '../../src/utils/search-operations';
+import { mockAnswersActions } from '../__utils__/mocks';
+
+jest.mock('@yext/answers-headless-react');
+
+const mockedFilterSearchResponse: FilterSearchResponse = {
```
nmanu1, line 11: I think we already have some mocked filter search responses in `__fixtures__/data/filtersearch.ts` so maybe we can consolidate these there

tests/components/FilterSearch.test.tsx
```diff
@@ -0,0 +1,295 @@
...
+    const searchBarElement = screen.getByRole('textbox');
+    userEvent.type(searchBarElement, 'J');
+    expect(mockedExecuteFilterSearch).toBeCalledWith('J');
+
+    const expectedScreenReaderMessage = '2 autocomplete options found.';
+    const screenReaderMessage = screen.getByText(expectedScreenReaderMessage);
+    expect(screenReaderMessage).toBeInTheDocument();
+  });
+
+  it('renders ScreenReader autocomplete result messages even if there is are no autocomplete suggestions', () => {
```
nmanu1, line 275: nit: 'is are'

tests/components/FilterSearch.test.tsx
oshi97: It's imported by the FilterSearch stories

tests/components/FilterSearch.test.tsx
alextaing: oops, fixed!

tests/components/FilterSearch.test.tsx
```diff
@@ -0,0 +1,295 @@
...
+import { mockAnswersActions } from '../__utils__/mocks';
+
+jest.mock('@yext/answers-headless-react');
+
+const mockedFilterSearchResponse: FilterSearchResponse = {
+  sections: [{
+    label: 'People',
+    results: [
+      { value: 'John Doe',
+        filter: { fieldId: null, matcher: null, value: 'test' } },
```
oshi97: In this test data can we use unique filters for each result? Instead of every single one having value 'test'. We should also add a matcher and fieldId to be more realistic.

oshi97: I'm ok not setting uuid to anything since that's currently unused

tests/components/FilterSearch.test.tsx
```diff
@@ -0,0 +1,256 @@
...
+    const searchBarElement = screen.getByRole('textbox');
+    userEvent.type(searchBarElement, 'J');
+
+    const firstAutocompleteSuggestion = screen.getByText('John Doe');
+    expect(firstAutocompleteSuggestion).toBeDefined();
+
+    userEvent.type(searchBarElement, '{arrowdown}');
+    expect(searchBarElement).toHaveValue('John Doe');
+    userEvent.type(searchBarElement, '{enter}');
+    expect(setFilterOption).toBeCalled();
```
oshi97, line 128: we want to make sure setFilterOption and setOffset were called with the right things, specifically (and respectively) the right Filter and 0

tests/components/FilterSearch.test.tsx
```diff
@@ -0,0 +1,256 @@
...
+
+const mockedExecuteFilterSearch = jest.fn();
+const setFilterOption = jest.fn();
+const setOffset = jest.fn();
+const searchFieldsProp = {
+  fieldApiName: 'name',
+  entityType: 'ce-person'
+};
+
+describe('tests using data with section labels', () => {
```
oshi97, line 19: Maybe this title could just be "with section labels"? "test using" is a little redundant

tests/components/FilterSearch.test.tsx
oshi97: it would also be good to check the invocation call order, since we want to make sure setFilterOption and setOffset are both called before executeSearch! You can check this with `setFilterOption.mock.invocationCallOrder` (I only learned you can do this from your pr)

tests/components/FilterSearch.test.tsx
oshi97: Yes for sure! If anything, FilterSearch.stories also wants the filters, they just weren't strictly necessary

tests/components/FilterSearch.test.tsx
```diff
@@ -0,0 +1,256 @@
...
+import * as searchOperations from '../../src/utils/search-operations';
+import { mockAnswersActions } from '../__utils__/mocks';
+import { mockedFilterSearchResponse, mockedFilterSearchResponseNoResults, mockedFilterSearchResponseNoLabels} from '../../tests/__fixtures__/data/filtersearch';
+
+jest.mock('@yext/answers-headless-react');
+
+const mockedExecuteFilterSearch = jest.fn();
+const setFilterOption = jest.fn();
+const setOffset = jest.fn();
+const searchFieldsProp = {
```
yen-tt, line 14: nit: let's make this an array so you don't have to put brackets around in `searchFields={**[**searchFieldsProp**]**}` each time.

tests/components/FilterSearch.test.tsx
```diff
@@ -0,0 +1,256 @@
...
+  beforeEach(() => {
+    mockAnswersActions({
+      setFilterOption,
+      setOffset
+    });
+
+    jest.spyOn(searchOperations, 'executeSearch').mockImplementation();
+    jest.spyOn(useSynchronizedRequestFunctions, 'useSynchronizedRequest').mockImplementation(() => {
+      return [mockedFilterSearchResponse, mockedExecuteFilterSearch, null];
+    });
```
yen-tt, line 29: Instead of mocking `useSynchronizedRequestFunctions`, if possible, I think it may be better if we just directly mock executeFilterSearch in `mockAnswersActions`?

tests/components/FilterSearch.test.tsx
```diff
@@ -0,0 +1,256 @@
...
+
+    userEvent.type(searchBarElement, '{arrowdown}');
+    expect(searchBarElement).toHaveValue('John Doe');
+    userEvent.type(searchBarElement, '{enter}');
+    expect(setFilterOption).toBeCalled();
+    expect(setOffset).toBeCalled();
+    expect(searchOperations.executeSearch).toBeCalled();
+  });
+
+  it('does not trigger executeSearch when "Enter" is pressed if an autocomplete suggestion is not selected with the arrow keys first, even if autocomplete values are returned', () => {
```
yen-tt, line 133: nit: maybe a little wordy, can we shorten the description?

tests/components/FilterSearch.test.tsx
```diff
@@ -0,0 +1,256 @@
...
+
+    userEvent.click(firstAutocompleteSuggestion);
+    expect(setFilterOption).toBeCalled();
+    expect(setOffset).toBeCalled();
+    expect(searchOperations.executeSearch).toBeCalled();
+  });
+});
+
+describe('tests using data without section labels', () => {
+  it('populates the dropdown menu with autocomplete results without section labels if they are not provided', () => {
```
yen-tt, line 167: nit: maybe just `populates the dropdown menu with autocomplete results` since describe() states the data is without section label? I'm also fine if we can combine this and the first `describe` section together

tests/components/FilterSearch.test.tsx
alextaing: refactored the tests to use the old data and added values to the filter fields to the old data

I am unsure about whether the values I added for filter are representative of real data, but I tried to make assumptions based on what I could gather from looking around the code.  I'll leave this unresolved for now and come back to it.

tests/components/FilterSearch.test.tsx
alextaing: added values to the filter fields to the old data, but I am unsure about whether the values I added for filter are representative of real data, but I tried to make assumptions based on what I could gather from looking around the code.  I'll leave this unresolved for now and come back to it.

tests/components/FilterSearch.test.tsx
alextaing: shortened describe names

tests/components/FilterSearch.test.tsx
alextaing: fixed!

tests/components/FilterSearch.test.tsx
alextaing: I originally played around with that idea, but since I needed to also mock a value for filterSearchResponse which is reliant on useSynchronizedSearch, I figured that I would have to mock useSynchronizedSearch anyways. So I just mocked useSynchronizedSearch instead of both useSynchronizedSearch and executeFilterSearch.

Is the direct mocking through AnswersActions preferable?  I guess by mocking useSynchronizedSearch, we miss out on testing the logic inside the function call.

tests/components/FilterSearch.test.tsx
alextaing: shortened description

tests/components/FilterSearch.test.tsx
```diff
@@ -0,0 +1,259 @@
+import { FilterSearch } from '../../src/components/FilterSearch';
+import { render, screen } from '@testing-library/react';
+import * as useSynchronizedRequestFunctions from '../../src/hooks/useSynchronizedRequest';
+import userEvent from '@testing-library/user-event';
+import * as searchOperations from '../../src/utils/search-operations';
+import { mockAnswersActions } from '../__utils__/mocks';
+import { sectionedFilterSearchResponse, unsectionedFilterSearchResponse, noResultsFilterSearchResponse} from '../../tests/__fixtures__/data/filtersearch';
```
oshi97, line 7: can we run eslint? some small nits there

tests/components/FilterSearch.test.tsx
```diff
@@ -0,0 +1,259 @@
...
+
+    userEvent.type(searchBarElement, '{arrowdown}');
+    expect(searchBarElement).toHaveValue('first name 1');
+    userEvent.type(searchBarElement, '{enter}');
+    expect(setFilterOption).toBeCalled();
+    expect(setOffset).toBeCalled();
+    expect(searchOperations.executeSearch).toBeCalled();
+  });
+
+  it('does not trigger executeSearch when "Enter" is pressed if no autocomplete suggestion is selected with the arrow keys first', () => {
```
oshi97, line 133: I think we can remove the "with the arrow keys first", since "selected" already kinda means that. We generally try to keep lines somewhere less than ~100 characters, this line is 139 characters!

oshi97: suuuper long tests are a sign that either 1. the name could be shortened and/or 2. the test is difficult to describe and understand and might be able to be broken up

alextaing: ahh that is good to know.  I'll keep that in mind for the future and get on all the other long lines in this file

tests/components/FilterSearch.test.tsx
```diff
@@ -0,0 +1,259 @@
...
+    const secondAutocompleteSuggestion = screen.getByText('first name 2');
+    expect(secondAutocompleteSuggestion).toBeDefined();
+
+    const thirdAutocompleteSuggestion = screen.getByText('last name 1');
+    expect(thirdAutocompleteSuggestion).toBeDefined();
+  });
+});
+
+describe('screen reader', () => {
+  it('renders ScreenReader autocomplete result messages with section labels if they are provided', () => {
```
oshi97, line 196: this can probably just be something like "renders ScreenReader messages with section labels"

oshi97: I thiiink the other test names can be shortened in a similar fashion

oshi97: You could even remove the "ScreenReader" part of the test name since that's part of the describe block if you really wanted to

tests/components/FilterSearch.test.tsx
```diff
@@ -0,0 +1,259 @@
...
+
+    const searchBarElement = screen.getByRole('textbox');
+
+    userEvent.type(searchBarElement, '{backspace}');
+    expect(searchBarElement).toHaveValue('');
+
+    expect(mockedExecuteFilterSearch).toHaveBeenCalledTimes(0);
+  });
+
+  it('shows autocomplete suggestions when a character is typed and suggestions are returned by useSynchronizedSearch', () => {
```
oshi97, line 74: I'd probably remove the `useSynchronizedSearch` since that's a bit on the implementation detail side of things.

alextaing: just did, one step ahead of ya! haha

tests/components/FilterSearch.test.tsx
alextaing: shortened all titles to under ~100 chars (maybe slightly over)

tests/components/FilterSearch.test.tsx
alextaing: shortened all test names

tests/components/FilterSearch.test.tsx
alextaing: added checks to make sure they were called with the correct parameters and called in the correct order.

tests/components/FilterSearch.test.tsx
yen-tt: Right, I think it's best if we can keep the logic of the hook used in FilterSearch as much as possible and only mock what is necessary. You can mock the return value filterSearchResponse for executeFilterSearch like so:
```
    mockAnswersActions({
      setFilterOption,
      setOffset,
      executeFilterSearch: jest.fn().mockResolvedValue(unsectionedFilterSearchResponse)
    });
```
And to spy on executeFilterSearch in your it() test, you can do something like:
```
    const actions = spyOnActions();
    expect(actions.executeFilterSearch).toHaveBeenCalledWith(blah);
```
You may encounter some warnings about wrapping stuff in act since there's some async request/response happening. So you may need `waitFor` in some of your assertions

tests/components/FilterSearch.test.tsx
```diff
@@ -0,0 +1,274 @@
...
+  });
+
+  it('calls executeFilterSearch every time a character is typed or backspaced', () => {
+    render(<FilterSearch searchFields={searchFieldsProp}/>);
+
+    const searchBarElement = screen.getByRole('textbox');
+
+    userEvent.type(searchBarElement, 'n');
+    expect(searchBarElement).toHaveValue('n');
+    expect(mockedExecuteFilterSearch).toHaveBeenLastCalledWith('n');
```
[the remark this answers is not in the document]
oshi97: I would probably have one test that ensures that typing in FilterSearch displays text in the searchBarElement, and we can remove these expects from all the other tests

tests/components/FilterSearch.test.tsx
oshi97: Could this test just be `"does not trigger search when "enter" is pressed if no option is selected"`?

oshi97: We generally want to avoid specific implementation details like the "executeSearch" function unless they're part of the component's agreed upon interface (in other words the component's contract)

tests/components/FilterSearch.test.tsx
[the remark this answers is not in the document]
tmeyer2115: Sounds good to me

tests/__fixtures__/data/filtersearch.ts
```diff
@@ -1,43 +1,59 @@
-export const unsectionedFilterSearchResponse = {
+import { FilterSearchResponse, Matcher } from '@yext/answers-headless-react';
+
+export const unsectionedFilterSearchResponse: FilterSearchResponse = {
```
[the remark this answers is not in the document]
alextaing: I see-- I think one of the important difference between the data is that one has labeled sections and one has unlabeled sections.  I can for sure rename the objs, but I'm worried that it will mess up for FilterSearch stories since it also uses  this data.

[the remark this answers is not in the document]
oshi97: @alextaing renaming won't mess up the stories since actual object will stay the same.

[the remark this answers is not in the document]
alextaing: I was also going to comment on the boolean value of sectioned-- could we discuss it a little during our check-in?

tests/components/FilterSearch.test.tsx
```diff
@@ -0,0 +1,274 @@
...
+
+    expect(mockedExecuteFilterSearch).toHaveBeenCalledTimes(0);
+  });
+
+  it('shows autocomplete results, if they exist, when a character is typed', () => {
+    render (<FilterSearch searchFields={searchFieldsProp}/>);
+
+    const searchBarElement = screen.getByRole('textbox');
+    userEvent.type(searchBarElement, 'n');
+    expect(mockedExecuteFilterSearch).toBeCalledWith('n');
```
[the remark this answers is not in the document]
alextaing: removed redundancies

tests/components/FilterSearch.test.tsx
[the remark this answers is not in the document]
alextaing: I see-- I went and revised the tests so that they are less redundant when possible.  Definitely good to know.

tests/components/FilterSearch.test.tsx
alextaing: got it, that makes sense.  I changed the names.

tests/components/FilterSearch.test.tsx
alextaing: Got it-- changed to use mockAnswersActions and spyOnActions

tests/components/FilterSearch.test.tsx
```diff
@@ -0,0 +1,298 @@
...
+      expect(firstAutocompleteSuggestion).toBeDefined();
+    });
+
+    await waitFor(() => {
+      const secondAutocompleteSuggestion = screen.getByText('first name 2');
+      expect(secondAutocompleteSuggestion).toBeDefined();
+    });
+
+    await waitFor(() => {
+      const secondAutocompleteSection= screen.getByText('Last name');
```
oshi97, line 112: does eslint point out the missing space before equals? If not we can update our common slapshot eslint sometime later

tests/components/FilterSearch.test.tsx
```diff
@@ -0,0 +1,298 @@
...
+
+  it('triggers a search when an autocomplete result is clicked', async () => {
+    render (<FilterSearch searchFields={searchFieldsProp}/>);
+
+    const searchBarElement = screen.getByRole('textbox');
+    userEvent.type(searchBarElement, 'n');
+
+    let firstAutocompleteSuggestion;
+
+    await waitFor(() => {
```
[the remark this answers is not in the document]
alextaing: Without the waitFors, the get functions can't find the elements that I'm looking for.  I assume it's because when you type in a letter, executeFilterSearch is called and can't return and render the autocomplete suggestions before the get function is called.

tests/components/FilterSearch.test.tsx
```diff
@@ -0,0 +1,254 @@
...
+    expect(searchBarElement.length).toBe(1);
+  });
+
+  it('displays characters typed in search bar correctly',() => {
+    render(<FilterSearch searchFields={searchFieldsProp}/>);
+
+    const searchBarElement = screen.getByRole('textbox');
+
+    userEvent.type(searchBarElement, 'n');
+    expect(searchBarElement).toHaveValue('n');
```
oshi97, line 51: your idea of condensing the "n" and "a" type calls

tests/components/FilterSearch.test.tsx
```diff
@@ -0,0 +1,254 @@
...
+  });
+
+  it('triggers a filter search every time a character is typed or backspaced', () => {
+    render(<FilterSearch searchFields={searchFieldsProp}/>);
+
+    const searchBarElement = screen.getByRole('textbox');
+
+    userEvent.type(searchBarElement, 'n');
+    expect(actions.executeFilterSearch).toHaveBeenLastCalledWith('n', false, expectedFilterSearchFields);
```
oshi97, line 66: can probably do a similar thing here, but up to you

tests/components/FilterSearch.test.tsx
```diff
@@ -0,0 +1,254 @@
...
+  });
+
+  it('does not trigger a filter search when backspacing in an empty text box', () => {
+    render(<FilterSearch searchFields={searchFieldsProp}/>);
+
+    const searchBarElement = screen.getByRole('textbox');
+
+    userEvent.type(searchBarElement, '{backspace}');
+    expect(searchBarElement).toHaveValue('');
+
```
oshi97, line 84: maybe remove this line

tests/components/FilterSearch.test.tsx
```diff
@@ -0,0 +1,254 @@
...
+    userEvent.type(searchBarElement, 'a');
+    expect(actions.executeFilterSearch).toHaveBeenLastCalledWith('na', false, expectedFilterSearchFields);
+
+    userEvent.type(searchBarElement, '{backspace}');
+    expect(actions.executeFilterSearch).toHaveBeenLastCalledWith('n', false, expectedFilterSearchFields);
+
+    expect(actions.executeFilterSearch).toHaveBeenCalledTimes(3);
+  });
+
+  it('does not trigger a filter search when backspacing in an empty text box', () => {
```
oshi97, line 77: remove line breaks like we disc

tests/components/FilterSearch.test.tsx
```diff
@@ -0,0 +1,254 @@
...
+
+    userEvent.type(searchBarElement, '{backspace}');
+    expect(searchBarElement).toHaveValue('');
+
+    expect(actions.executeFilterSearch).toHaveBeenCalledTimes(0);
+  });
+
+  it('shows autocomplete results, if they exist, when a character is typed', async () => {
+    render (<FilterSearch searchFields={searchFieldsProp}/>);
+
```
oshi97, line 90: maybe this line break too

tests/components/FilterSearch.test.tsx
```diff
@@ -0,0 +1,254 @@
...
+    expect(searchBarElement).toHaveValue('first name 1');
+
+    userEvent.type(searchBarElement, '{arrowdown}');
+    expect(searchBarElement).toHaveValue('first name 2');
+  });
+
+  it('triggers a search on pressing "enter" when an autocomplete result is selected', async () => {
+    const mockExecuteSearch = jest.spyOn(searchOperations, 'executeSearch').mockImplementation();
+
+    render (<FilterSearch searchFields={searchFieldsProp}/>);
```
oshi97, line 118: extra space

tests/components/FilterSearch.test.tsx
```diff
@@ -0,0 +1,254 @@
...
+
+    const searchBarElement = screen.getByRole('textbox');
+    userEvent.type(searchBarElement, 'n');
+
+    await waitFor(() => screen.findByText('first name 1'));
+
+    userEvent.type(searchBarElement, '{arrowdown}');
+    expect(searchBarElement).toHaveValue('first name 1');
```
oshi97, line 126: remove expect

tests/components/FilterSearch.test.tsx
```diff
@@ -0,0 +1,298 @@
...
+      displayName: 'first name 1',
+      selected: true
+    };
+    const expectedSetOffsetParam = 0;
+    expect(setFilterOption).toBeCalledWith(expectedSetFilterOptionParam);
+    expect(setOffset).toBeCalledWith(expectedSetOffsetParam);
+
+    const setFilterOptionCallOrder = setFilterOption.mock.invocationCallOrder[0];
+    const setOffsetCallOrder = setOffset.mock.invocationCallOrder[0];
+    expect(setFilterOptionCallOrder).toBeLessThan(setOffsetCallOrder);
```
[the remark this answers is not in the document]
oshi97: compare to mockExecuteSearchCallOrder instead of setOffsetCallOrder

tests/components/FilterSearch.test.tsx
```diff
@@ -0,0 +1,254 @@
...
+    const mockExecuteSearch = jest.spyOn(searchOperations, 'executeSearch').mockImplementation();
+    render (<FilterSearch searchFields={searchFieldsProp}/>);
+
+    const searchBarElement = screen.getByRole('textbox');
+    userEvent.type(searchBarElement, 'n');
+
+    const autocompleteSuggestion = await waitFor(() => screen.findByText('first name 1'));
+
+    userEvent.click(autocompleteSuggestion);
+    expect(setFilterOption).toBeCalled();
```
oshi97, line 169: I would just copy paste all the
```ts
const setFilterOptionCallOrder = setFilterOption.mock.invocationCallOrder[0];
    const setOffsetCallOrder = setOffset.mock.invocationCallOrder[0];
    const mockExecuteSearchCallOrder = mockExecuteSearch.mock.invocationCallOrder[0];
    expect(setFilterOptionCallOrder).toBeLessThan(setOffsetCallOrder);
// ...
```
stuff from 'triggers a search on pressing "enter" when an autocomplete result is selected'

tests/components/FilterSearch.test.tsx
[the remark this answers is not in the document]
alextaing: fixed

tests/components/FilterSearch.test.tsx
alextaing: combined

tests/components/FilterSearch.test.tsx
alextaing: removed excess line breaks

tests/components/FilterSearch.test.tsx
alextaing: removed excess line breaks

tests/components/FilterSearch.test.tsx
alextaing: checks that corrected parameters were passed, and that both functions were called before executeSearch (but without comparison with each other)

tests/components/FilterSearch.test.tsx
```diff
@@ -0,0 +1,254 @@
...
+
+    render(<FilterSearch searchFields={searchFieldsProp}/>);
+
+    const searchBarElement = screen.getByRole('textbox');
+    userEvent.type(searchBarElement, 'n');
+
+    const expectedScreenReaderMessage = '2 First name autocomplete options found. 1 Last name autocomplete option found.';
+    await waitFor(() => {
+      const screenReaderMessage = screen.getByText(expectedScreenReaderMessage);
+      expect(screenReaderMessage).toBeInTheDocument();
```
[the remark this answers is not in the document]
alextaing: changed all toBeInTheDocument to toBeDefined

tests/components/FilterSearch.test.tsx
alextaing: removed excess expect

tests/components/FilterSearch.test.tsx
```diff
@@ -0,0 +1,253 @@
...
+    const searchBarElement = screen.getByRole('textbox');
+
+    userEvent.type(searchBarElement, 'n{enter}');
+
+    await waitFor(() => {
+      expect(setFilterOption).not.toBeCalled();
+    });
+    expect(setOffset).not.toBeCalled();
+    expect(mockExecuteSearch).not.toBeCalled();
+
```
oshi97, line 147: extra line break

alextaing: removed

oshi97 (approved):

PR #172, opened 2022-06-09
React 16.8 build

nmanu1:
Create a React 16.8-compatible build of the components. Jest tests had many errors, but since this is a one-off build, they were ignored and the test command was changed to no longer run the Jest tests. The Storybook site still works, although without some of the add-ons like controls. I removed the code coverage and Percy snapshots workflows because they had errors.

The only real change needed for the components was in the `Dropdown`. `@restart/ui` does not support React 16.8, so a listener was added to close the dropdown when the user clicks outside the component or presses the 'Escape' key.

I also made a tweak to the `SearchBar` component to fix the screen reader announcements so that on vertical pages, the recent searches aren't announced since they aren't displayed in the dropdown.

J=SLAP-2130
TEST=manual

Clicked through the test-site and made sure all the components functioned as expected.

Notes:
- The Snyk license check is failing because of a new license issue. But, Unlicense is already approved by legal for libraries used for testing, as long as the code is not distributed by our code. This is the case here because the transitive dependency is introduced through `@storybook/react`, so this license can be ignored.
- The failing Snyk security check can also be ignored. It's caused because of a new vulnerability introduced in a transitive dependency of `@storybook/react`, and isn't from our distributed code.


src/components/SearchBar.tsx
```diff
@@ -383,7 +383,7 @@ export function SearchBar({
     || (!isVertical && filteredRecentSearches?.length) || entityPreviews);
   const screenReaderText = getScreenReaderText(
     autocompleteResponse?.results.length,
-    filteredRecentSearches?.length,
+    isVertical ? 0 : filteredRecentSearches?.length,
     entityPreviewsCount
```
tmeyer2115, line 387: Do we similarly need to guard the `entityPreviewsCount` as well? That doesn't appear on vertical searches iirc.

tsconfig.base.json
```diff
@@ -9,7 +9,7 @@
     "declaration": true,
     "declarationMap": true,
     "sourceMap": true,
-    "jsx": "react-jsx",
+    "jsx": "react",
```
tmeyer2115, line 12: For my own edification, what's the difference here?

src/components/SearchBar.tsx
nmanu1: We don't need a guard for `entityPreviewsCount` because there is no `renderEntityPreviews` function provided to the `SearchBar` unless VA is enabled (i.e. only on the universal page), so no `entityPreviews` are generated and the count will be 0 on vertical pages.

src/components/Dropdown/Dropdown.tsx
```diff
@@ -73,9 +72,14 @@ export function Dropdown(props: PropsWithChildren<DropdownProps>): JSX.Element {
     }
   }, [parentQuery, lastTypedOrSubmittedValue, updateFocusedItem, setLastTypedOrSubmittedValue]);
 
-  useRootClose(containerRef, () => {
-    toggleDropdown(false);
-  }, { disabled: !isActive });
+  useGlobalListener('click', e => {
+    if (isActive) {
```
oshi97, line 76: useRootClose also closes the Dropdown when somebody hits the ESC key. I played around with it a bit and it feels a bit weird for the SearchBar to close on ESC but it's also what google does so 🤷 I just wanted to point this out in case people have other ideas.

tsconfig.base.json
nmanu1: The new JSX transform is only introduced in newer versions of React. Setting `react-jsx` here causes errors because the compiler tries to use `react/jsx-runtime` to do the JSX transform, which isn't available for React 16.8. Changing this to `react` tells the compiler to not use that transform and just convert JSX to `React.createElement` calls

src/components/Dropdown/Dropdown.tsx
oshi97: OH I missed that you added it below (and also have it in your commit msg lol)

PR #178, opened 2022-06-13
useRef to return stable function references in UseSynchronizedRequest

alextaing:
This PR addresses useSynchronizedRequest used in the FilterSearch and SearchBar component.  Before, useSynchronizedRequest returned new executeRequest, handleRejectedPromise, and clearResponseData functions on each call.  This change uses refs to return stable references to these functions to reduce re-rendering.

J=SLAP-2132
TEST=auto


oshi97:
you also did manual testing you can add to your pr description!

tests/hooks/useSynchronizedRequest.test.tsx
```diff
@@ -0,0 +1,26 @@
+import { useSynchronizedRequest } from '../../src/hooks/useSynchronizedRequest';
+import { renderHook } from '@testing-library/react-hooks';
+import { waitFor } from '@testing-library/react';
+
+
+let requestFunction = async () => 0;
```
oshi97, line 6: could you move these into the it block?

tests/hooks/useSynchronizedRequest.test.tsx
```diff
@@ -0,0 +1,26 @@
...
+const errorFunction = () => 'error';
+
+it('returns an updated function with the same reference', async () => {
```
oshi97, line 9: I know I said one test, but looking at it again maybe we could also repeat this test for errorFunction?

tests/hooks/useSynchronizedRequest.test.tsx
```diff
@@ -0,0 +1,26 @@
...
+  const { result, rerender } = renderHook(() => useSynchronizedRequest(requestFunction, errorFunction));
```
oshi97, line 10: in this particular test I would probably omit errorFunction since it's not being used by anything

tests/hooks/useSynchronizedRequest.test.tsx
```diff
@@ -0,0 +1,26 @@
+import { useSynchronizedRequest } from '../../src/hooks/useSynchronizedRequest';
+import { renderHook } from '@testing-library/react-hooks';
+import { waitFor } from '@testing-library/react';
+
```
oshi97, line 4: extra line

tests/hooks/useSynchronizedRequest.test.tsx
alextaing: looking back, was there any benefit to using useRef on handleRejectedPromise?  (i.e. error function stuff) since it's not returned by `useSynchronizedRequest`?

tests/hooks/useSynchronizedRequest.test.tsx
alextaing: got it!

tests/hooks/useSynchronizedRequest.test.tsx
alextaing: Done!  I made a test that checks: if you replace the error function and re-render, then the error function will change, but the `executeFunction` returned will be the same.

tests/hooks/useSynchronizedRequest.test.tsx
alextaing: removed

tests/hooks/useSynchronizedRequest.test.tsx
alextaing: removed

oshi97 (approved):
looks great!

PR #179, opened 2022-06-13
add searchOnSelect to FilterSearch

oshi97:
Defaults to false. When set to true, will run a search when a filter is selected.

J=SLAP-2139
TEST=manual,auto

in the test site, see that it defaults to false and will not
execute a search on selection

update older unit tests, add new one


src/components/FilterSearch.tsx
```diff
@@ -93,9 +96,11 @@ export function FilterSearch({
     if (filter && displayName) {
       answersActions.setFilterOption({ ...filter, displayName, selected: true });
       answersActions.setOffset(0);
-      executeSearch(answersActions);
+      if (searchOnSelect) {
```
nmanu1, line 99: in the case where `searchOnSelect` is false, how will the user know which filter will be applied when they click an apply/search button? currently, the search input will update to show the filter value when it's selected. but, the user could start typing in the input again to look for other filters. in this case, should the previously selected filter be removed from headless? or do we want to display it in some way so the user knows what has already been selected and will be applied to the next search?

src/components/FilterSearch.tsx
```diff
@@ -93,9 +96,11 @@ export function FilterSearch({
...
+        executeSearch(answersActions);
```
nmanu1, line 100: for `searchOnSelect=false`, would people use the `ApplyFiltersButton` to execute the search like they do for facets and static filters?

src/components/FilterSearch.tsx
oshi97: Product is ok with this! They referenced zocdoc as having this exact behavior and it feeels good to me? I guess because in my brain if you have already selected a filter you're unlikely to go back and select a different one. Even if you did you probably wouldn't type a different value, then decide not to select a new value, and THEN hit the search button. Even if you did it doesn't seem terribly jank?
https://yext.slack.com/archives/C032CKFARGS/p1655220249371579

tests/components/FilterSearch.test.tsx
```diff
@@ -104,74 +104,90 @@ describe('search with section labels', () => {
...
-    const setFilterOptionCallOrder = setFilterOption.mock.invocationCallOrder[0];
-    const setOffsetCallOrder = setOffset.mock.invocationCallOrder[0];
-    const mockExecuteSearchCallOrder = mockExecuteSearch.mock.invocationCallOrder[0];
-    expect(setFilterOptionCallOrder).toBeLessThan(mockExecuteSearchCallOrder);
-    expect(setOffsetCallOrder).toBeLessThan(mockExecuteSearchCallOrder);
+      userEvent.type(searchBarElement, 'n');
+      await waitFor(() => screen.findByText('first name 1'));
+
+      userEvent.type(searchBarElement, '{arrowdown}{enter}');
+      expect(mockExecuteSearch).not.toHaveBeenCalled();
```
nmanu1, line 189: can we add a check that `setFilterOption` is still called?

oshi97: updated!

nmanu1 (commented):
+1 (will approve after merge conflicts are resolved)

nmanu1 (approved):

PR #182, opened 2022-06-15
Jest tests for ResultsCount component

juliannzhou:
This PR adds Jest tests for the ResultsCount component.

J=SLAP-2137
TEST=auto

### DIFF
--- a/tests/components/ResultsCount.test.tsx
+++ b/tests/components/ResultsCount.test.tsx
@@ -4,13 +4,17 @@ import { ResultsCount } from '../../src/components/ResultsCount';
 import { mockAnswersState } from '../__utils__/mocks';
 import { RecursivePartial } from '../__utils__/mocks';
 
-const mockedStateUniversalMultiple: RecursivePartial<State> = {
+const mockedUniversalState: Partial<State> = {
   searchStatus: {
     isLoading: false
   },
   meta: {
     searchType: 'universal'
-  },
+  }
+};
+
+const mockedStateUniversalMultiple: RecursivePartial<State> = {
+  ...mockedUniversalState,
   universal: {
     verticals: [{
       resultsCount: 2,
@@ -18,69 +22,53 @@ const mockedStateUniversalMultiple: RecursivePartial<State> = {
 
     {
       resultsCount: 3,
-    }
-    ] }
+    }]
+  }
 };
 
 const mockedStateUniversalSingle: RecursivePartial<State> = {
-  searchStatus: {
-    isLoading: false
-  },
-  meta: {
-    searchType: 'universal'
-  },
+  ...mockedUniversalState,
   universal: {
     verticals: [{
       resultsCount: 1,
-    }
-    ] }
+    }]
+  }
 };
 
 const mockedStateUniversalNoResult: RecursivePartial<State> = {
-  searchStatus: {
-    isLoading: false
-  },
-  meta: {
-    searchType: 'universal'
-  },
+  ...mockedUniversalState,
   universal: {
     verticals: [{
       resultsCount: 0,
-    }
-    ] }
+    }]
+  }
 };
 
-const mockedStateVerticalMultiple: Partial<State> = {
+const mockedVerticalState: Partial<State> = {
   searchStatus: {
     isLoading: false
   },
   meta: {
     searchType: 'vertical'
-  },
+  }
+};
+
+const mockedStateVerticalMultiple: Partial<State> = {
+  ...mockedVerticalState,
   vertical: {
     resultsCount: 2
   }
 };
 
 const mockedStateVerticalSingle: Partial<State> = {
-  searchStatus: {
-    isLoading: false
-  },
-  meta: {
-    searchType: 'vertical'
-  },
+  ...mockedVerticalState,
   vertical: {
     resultsCount: 1
   }
 };
 
 const mockedStateVerticalNoResult: Partial<State> = {
-  searchStatus: {
-    isLoading: false
-  },
-  meta: {
-    searchType: 'vertical'
-  },
+  ...mockedVerticalState,
   vertical: {
     resultsCount: 0
   }
@@ -131,5 +119,3 @@ describe('Results count for universal search', () => {
     expect(screen.queryByText('0 Results')).toBeNull();
   });
 });
-
-
--- a/tests/components/ResultsCount.test.tsx
+++ b/tests/components/ResultsCount.test.tsx
@@ -19,7 +19,6 @@ const mockedStateUniversalMultiple: RecursivePartial<State> = {
     verticals: [{
       resultsCount: 2,
     },
-
     {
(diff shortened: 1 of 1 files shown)


tests/components/ResultsCount.test.tsx
```diff
@@ -0,0 +1,97 @@
+import React from 'react';
```
yen-tt, line 1: I think this is an unused import statement, can we remove it?

tests/components/ResultsCount.test.tsx
```diff
@@ -0,0 +1,97 @@
...
+  universal: {
+    verticals: [{
+      appliedQueryFilters: [],
+      queryDurationMillis: 2,
+      results: [],
+      resultsCount: 2,
+      source: Source.Google,
+      verticalKey: 'string'
+    },
+
```
yen-tt, line 24: remove new lines! (eslint did show errors here?)

tests/components/ResultsCount.test.tsx
```diff
@@ -0,0 +1,97 @@
...
+      expect(screen.getByText(expectedResultsCountNumber + ' Results')).toBeDefined();
+    });
+  });
+
+  describe('ResultCountUniversal', () => {
+    beforeEach(() => {
+      mockAnswersState(mockedStateUniversal);
+    });
+
+    it('Results count for vertical search is displayed correctly', () => {
```
yen-tt, line 81: description should be for universal search

tests/components/ResultsCount.test.tsx
```diff
@@ -0,0 +1,97 @@
...
+  },
+  vertical: {
+    resultsCount: 0
+  }
+};
+
+jest.mock('@yext/answers-headless-react');
+
+describe('ResultCounts', () => {
+  describe('ResultCountVertical', () => {
```
yen-tt, line 64: nit: can we update the description to not be in camel case? maybe something like 'Results count for vertical results'. Similarly for the describe block for universal

tests/components/ResultsCount.test.tsx
oshi97: alex also didn't see eslint complain here, feel free to make a note here!

oshi97: https://yexttest.atlassian.net/browse/SLAP-2135

oshi97: I'm not sure if there's a rule for this though

tests/components/ResultsCount.test.tsx
```diff
@@ -0,0 +1,94 @@
...
+    searchType: 'vertical'
+  },
+  vertical: {
+    resultsCount: 0
+  }
+};
+
+jest.mock('@yext/answers-headless-react');
+
+describe('ResultsCount', () => {
```
alextaing, line 61: Since all these tests will be for ResultCount, this outer describe block might not be necessary.  Depending on if you plan to add more tests to the inner describe blocks, they could also be removed since the test names are already give the same information

tests/components/ResultsCount.test.tsx
```diff
@@ -0,0 +1,94 @@
...
+    });
+  });
+
+  it('Renders nothing if there is no results' , () => {
+    mockAnswersState(mockedStateNoResult);
+    render(<ResultsCount />);
+    const expectedResultsCountNumber = mockedStateNoResult.vertical.resultsCount;
+    expect(screen.queryByText(expectedResultsCountNumber + ' Results')).toBeNull();
+  });
+});
```
alextaing, line 94: I'm not sure about this-- but will we also want to test if it renders the correct pluralization of "result" if it returns just 1 result?  I see that we use an imported function `processTranslation` that is alone it its own file.  Should we leave that testing to another test suite, or would it be good to include in this as well?

tests/components/ResultsCount.test.tsx
yen-tt: updated the item to include this example! I'm also not sure if there's one but it's worth doing a quick scan through their rules about lists and objects styling

tests/components/ResultsCount.test.tsx
```diff
@@ -0,0 +1,135 @@
...
+    isLoading: false
+  },
+  meta: {
+    searchType: 'universal'
+  },
+  universal: {
+    verticals: [{
+      resultsCount: 1,
+    }
+    ] }
```
yen-tt, line 36: nit: styling is a bit weird here, can we fix it to: 
```
universal: {
    verticals: [{
      resultsCount: 1,
    }]
}
```

tests/components/ResultsCount.test.tsx
```diff
@@ -0,0 +1,135 @@
...
+  },
+  universal: {
+    verticals: [{
+      resultsCount: 2,
+    },
+
+    {
+      resultsCount: 3,
+    }
+    ] }
```
yen-tt, line 22: nit: let's update the styling here and in other spots similar to this as well

tests/components/ResultsCount.test.tsx
```diff
@@ -0,0 +1,135 @@
...
+  meta: {
+    searchType: 'vertical'
+  },
+  vertical: {
+    resultsCount: 2
+  }
+};
+
+const mockedStateVerticalSingle: Partial<State> = {
+  searchStatus: {
```
yen-tt, line 66: i'm thinking we could probably do something like:
```
const mockedVerticalState: Partial<State> = {
searchStatus: {
    isLoading: false
  },
  meta: {
    searchType: 'vertical'
  }
}
```
and then spread that into mockedStateVerticalSingle, mockedStateVerticalMultiple, and mockedStateVerticalNoResult so we can avoid duplicating searchStatus and meta for all these states. Similarly for universal ones.
```
const mockedStateVerticalSingle: Partial<State> = {
  ...mockedVerticalState,
  vertical: {
    resultsCount: 1
  }
}
```

tests/components/ResultsCount.test.tsx
```diff
@@ -0,0 +1,135 @@
...
+  });
+
+  it('Renders nothing if there is no result', () => {
+    mockAnswersState(mockedStateUniversalNoResult);
+    render(<ResultsCount />);
+    expect(screen.queryByText('0 Results')).toBeNull();
+  });
+});
+
+
```
nmanu1, line 135: nit: the extra newlines here can be removed

yen-tt (approved):

tests/components/ResultsCount.test.tsx
```diff
@@ -0,0 +1,121 @@
...
+  }
+};
+
+const mockedStateUniversalMultiple: RecursivePartial<State> = {
+  ...mockedUniversalState,
+  universal: {
+    verticals: [{
+      resultsCount: 2,
+    },
+
```
oshi97, line 22: oh we still have the extra line here

PR #132, opened 2022-05-02
Location Bias Loading Indicator

cea2aj:

PR #184, opened 2022-06-16
Update version to 0.2.0-beta.184

nmanu1:
### Features
- Added a loading indicator for `LocationBias` (#132)
- Export a standalone `Pagination` component (#147)
- Added controls to the Storybook site (#153)
- Added support for React 16.14 and React 18 (#161, #166)
- Added `placeholder` and `searchOnSelect` props to `FilterSearch` (#177, #179)
### Changes
- Made `StaticFilters`, `StandardFacets`, `NumericalFacets`, and `HierarchicalFacets` components with baked-in layouts and stopped exporting the `Filters` compound components. `ApplyFiltersButton` is now exported directly and works with the new filters components (#156, #160)
- Removed `validateData` now that `PropTypes` are used instead (#164)
- Updated Visual Autocomplete to delete `EntityPreviews`. The restricted verticals and universal limit must be passed into the `VisualAutocompleteConfig` and are no longer computed automatically (#168)
- Removed the `cssCompositionMethod` from all components. Now, a deep merge of the built-in and custom styles will always be performed (#170)
### Bugfixes
- Updated `FilterSearch` to remove the previous filter value when a new one is selected (#185)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@yext/answers-react-components",
-  "version": "0.1.2-beta.184",
+  "version": "0.2.0-beta.184",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@yext/answers-react-components",
-      "version": "0.1.2-beta.184",
+      "version": "0.2.0-beta.184",
       "dependencies": {
         "@microsoft/api-documenter": "^7.15.3",
         "@microsoft/api-extractor": "^7.19.4",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@yext/answers-react-components",
-  "version": "0.1.2-beta.184",
+  "version": "0.2.0-beta.184",
   "description": "",
   "author": "slapshot@yext.com",
   "main": "./lib/esm/index.js",
--- a/test-site/package-lock.json
+++ b/test-site/package-lock.json
@@ -27,7 +27,7 @@
     },
     "..": {
       "name": "@yext/answers-react-components",
-      "version": "0.1.2-beta.184",
+      "version": "0.2.0-beta.184",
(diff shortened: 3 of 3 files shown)


oshi97:
I think I would prefer to not merge this directly into main, but just have it be a tag branching off of main

oshi97:
Actually hmmm I know we said betas are for product but I don't see the distinction between alpha and beta here? I was thinking alpha would be the main branch and betas are special feature branches off of main.

nmanu1:
I thought beta indicates that we've completed our own testing and it might include multiple new features, unlike alpha which would be for testing a single feature?

oshi97:
So far alpha has meant "here is a random point on our actively developed branch", which doesn't mean internal testing for a single feature. I think beta being a more official set of features like you said makes sense though! We might want to formalize release notes and also add it to the releases page in that case. 
I was thinking beta would mean one-off special "big" features that we have in a feature branch, for example something like hierarchical filters or the search loading indicator, which are isolated off of the main development branch. But I think it also makes sense for beta to mean "more official alpha". If we do that instead, the problem is that we won't have any way to designate those "special feature branches" - we wouldn't want to use beta because that would be conflating the two. In that case, maybe we just arbitrarily add additional tags, for instance maybe the version would be `v1.1.1-loadingIndicator.153` or something for a loading indicator feature. Does this kind of make sense? Want to make sure I'm not going off the rails here.

oshi97:
I think it would be nice if this commit's description had the release notes, since we need them either way.

nmanu1:
Updated the description with the release notes! Since we have some breaking changes and additional functionality in here, maybe we should change the version to 0.2.0-beta.184. Thoughts?

oshi97:
sure 0.2.0 sgtm!

oshi97 (approved):

PR #186, opened 2022-06-16
Extract pagination Jest tests

alextaing:
This PR separated pagination Jest tests from the verticalResults Jest tests.  Since they shared data, a fixture was made containing the data, called `verticalResults.ts`.  There was another fixture with a similar name, `verticalnoresults.ts`, which was used in the alternativeVerticals test.  This was renamed `noresults.ts` for clairity.  Tests for the `paginateAllOnNoResults` prop were also added. 

J=SLAP-2148
TEST=auto


nmanu1 (commented):
> There was another fixture with a similar name, verticalnoresults.ts, which was used in the alternativeVerticals test. This was renamed alternativenoresults.ts for clairity.

`alternativenoresults.ts` sounds a little odd to me. if it's only used to test `AlternativeVerticals`, maybe we can rename it to `alternativeverticals.ts`? or just `noresults.ts`

tests/components/Pagination.test.tsx
```diff
@@ -0,0 +1,148 @@
...
+  };
+  mockVerticalSearchState(mockedVerticalSearchState);
+  const actions = spyOnActions();
+  render(<VerticalResults {...verticalResultsProps} />);
+
+  const paginationNavEl = screen.getByRole('navigation', { name: 'Pagination' });
+  expect(paginationNavEl).toBeDefined();
+
+  // navigate to the last results page
+  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
yen-tt, line 136: this is because of the "!" usage? Can we instead have something like `const mockedResultsCount = 5` and use that variable in `mockedVerticalSearchState` for resultsCount in line 123 and in the lines below?

tests/__fixtures__/data/verticalresults.ts
```diff
@@ -0,0 +1,44 @@
...
+    linkType: 'link'
+  },
+  c_secondaryCTA: {
+    link: 'link2',
+    label: 'job2',
+    linkType: 'link'
+  }
+};
+
+export const mockedVerticalResults = [
```
yen-tt, line 16: can we add type for this?

tests/components/Pagination.test.tsx
```diff
@@ -0,0 +1,148 @@
...
+  executeVerticalQuery: jest.fn()
+};
+
+jest.mock('@yext/answers-headless-react');
+
+beforeEach(() => {
+  mockAnswersHooks({ mockedState, mockedActions });
+});
+
+it('Don\'t display pagination component when allowPagination is false', () => {
```
nmanu1, line 37: nit: I think the description here and below reads a little better with "doesn't" instead of "don't"

tests/components/Pagination.test.tsx
```diff
@@ -0,0 +1,148 @@
...
+  const verticalResultsProps: VerticalResultsProps = {
+    CardComponent: StandardCard,
+    allowPagination: false
+  };
+  render(<VerticalResults {...verticalResultsProps} />);
+  const paginationNavEl = screen.queryByRole('navigation', { name: 'Pagination' });
+  expect(paginationNavEl).toBeNull();
+});
+
+it('Don\'t display pagination component when there\'s no results', () => {
```
nmanu1, line 47: nit: "there are" instead of "there's"

tests/components/Pagination.test.tsx
oshi97: +1 this eslint disable should not be necessary

tests/components/Pagination.test.tsx
```diff
@@ -0,0 +1,148 @@
...
+beforeEach(() => {
+  mockAnswersHooks({ mockedState, mockedActions });
+});
+
+it('Don\'t display pagination component when allowPagination is false', () => {
+  const verticalResultsProps: VerticalResultsProps = {
+    CardComponent: StandardCard,
+    allowPagination: false
+  };
+  render(<VerticalResults {...verticalResultsProps} />);
```
oshi97, line 42: We should be rendering the `Pagination` component, not `VerticalResults`

tests/components/Pagination.test.tsx
alextaing: Fixed!

tests/__fixtures__/data/verticalresults.ts
alextaing: type added!

tests/components/Pagination.test.tsx
alextaing: renamed tests for fit 'it' sentence structure

tests/components/Pagination.test.tsx
alextaing: fixed grammar

tests/components/Pagination.test.tsx
alextaing: fixed!

tests/components/VerticalResults.test.tsx
```diff
@@ -72,152 +29,50 @@ const mockedActions = {
...
+      resultsCount: mockedVerticalResults.length
+    },
+    alternativeVerticals: []
+  };
+  mockVerticalSearchState({
+    noResults: mockedNoResults
   });
+  render(<VerticalResults {...verticalResultsProps} />);
+  mockedNoResults.allResultsForVertical.results.forEach(result =>
+    expect(screen.getByText(result.rawData.name as string)).toBeDefined()
```
oshi97, line 64: is this assert necessary?

oshi97: looks like it was here before but if we can remove it that'd be good!

oshi97: (also the other one)

tests/components/Pagination.test.tsx
```diff
@@ -0,0 +1,120 @@
+import { render, screen, within } from '@testing-library/react';
+import { State, VerticalSearchState } from '@yext/answers-headless-react';
+import { spyOnActions, mockAnswersState, mockAnswersHooks } from '../__utils__/mocks';
+import { mockedVerticalResults } from '../__fixtures__/data/verticalresults';
+import userEvent from '@testing-library/user-event';
+import { Pagination } from '../../src/components/Pagination';
+
+const mockedState: Partial<State> = {
+  vertical: {
```
nmanu1, line 9: I think some of the vertical state can be cut down now that we're only using the `Pagination` component. some of it (like `results`) was probably only needed for `VerticalResults`

alextaing: I'm taking a look at that at the moment-- I might be wrong, but I think the pagination component might behave differently depending on the presence of results or not, which requires looking at vertical state.  The first test is making me scratch my head a little because I think something might be going wrong from before, but I'm looking into it.

nmanu1 (commented):
can we add a test or two for the `paginateAllOnNoResults` prop?

tests/components/VerticalResults.test.tsx
alextaing: When it's removed, ts complains. Since I think `getByText` is an overloaded function and `result.rawData.name` ends up being type `unknown`, so it doesn't really know what to do unless we specify the type of the input.

alextaing:
The names for tests 2 and 3 are a little long, but I'm not sure how to trim them down without losing the essence of the test.  I also think that they are fairly atomic already, and I don't think they can be reasonably split further.  Do you have any suggestions?

tests/components/VerticalResults.test.tsx
oshi97: oh I see makes sense! here result.rawData is a map of strings to unknown values which is something the generate-types CLI command is going to help with

oshi97 (dismissed):
lgtm!

tests/components/Pagination.test.tsx
```diff
@@ -0,0 +1,168 @@
...
+};
+
+jest.mock('@yext/answers-headless-react');
+
+beforeEach(() => {
+  mockAnswersHooks({ mockedState, mockedActions });
+});
+
+describe('no results are returned from search', () => {
+  it('doesn\'t display pagination when paginateAllOnNoResults is false', () => {
```
yen-tt, line 31: I view this test as more of there's nothing to display (no results from the search and from alternativeresults section in the mocked state) so there's no pagination component, not necessarily because paginateAllOnNoResults is false.

tests/components/Pagination.test.tsx
```diff
@@ -0,0 +1,168 @@
+import { render, screen, within } from '@testing-library/react';
+import { State, VerticalSearchState } from '@yext/answers-headless-react';
+import { spyOnActions, mockAnswersState, mockAnswersHooks } from '../__utils__/mocks';
+import userEvent from '@testing-library/user-event';
+import { Pagination } from '../../src/components/Pagination';
+import { verticalNoResults } from '../__fixtures__/data/noresults';
```
yen-tt, line 6: nit: maybe we can make a vertical/ folder in data/ and have noresults and results file? having just noresults may be a bit less clear that this related to vertical results only, not universal as well

tests/components/Pagination.test.tsx
alextaing: renamed the test and rendered the pagination component without props (defaults to paginateAllOnNoResults=false)

tests/components/Pagination.test.tsx
alextaing: sure, done!

yen-tt (approved):

PR #191, opened 2022-06-22
Jest tests for RangeInput component

juliannzhou:
This PR adds Jest tests for RangeInput component. Tests were not passing in the first commit because of an error in useEffect function of RangeInput - instead of returning nothing, it returned a non-function.

J=SLAP-2145
TEST=auto

### DIFF
--- a/tests/components/RangeInput.test.tsx
+++ b/tests/components/RangeInput.test.tsx
@@ -24,36 +24,6 @@ const mockedActions = {
   executeVerticalQuery: jest.fn()
 };
 
-const selectableFilter: SelectableFilter = {
-  selected: true,
-  fieldId: '123',
-  matcher: Matcher.Between,
-  value: 'test'
-};
-
-const filterContextValue: FiltersContextType = {
-  selectFilter: () => null,
-  applyFilters: () => null,
-  filters: []
-};
-
-const filterContextValueDisabled: FiltersContextType = {
-  selectFilter: () => null,
-  applyFilters: () => null,
-  filters: [selectableFilter]
-};
-
-const filterGroupContextValue: FilterGroupContextType = {
-  searchValue: '',
-  fieldId: '123',
-  setSearchValue: () => null,
-  getCollapseProps: null,
-  getToggleProps: null,
-  isExpanded: null,
-  isOptionsDisabled: null,
-  setIsOptionsDisabled: () => null
-};
-
 jest.mock('@yext/answers-headless-react');
 
 it('renders the correct inital state', () => {
@@ -69,18 +39,7 @@ describe('Renders correctly for min input', () => {
   beforeEach(() => {
     mockAnswersHooks({ mockedState, mockedActions });
   });
-  it('renders input value, clear, and apply option when inputing min value', async () => {
-    renderRangeInput(filterContextValue);
-    const minTextbox = screen.getAllByRole('textbox')[0];
-    userEvent.type(minTextbox, '10');
-    await waitFor(() => {
-      expect(minTextbox).toHaveValue('10');
-    });
-    expect(screen.getByText('Clear min and max')).toBeDefined();
-    expect(screen.getByText('Apply')).toBeDefined();
-  });
-
-  it('record proper values in state when applying min', async () => {
+  it('renders correctly when inputing min and applies proper values in state', async () => {
     renderRangeInput(filterContextValue);
     const actions = spyOnActions();
     const minTextbox = screen.getAllByRole('textbox')[0];
@@ -88,6 +47,8 @@ describe('Renders correctly for min input', () => {
     await waitFor(() => {
       expect(minTextbox).toHaveValue('10');
     });
+    expect(screen.getByText('Clear min and max')).toBeDefined();
+    expect(screen.getByText('Apply')).toBeDefined();
     userEvent.click(screen.getByText('Apply'));
     expect(actions.setFilterOption).toHaveBeenCalledWith({
       displayName: 'Over 10',
@@ -122,18 +83,7 @@ describe('Renders correctly for max input', () => {
     mockAnswersHooks({ mockedState, mockedActions });
   });
 
-  it('renders input value, clear, and apply option when inputing max value', async () => {
-    renderRangeInput(filterContextValue);
-    const maxTextbox = screen.getAllByRole('textbox')[1];
-    userEvent.type(maxTextbox, '20');
-    await waitFor(() => {
-      expect(maxTextbox).toHaveValue('20');
-    });
-    expect(screen.getByText('Clear min and max')).toBeDefined();
-    expect(screen.getByText('Apply')).toBeDefined();
-  });
-
-  it('record proper values in state when applying max', async () => {
+  it('renders correctly when inputing max and applies proper values in state', async () => {
     renderRangeInput(filterContextValue);
     const actions = spyOnActions();
     const maxTextbox = screen.getAllByRole('textbox')[1];
@@ -141,6 +91,8 @@ describe('Renders correctly for max input', () => {
     await waitFor(() => {
       expect(maxTextbox).toHaveValue('20');
     });
+    expect(screen.getByText('Clear min and max')).toBeDefined();
+    expect(screen.getByText('Apply')).toBeDefined();
     userEvent.click(screen.getByText('Apply'));
     expect(actions.setFilterOption).toHaveBeenCalledWith({
       displayName: 'Up to 20',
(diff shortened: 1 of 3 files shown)


tests/components/RangeInput.test.tsx
```diff
@@ -0,0 +1,296 @@
...
+      expect(minTextbox).toHaveValue('20');
+    });
+    expect(maxTextbox).toHaveValue('10');
+    expect(screen.getByText('Invalid range')).toBeDefined();
+    expect(actions.setFilterOption).toHaveBeenCalledTimes(0);
+  });
+});
+
+it('renders correctly when disabled', () => {
+  render(<FilterGroupContext.Provider value={filterGroupContextValue}>
```
oshi97, line 291: what is this test doing, just making sure no error is thrown?

tests/components/RangeInput.test.tsx
```diff
@@ -0,0 +1,296 @@
...
+  filters: []
+};
+
+const filterContextValueDisabled: FiltersContextType = {
+  selectFilter: () => null,
+  applyFilters: () => null,
+  filters: [selectableFilter]
+};
+
+const filterGroupContextValue: FilterGroupContextType = {
```
oshi97, line 46: Is it possible to have this be `Partial<FilterGroupContextType>` and exclude the null values? not sure if typescript would yell at you

tests/components/RangeInput.test.tsx
```diff
@@ -0,0 +1,296 @@
...
+        },
+        end: {
+          matcher: '$le',
+          value: 20
+        },
+      }
+    });
+  });
+
+  it('renders correctly when clearing inputs', async () => {
```
yen-tt, line 250: can we ensures that actions.setFilterOption is not called at the end? (similarly for other `renders correctly when clearing input` tests)

tests/components/RangeInput.test.tsx
```diff
@@ -0,0 +1,296 @@
...
+    userEvent.click(screen.getByText('Clear min and max'));
+    expect(maxTextbox).toHaveValue('');
+  });
+});
+
+describe('Renders correctly for min and max inputs', () => {
+  beforeEach(() => {
+    mockAnswersHooks({ mockedState, mockedActions });
+  });
+  it('renders input value, clear, and apply option when inputing min and max values', async () => {
```
yen-tt, line 197: This test can probably be remove since the test below `it('record proper values in state when applying min and max')` already check for all the expected text on screen AND check for the state when apply is clicked. Similarly with the describe blocks for min only and max only.

tests/components/RangeInput.test.tsx
oshi97: Ah ok keep them then!

oshi97: We're not using ts-jest so when jest runs it ignores the typescript errors

tests/components/RangeInput.test.tsx
```diff
@@ -0,0 +1,260 @@
...
+  </FilterGroupContext.Provider>);
+  const [minTextbox, maxTextbox] = screen.getAllByRole('textbox');
+  expect(minTextbox).toHaveAttribute('disabled');
+  expect(maxTextbox).toHaveAttribute('disabled');
+  expect(screen.getByText('Unselect an option to enter in a range.')).toBeDefined();
+});
+
+function renderRangeInput() {
+  return (
+    render(<FilterGroupContext.Provider value={filterGroupContextValue}>
```
oshi97, line 254: line break after paren in render

tests/components/RangeInput.test.tsx
```diff
@@ -0,0 +1,260 @@
...
+    });
+    expect(maxTextbox).toHaveValue('10');
+    expect(screen.getByText('Invalid range')).toBeDefined();
+    expect(actions.setFilterOption).toHaveBeenCalledTimes(0);
+  });
+});
+
+it('renders correctly when disabled', () => {
+  render(<FilterGroupContext.Provider value={filterGroupContextValue}>
+    <FiltersContext.Provider value={filterContextValueDisabled}>
```
oshi97, line 242: I think you can use `renderRangeInput` here if you add `filterContextValueDisabled` as a parameter?

tests/components/RangeInput.test.tsx
```diff
@@ -0,0 +1,260 @@
...
+    expect(actions.setFilterOption).toHaveBeenCalledTimes(0);
+  });
+});
+
+describe('Renders correctly for min and max inputs', () => {
+  beforeEach(() => {
+    mockAnswersHooks({ mockedState, mockedActions });
+  });
+
+  it('renders input values, clear, apply options and record proper values in state when applying min and max', async () => {
```
oshi97, line 178: I think it might be clearer to split this test up into one for checking rendering and one for checking what setFilterOption is called with

oshi97: Long names for things like variables, tests, and functions generally mean they're doing more than one thing and can be broken up. The general idea for breaking things up is called the "single responsibility principle", with the general idea being that small building blocks are easier to understand than big complicated things

yen-tt: She had two separate tests in her original commit, but I made a comment to combine them to avoid duplicating the same setup since she had to type up all the values and confirm the necessary button(s) before checking actions.setFilterOption. The difference, beside checking for functions, might be not needing to check for `Clear min and max` label.

oshi97: I see, would it be possible to shorten the test name?

tests/components/SpellCheck.test.tsx
```diff
@@ -1,7 +1,7 @@
 import { render, screen } from '@testing-library/react';
 
 import { SpellCheck } from '../../src/components/SpellCheck';
-import { State } from '@yext/answers-headless-react';
+import { Source, State } from '@yext/answers-headless-react';
```
nmanu1, line 4: is the added `Source` import used anywhere in the file?

juliannzhou: added this file to staged changes by mistake!

tests/components/RangeInput.test.tsx
```diff
@@ -0,0 +1,257 @@
...
+    expect(actions.setFilterOption).toHaveBeenCalledTimes(0);
+  });
+});
+
+describe('Renders correctly for min and max inputs', () => {
+  beforeEach(() => {
+    mockAnswersHooks({ mockedState, mockedActions });
+  });
+
+  it('renders correctly when inputing range and applies proper values in state', async () => {
```
yen-tt, line 178: Similar to the way you combined the "render correctly" and "apply to state" tests, lets update the two describe block above to do the same

tests/components/RangeInput.test.tsx
```diff
@@ -0,0 +1,257 @@
...
+      expect(minTextbox).toHaveValue('10');
+    });
+    expect(maxTextbox).toHaveValue('20');
+    userEvent.click(screen.getByText('Clear min and max'));
+    expect(minTextbox).toHaveValue('');
+    expect(maxTextbox).toHaveValue('');
+    expect(actions.setFilterOption).toHaveBeenCalledTimes(0);
+  });
+
+  it('renders correctly when input range is invalid and no filter is set in state', async () => {
```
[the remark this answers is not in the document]
oshi97: the indentation errors should be auto fixable with either eslint with autofix turned on or with a vscode formatter

tests/components/RangeInput.test.tsx
```diff
@@ -0,0 +1,257 @@
...
+
+it('renders correctly when disabled', () => {
+  renderRangeInput(filterContextValueDisabled);
+  const [minTextbox, maxTextbox] = screen.getAllByRole('textbox');
+  expect(minTextbox).toHaveAttribute('disabled');
+  expect(maxTextbox).toHaveAttribute('disabled');
+  expect(screen.getByText('Unselect an option to enter in a range.')).toBeDefined();
+});
+
+function renderRangeInput(expectedFilterContextValue) {
```
oshi97, line 248: I would probably set the default here to `filterContextValue` just to reduce a little typing.

tests/components/RangeInput.test.tsx
```diff
@@ -0,0 +1,257 @@
...
+      expect(minTextbox).toHaveValue('20');
+    });
+    expect(maxTextbox).toHaveValue('10');
+    expect(screen.getByText('Invalid range')).toBeDefined();
+    expect(actions.setFilterOption).toHaveBeenCalledTimes(0);
+  });
+});
+
+it('renders correctly when disabled', () => {
+  renderRangeInput(filterContextValueDisabled);
```
oshi97, line 241: What do you think about defining `filterContextValueDisabled` within this test since it's only used here? Having a huge block of declared fixtures at the top of the file on one hand could be viewed as organized, on the other hand it's a little overwhelming when you first open things up

oshi97: This is more a matter a taste I think so feel free to ignore?

yen-tt (approved):

PR #192, opened 2022-06-22
Consistent naming for CssClass interface containers

alextaing:
Names for styles for component containers and container loading in CSS class interfaces were inconsistent between components.  To increase consistency, they were renamed to the standard `container` and `container___loading`.  See the following slack thread for information.

https://yext.slack.com/archives/C032CKFARGS/p1655911645857069 

J=SLAP-2172
TEST=auto

Storybook will validate that no visual changes have been made.

### DIFF
--- a/src/components/AppliedFilters.tsx
+++ b/src/components/AppliedFilters.tsx
@@ -14,8 +14,8 @@ import { useStateUpdatedOnSearch } from '../hooks/useStateUpdatedOnSearch';
  * @public
  */
 export interface AppliedFiltersCssClasses {
-  appliedFiltersContainer?: string,
-  appliedFiltersContainer___loading?: string,
+  container?: string,
+  container___loading?: string,
   nlpFilter?: string,
   removableFilter?: string,
   removeFilterButton?: string,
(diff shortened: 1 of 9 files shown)


src/components/AppliedFilters.tsx
```diff
@@ -90,8 +90,8 @@ export function AppliedFilters(props: AppliedFiltersProps): JSX.Element {
   ]);
 
   const cssClasses = useComposedCssClasses(builtInCssClasses, customCssClasses);
-  cssClasses.appliedFiltersContainer = classNames(cssClasses.appliedFiltersContainer, {
-    [cssClasses.appliedFiltersContainer___loading ?? '']: isLoading
+  cssClasses.container = classNames(cssClasses.container, {
+    [cssClasses.container___loading ?? '']: isLoading
```
tmeyer2115, line 94: Can we fit this all on one line now?

tmeyer2115:
Seems like something is borking the unit tests. As I looked through this PR, I'm wondering if `container_loading` is the best name. It's not really the container, but the Component that's loading. I'm wondering if having `container` and `component_loading` would work or if that's too odd. I think it's worth getting Product's take.

src/components/AppliedFilters.tsx
```diff
@@ -14,8 +14,8 @@ import { useStateUpdatedOnSearch } from '../hooks/useStateUpdatedOnSearch';
  * @public
  */
 export interface AppliedFiltersCssClasses {
-  appliedFiltersContainer?: string,
-  appliedFiltersContainer___loading?: string,
+  container?: string,
+  container___loading?: string,
```
oshi97, line 18: @tmeyer2115 do you remember the reason we use three underscores here? I think Connor was the one who originally had the idea and but I can't remember exactly what it was

PR #193, opened 2022-06-23
Export a content path for user's tailwind config

yen-tt:
Export a string AnswersTailwindPath representing the content path to files in component lib that uses tailwind classnames. This is intended to be imported and used in user's tailwind file:
`node_modules/@yext/answers-react-components/lib/**/*.{js,ts,jsx,tsx}`
J=SLAP-2133
TEST=manual

updated test-site to use AnswersTailwindPath in its tailwind.config.js and successfully started the site with UI looking as expected.


src/index.ts
```diff
@@ -2,3 +2,24 @@ export * from './components';
...
+ * @example
+ * In user's tailwind.config.js file:
+ *
+ * const \{ AnswersTailwindPath \} = require('\@yext/answers-react-components')
+ * module.exports = \{
+ *  content: [ AnswersTailwindPath ],
+ *  ...
+ * \};
+ */
+export const AnswersTailwindPath = 'node_modules/@yext/answers-react-components/lib/**/*.{js,ts,jsx,tsx}';
```
nmanu1, line 25: can the name be changed so it doesn't have "Answers" in it? I'd rather not add more things to our libraries that reference Answers since we're re-branding

tmeyer2115, line 25: In `lib` will there be `ts` and `tsx` files? I thought they would all be trans-piled to JS/JSX at that point. Also, I think the name `AnswersTailwindPath` is a bit confusing. It sounds like we're pointing them to our Tailwind config. We're pointing them to the package's Tailwind "content" in a sense.

yen-tt: right, I forgot to update! I originally had it point to  'node_modules/@yext/answers-react-components/**/*.{js,ts,jsx,tsx}' which includes other ts/tsx files outside of lib/. During testing, test-site would get an error about recursive content paths since that glob file path includes test-site's stuff as well. I updated it to restrict to lib/, user shouldn't be importing anything outside of that anyway.
For the name, maybe "componentSourcePath" or "componentContentPath"?

tmeyer2115: Yeah, something like `ComponentsContentPath` sounds good!

src/index.ts
```diff
@@ -2,3 +2,24 @@ export * from './components';
...
+ * @public
+ *
+ * @remarks
+ * This assumes that the node_modules folder containing the component library
+ * is in adjacent level with the user's tailwind.config.js file.
+ *
+ * @example
+ * In user's tailwind.config.js file:
+ *
+ * const \{ AnswersTailwindPath \} = require('\@yext/answers-react-components')
```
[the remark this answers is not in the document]
yen-tt: done!

tmeyer2115 (dismissed):

src/index.ts
```diff
@@ -2,3 +2,26 @@ export * from './components';
...
+ * @remarks
+ * This assumes that the node_modules folder containing the component library
+ * is in adjacent level with the user's tailwind.config.js file.
+ *
+ * @example
+ * In user's tailwind.config.js file:
+ *
+ * ```js
+ * const { ComponentsContentPath } = require('@yext/answers-react-components')
+ * module.exports = {
```
oshi97, line 21: nit: can we have a line break before module.exports

src/index.ts
```diff
@@ -2,3 +2,26 @@ export * from './components';
...
+ *
+ * @remarks
+ * This assumes that the node_modules folder containing the component library
+ * is in adjacent level with the user's tailwind.config.js file.
+ *
+ * @example
+ * In user's tailwind.config.js file:
+ *
+ * ```js
+ * const { ComponentsContentPath } = require('@yext/answers-react-components')
```
oshi97, line 20: semicolon

test-site/tailwind.config.js
```diff
@@ -1,9 +1,10 @@
 const tailwindConfig = require('../tailwind.config.cjs');
+const { ComponentsContentPath } = require('@yext/answers-react-components')
```
nmanu1, line 2: nit: semicolon

src/index.ts
```diff
@@ -2,3 +2,26 @@ export * from './components';
...
+ * module.exports = {
+ *  content: [ ComponentsContentPath ],
+ *  ...
```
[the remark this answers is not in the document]
yen-tt: sure!

oshi97 (approved):

PR #199, opened 2022-06-24
Add GH Action to run ES Linter on PRs into `main`.

tmeyer2115:
TEST=manual

Set ESLint Warnings to Errors and saw the PR check fail. Reverted this change and saw the PR check pass.

### DIFF
--- a/.github/workflows/run-linting.yml
+++ b/.github/workflows/run-linting.yml
@@ -13,4 +13,5 @@ jobs:
       - uses: actions/checkout@v3
       - name: Use Node.js
         uses: actions/setup-node@v3
+      - run: npm ci
       - run: npm run lint
--- a/.github/workflows/run-linting.yml
+++ b/.github/workflows/run-linting.yml
@@ -1,0 +1,16 @@
+name: Run Linter
+
+on:
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  linting:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3
+      - name: Use Node.js
+        uses: actions/setup-node@v3
+      - run: npm lint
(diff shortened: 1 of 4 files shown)


.github/workflows/run-linting.yml
```diff
@@ -0,0 +1,17 @@
...
+jobs:
+  linting:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3
+      - name: Use Node.js
+        uses: actions/setup-node@v3
+      - run: npm ci
+      - run: npm run lint
```
oshi97, line 17: Do we want to make the workflow fail with --max-warnings? It's super sweet we get "check annotations" even without the workflow failing

.github/workflows/run-linting.yml
```diff
@@ -0,0 +1,17 @@
+name: Run Linter
+
+on:
+  pull_request:
+    branches:
+      - main
```
oshi97, line 6: maybe we just run this on all pull requests?

tmeyer2115: We could! I don't have a strong opinion either way. My thoughts for not having it on all PRs was that people may have dev branches going into feature branches. Maybe their work is still in progress and they'll get to the linting stuff later on.

.github/workflows/run-linting.yml
tmeyer2115: Sure, I can do that!

.github/workflows/run-linting.yml
oshi97: Ah ok makes sense, I was thinking about our forgetting to add our future develop branch but I guess we can just not forget

.github/workflows/run-linting.yml
```diff
@@ -0,0 +1,17 @@
+name: Run Linter
```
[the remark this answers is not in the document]
oshi97: Tom said we plan to add it later, we can probably say so in the PR description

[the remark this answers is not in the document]
tmeyer2115: Eventually we will migrate it, yes.

oshi97 (approved):

PR #201, opened 2022-06-27
Made 'filter' not show up when label is undefined

ElemelonWind:
J=SLAP-2178
TEST=auto,manual


package.json
```diff
@@ -126,6 +126,7 @@
     "classnames": "^2.3.1",
     "prop-types": "^15.8.1",
     "react-collapsed": "^3.3.0",
+    "react-router-dom": "^6.3.0",
```
oshi97, line 129: I think we don't need react-router-dom as a dep

test-site/src/pages/PeoplePage.tsx
```diff
@@ -46,6 +46,7 @@ export function PeoplePage() {
             ]}
             searchOnChange={false}
           />
+          <br/>
```
oshi97, line 49: Was this just for testing or do we want to keep this?

src/components/FilterSearch.tsx
```diff
@@ -66,7 +66,7 @@ export interface FilterSearchProps {
  */
 export function FilterSearch({
   searchFields,
-  label = 'Filter',
```
oshi97, line 69: Just checking that we want to remove the default? (I would be down to remove it, it's not a label I can imagine people using...)

tmeyer2115 (commented):
We shouldn't need the `package-lock.json` or `test-site/package-lock.json` changes in this PR. The code doesn't add any new dependencies. Those files may have changed for a couple reasons: you weren't initially on the latest version of `main` or you have a different `npm`/Node version. Either way, you can un-stage the changes to these two files from your PR.

tests/components/FilterSearch.test.tsx
```diff
@@ -26,7 +26,7 @@ describe('search with section labels', () => {
   });
 
   it('renders the filter search bar, "Filter" label, and default placeholder text', () => {
-    render(<FilterSearch searchFields={searchFieldsProp} />);
+    render(<FilterSearch searchFields={searchFieldsProp} label='Filter' />);
```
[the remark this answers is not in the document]
oshi97: > We shouldn't need the `package-lock.json` or `test-site/package-lock.json` changes in this PR. The code doesn't add any new dependencies. Those files may have changed for a couple reasons: you weren't initially on the latest version of `main` or you have a different `npm`/Node version. Either way, you can un-stage the changes to these two files from your PR.

I believe the package-lock changed because in an earlier commit there was a react-router-dom dep was added (which has since been removed). Now that it's gone, if you rerun npm install in the base folder and test-site folder (or just unstage the changes like Tom suggested) the changes should go away.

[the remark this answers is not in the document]
oshi97: Could we just add a story? I think we probably don't need these "render checking" tests we keep adding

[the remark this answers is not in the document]
oshi97: We've kinda kept adding more jest tests that check rendering, which works but leads to pretty bloated test files

tests/components/FilterSearch.stories.tsx
```diff
@@ -77,3 +77,11 @@ DropdownSectioned.play = ({ canvasElement }) => {
   const canvas = within(canvasElement);
   userEvent.type(canvas.getByRole('textbox'), 'name');
 };
+
+export const NoLabel = (args: FilterSearchProps) => {
+  return (
+    <AnswersHeadlessContext.Provider value={generateMockedHeadless(mockedHeadlessState)}>
+      <FilterSearch searchFields={searchFields} {...args} />
+    </AnswersHeadlessContext.Provider>
+  )
```
oshi97, line 86: oh for these you can run `npx eslint . --fix` and it'll autofix simple things like this

tests/components/FilterSearch.test.tsx
```diff
@@ -36,6 +36,17 @@ describe('search with section labels', () => {
     expect(searchBarElement[0].placeholder).toBe('Search here...');
   });
 
+  it('renders the filter search bar, no label, and default placeholder text', () => {
```
oshi97, line 39: we can remove this test since the storybook covers us here

oshi97 (approved):
lgtm!

PR #203, opened 2022-06-27
added hover styling to "option"

alextaing:
This PR makes hovering over filter search autocomplete suggestions trigger a highlight on the selected option.  The JIRA item requested that the `focusedOption` CSS should be applied to options that are hovered over.  However, adding a `hover` style to `option` more closely replicates what is done in `SearchBar` and seems to be a more straightforward approach.

image

image

(Unfortunately screenshots do not show cursors.  Cursor is over 'first name 1', not selected with arrow keys.)

J=SLAP-2179
TEST=manual

Attempted to add hovering test to storybook, but ran into issues with simulating hover.

### DIFF
--- a/src/components/FilterSearch.tsx
+++ b/src/components/FilterSearch.tsx
@@ -30,7 +30,7 @@ const builtInCssClasses: Readonly<FilterSearchCssClasses> = {
   inputElement: 'text-sm bg-white outline-none h-9 w-full p-2 rounded-md border border-gray-300 focus:border-primary text-neutral-dark placeholder:text-neutral',
   sectionLabel: 'text-sm text-neutral-dark font-semibold pb-2',
   focusedOption: 'bg-gray-100',
-  option: 'text-sm text-neutral-dark pb-1 cursor-pointer'
+  option: 'text-sm text-neutral-dark pb-1 cursor-pointer hover:bg-gray-100'
 };
 
 /**


src/components/FilterSearch.tsx
```diff
@@ -30,7 +30,7 @@ const builtInCssClasses: Readonly<FilterSearchCssClasses> = {
   inputElement: 'text-sm bg-white outline-none h-9 w-full p-2 rounded-md border border-gray-300 focus:border-primary text-neutral-dark placeholder:text-neutral',
   sectionLabel: 'text-sm text-neutral-dark font-semibold pb-2',
   focusedOption: 'bg-gray-100',
-  option: 'text-sm text-neutral-dark pb-1 cursor-pointer'
+  option: 'text-sm text-neutral-dark pb-1 cursor-pointer hover:bg-gray-100'
```
oshi97, line 33: Is there a way we can apply the styling defined in focusedOption on hover? Right now if somebody has a custom focusedOption class the hover styling will be different. This isn't crucial so don't spend forever if it seems impossible

oshi97: HOWEVER I suspect it's possible

oshi97: Actually hmm, on second thought I'm not sure if we want to enforce the same styling for both highlighting and on hover

oshi97 (approved):

src/components/FilterSearch.tsx
alextaing: OK--I think that it might be a good idea, but we can address it in a later item if it comes up as something we want.

alextaing: * note that if we do end up wanting to change it, we should also change the behavior to match in the SearchBar dropdown.

PR #206, opened 2022-06-28
Clicking outside dropdown triggers onSelect

ElemelonWind:
Edited `onToggle` in `Dropdown` to fix a bug in `FilterSearch` where clicking outside the dropdown doesn't search the correct value when 'enter' is pressed. Note: to fix this same bug with `SearchBar` we have to pass in an `onToggle` prop in that component too (which will be resolved in a future item)

J=SLAP-2198
TEST=auto


oshi97:
is there a jira item for this PR?

src/components/Dropdown/Dropdown.tsx
```diff
@@ -74,6 +74,7 @@ export function Dropdown(props: PropsWithChildren<DropdownProps>): JSX.Element {
   }, [parentQuery, lastTypedOrSubmittedValue, updateFocusedItem, setLastTypedOrSubmittedValue]);
 
   useRootClose(containerRef, () => {
+    onSelect?.(value, focusedIndex, focusedItemData);
```
oshi97, line 77: This change improves the FilterSearch edge case we have in mind, but seems a little weird for Dropdown? I'm not sure if it makes sense for navigating away from the Dropdown to result in it selecting something. For instance if I had a regular searchbar that used onSelect to trigger searches, I don't think I would want clicking away from the SearchBar to run a search.

I like what https://www.zocdoc.com/ does here (this is one of the examples product used for their FilterSearch vision). If you type a bit into zocdoc's filtersearch, use the arrow keys to highlight an option, and then click away, the query will revert to the little bit you initially typed. 

Might be worth asking product what they want as well though I'm preettttty sure it would be this behavior? The drawback is that if you happened to type out the full name of a filter and then don't select it, it will have the same behavior as now. But in that case the user literally typed out the whole thing so I don't think it's that weird, and would still be my preferred behavior.

ElemelonWind:
> 

Tom assigned me the task yesterday but I don't think he's made a Jira item for it yet.

src/components/Dropdown/Dropdown.tsx
ElemelonWind: Actually, `SearchBar` doesn't have anything to handle the `onSelect` event - the `DropdownItem`s in that component  only have an `onClick` event that triggers the search. So if we just write a default `onSelect` handler for the `SearchBar` we should be fine right? Your comment made me realize that I only fixed `FilterSearch`; when a user clicks out of the `SearchBar` dropdown and then hits search, it's still broken.

tmeyer2115: I wouldn't worry about the `SearchBar` for now; I think we may need to talk with Product about their desired behavior on blur for the `SearchBar`. So, we can make another item there if the current `SearchBar` UX is not what they want.

oshi97: Ah yeah when I said "if I had a searchbar that used onSelect" I didn't mean our current searchbar, but a searchbar we or maybe even somebody else implements in the future. For lower level components like Dropdown, we don't want to just make things work in the current world, they also have to make sense with our "mental model" for what a Dropdown is.

oshi97: @tmeyer2115 I think an onBlur makes sense

ElemelonWind: > What is broken about the current `SearchBar`'s behavior?

Currently the `SearchBar` has the same problem that we're fixing with `FilterSearch`, where if you arrow down to select something then click out of the `Dropdown`, the UI shows that it's selected, but when you trigger the search it will still run on the previous query.

alextaing:
Maybe this is outside the scope of this item, but could we write a jest test replicating what you did manually?

src/components/FilterSearch.tsx
```diff
@@ -109,6 +109,22 @@ export function FilterSearch({
...
+    const newFilter = itemData?.filter as Filter;
+    const newDisplayName = itemData?.displayName as string;
+    if (newFilter && newDisplayName) {
+      if (currentFilter) {
+        answersActions.setFilterOption({ ...currentFilter, selected: false });
+      }
+      answersActions.setFilterOption({ ...newFilter, displayName: newDisplayName, selected: true });
+      setCurrentFilter(newFilter);
+      answersActions.setOffset(0);
+      if (searchOnSelect) {
```
alextaing, line 122: Do we still want this behavior to depend on `searchOnSelect` or should it depend on a new prop?

ElemelonWind: You're right, I'll remove that conditional. I don't think it would make sense for a search to be triggered anytime the dropdown is blurred, so I'm not sure if another prop is needed.

ElemelonWind:
> Maybe this is outside the scope of this item, but could we write a jest test replicating what you did manually?

Yeah, I can write a jest test!

src/components/Dropdown/Dropdown.tsx
```diff
@@ -74,6 +81,7 @@ export function Dropdown(props: PropsWithChildren<DropdownProps>): JSX.Element {
   }, [parentQuery, lastTypedOrSubmittedValue, updateFocusedItem, setLastTypedOrSubmittedValue]);
 
   useRootClose(containerRef, () => {
+    onBlur?.(lastTypedOrSubmittedValue, value, focusedIndex, focusedItemData);
```
oshi97, line 84: Nitting but should onBlur also run when you use the tab key to leave the Dropdown? Maybe we could update FilterSearch to use `onToggle` instead?

src/components/Dropdown/Dropdown.tsx
```diff
@@ -21,7 +21,13 @@ export interface DropdownProps {
   initialValue?: string,
   parentQuery?: string,
   onSelect?: (value: string, index: number, focusedItemData: Record<string, unknown> | undefined) => void,
-  onToggle?: (isActive: boolean, value: string) => void,
+  onToggle?: (
+    prevValue: string,
```
tmeyer2115, line 25: I'm wondering if some of these things need to be optional (or `undefined`). Supposing I toggle the drop-down open for the first time. There shouldn't be a previous value, a value, or any focusedItemData.

src/components/Dropdown/Dropdown.tsx
```diff
@@ -61,9 +67,17 @@ export function Dropdown(props: PropsWithChildren<DropdownProps>): JSX.Element {
     screenReaderKey,
     setScreenReaderKey
   );
-  const { focusedIndex, updateFocusedItem } = focusContext;
+  const { focusedIndex, focusedItemData, updateFocusedItem } = focusContext;
 
-  const dropdownContext = useDropdownContextInstance(value, screenReaderUUID, onToggle, onSelect);
+  const dropdownContext = useDropdownContextInstance(
+    lastTypedOrSubmittedValue,
```
tmeyer2115, line 73: What is this value when the `Dropdown` is open for the first time and the input is empty?

src/components/Dropdown/Dropdown.tsx
ElemelonWind: All of the things except focusedItemData start with a default value- the strings start as '' and the index starts as -1

src/components/FilterSearch.tsx
```diff
@@ -93,7 +93,7 @@ export function FilterSearch({
   const hasResults = sections.flatMap(s => s.results).length > 0;
   const [currentFilter, setCurrentFilter] = useState<Filter>();
 
-  const handleSelectDropdown = useCallback((_value, _index, itemData) => {
+  const handleFilterDropdown = useCallback((itemData, select) => {
```
nmanu1, line 96: can this be renamed? right now it sounds like this function should handle what happens after some "filtering" behavior is performed on the dropdown, which doesn't make too much sense

tests/components/Dropdown.test.tsx
```diff
@@ -184,6 +184,40 @@ describe('Dropdown', () => {
...
   });
 
+  it('selects when an option is focused on toggle', () => {
+    const mockedOnToggleFn = jest.fn();
+    const dropdownProps: DropdownProps = {
+      screenReaderText: 'screen reader text here',
+      onToggle: mockedOnToggleFn
+    };
+    render(
+      <div data-testid='container'>
```
nmanu1, line 194: why is a `data-testid` attribute being added here? is it to select the dropdown container or the document?

ElemelonWind: Oh I copied part of a previous test and forgot to remove it, I'll fix that

src/components/Dropdown/Dropdown.tsx
```diff
@@ -21,7 +21,13 @@ export interface DropdownProps {
   initialValue?: string,
   parentQuery?: string,
   onSelect?: (value: string, index: number, focusedItemData: Record<string, unknown> | undefined) => void,
-  onToggle?: (isActive: boolean, value: string) => void,
+  onToggle?: (
+    prevValue: string,
+    value: string,
+    index: number,
+    focusedItemData: Record<string, unknown> | undefined,
+    isActive: boolean
```
tmeyer2115, line 29: Since we moved `isActive` last, are there any usages of `Dropdown` that supply `onToggle` that need updating?

src/components/FilterSearch.tsx
ElemelonWind: that's true, what would you suggest? `handleDropdown` seemed too vague but it's being used for both `onSelect` and `onToggle` in this component.

tmeyer2115 (approved):

PR #214, opened 2022-07-01
Fix SearchBar stories

nmanu1:
This PR fixes the `SearchBar` stories so they display recent searches as expected. Adding recent searches manually to the browser's local storage was causing problems. This is likely because of the hard-coded timestamp, which may cause the data to be removed from storage if it is too far in the past. I changed the recent searches back to being added using `userEvent`s.

J=none
TEST=auto

See that Percy snapshots include the recent searches for `SearchBar` stories that have the dropdown expanded.

### DIFF
--- a/tests/components/SearchBar.stories.tsx
+++ b/tests/components/SearchBar.stories.tsx
@@ -38,19 +38,6 @@ const meta: ComponentMeta<typeof SearchBar> = {
 export default meta;
 
 export const Primary = (args: SearchBarProps) => {
-  localStorage.clear();
-  const recentSearches = [
-    {
-      query: 'recent search 2',
-      timestamp: 1656512443860
-    },
-    {
-      query: 'recent search 1',
-      timestamp: 1656512440493
-    }
-  ];
-  localStorage.setItem('__yxt_recent_searches__', JSON.stringify(recentSearches));
-
   return (
     <AnswersHeadlessContext.Provider value={generateMockedHeadless()}>
       <SearchBar {...args} />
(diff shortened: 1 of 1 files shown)


tests/components/SearchBar.stories.tsx
```diff
@@ -38,19 +38,6 @@ const meta: ComponentMeta<typeof SearchBar> = {
 export default meta;
 
 export const Primary = (args: SearchBarProps) => {
```
yen-tt, line 40: can we have Primary.play runs `conductRecentSearches` instead? I believe the other stories would get the recent searches since Primary would set those to local storage beforehand, then we don't need to call `conductRecentSearches` on DropdownExpanded and DropdownHighlight

oshi97: Oh the recent searches setting is needed by those two stories! Alex and I tweaked those a little while back. https://github.com/yext/answers-react-components/commit/1a3dde7da53b9f18f2127bab3bba12d82dc30732

Adding a play(0 function to `Primary` won't work for the stories that need to redefine the play() in order to do additional stuff

tests/components/SearchBar.stories.tsx
```diff
@@ -61,12 +48,24 @@ export const Primary = (args: SearchBarProps) => {
...
+  conductRecentSearches(textboxEl);
+  userEvent.click(textboxEl);
 };
 
 export const DropdownHighlight = Primary.bind({});
 DropdownHighlight.play = ({ canvasElement }) => {
   const canvas = within(canvasElement);
-  userEvent.click(canvas.getByRole('textbox'));
+  const textboxEl = canvas.getByRole('textbox');
+  conductRecentSearches(textboxEl);
```
oshi97, line 60: previously this didn't work, but I think it's because we needed to have a delay before hitting the Tabs

oshi97 (approved):

tests/components/SearchBar.stories.tsx
nmanu1: conducting the recent searches in only one of the stories does set it in local storage for the other stories to use. but, there's no guarantee that the stories run in a particular order (even `Primary` before the others) so it's not dependable to only have it in one

PR #215, opened 2022-07-06
fix StaticFiltersProvider styling

yen-tt:
update className's default styling from `md-56` to `md:w-56`

J=SLAP-2207
TEST=manual

include only StaticFilters on the left hand side of people page. With `md:w-56`, see that the width of it now changes when the page size changes from >768px to <768px

### DIFF
--- a/src/components/Filters/StaticFiltersProvider.tsx
+++ b/src/components/Filters/StaticFiltersProvider.tsx
@@ -28,7 +28,7 @@ export type StaticFiltersProviderProps = PropsWithChildren<{
  */
 export function StaticFiltersProvider({
   children,
-  className = 'md-56',
+  className = 'md:w-56',
   searchOnChange = true
 }: StaticFiltersProviderProps): JSX.Element {
   const answersActions = useAnswersActions();


src/components/Filters/StaticFiltersProvider.tsx
```diff
@@ -28,7 +28,7 @@ export type StaticFiltersProviderProps = PropsWithChildren<{
  */
 export function StaticFiltersProvider({
   children,
-  className = 'md-56',
```
oshi97, line 31: for my edification what was md-56 doing before? did we just miss a breaking change when we upgraded to tailwind 3 or something?

oshi97: tried looking for it but couldn't find anything :(

yen-tt: I believe this was a typo. `md-56` does not apply any additional css to the div. md sets breakpoint (768px) of when the styling on the right hand side of the colon should be applied.
Screen Shot 2022-07-06 at 11 45 35 AM

oshi97: awesome!

oshi97 (approved):

PR #217, opened 2022-07-06
Removed allowPagination prop from VerticalResults

ElemelonWind:
J=SLAP-2211
TEST=auto


oshi97 (approved):
🔥

src/components/VerticalResults.tsx
```diff
@@ -1,7 +1,7 @@
 import { CardComponent } from '../models/cardComponent';
 import { useAnswersState } from '@yext/answers-headless-react';
 import { VerticalResultsDisplay } from './VerticalResultsDisplay';
-import { Pagination, PaginationCssClasses } from './Pagination';
+import { PaginationCssClasses } from './Pagination';
```
nmanu1, line 4: I don't think we should include the pagination classes in vertical results anymore because they won't be used

nmanu1: Yes, since we should only be using `verticalResultsContainer` and `verticalResultsLoading` now

PR #223, opened 2022-07-07
update docs for RenderEntityPreviews

yen-tt:
Previously, for a search without any query suggestions and recent search option, if user returned JSX.element for entity previews doesn’t contain DropdownItem, this would cause calculateEntityPreviewsCount to return 0. Thus not rendering anything in dropdown. This is fixed in a merged commit when refactoring code: https://github.com/yext/answers-react-components/commit/d6bd209942bfc40be728ae14f40cca7381738980

Update documentation of RenderEntityPreviews for the expected usage of DropdownItem if user wants their entity preview to be navigable.

J=SLAP
TEST=manual

- check out previous commit of commit PR 168, and setup universal search with visual autocomplete. Clear recent searches in local storage. Input `Ryan` and see that there's no entity previews in dropdown even though it's in the network's response. Switch to the fixed commit 168 and the the latest change in main, see that entity previews are shown when inputting `Ryan`.
- Confirm that entity previews are rendered but NOT navigable without DropdownItem. 
- Wrapping results with DropdownItem, confirm I can tab through the entity previews

### DIFF
--- a/docs/answers-react-components.renderentitypreviews.md
+++ b/docs/answers-react-components.renderentitypreviews.md
@@ -15,5 +15,5 @@ export declare type RenderEntityPreviews = (autocompleteLoading: boolean, vertic
 
 ## Remarks
 
-An onSubmit function is provided to allow an entity preview to be submitted.
+For the entity previews to be navigable in the search bar's dropdown section, wrap each entity preview in a [DropdownItem()](./answers-react-components.dropdownitem.md) component.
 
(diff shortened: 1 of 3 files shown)


docs/answers-react-components.renderentitypreviews.md
```diff
@@ -15,5 +15,5 @@ export declare type RenderEntityPreviews = (autocompleteLoading: boolean, vertic
 
 ## Remarks
 
-An onSubmit function is provided to allow an entity preview to be submitted.
+For the entity previews to be navigable in the search bar's dropdown section, wrap each entity preview in a [DropdownItem()](./answers-react-components.dropdownitem.md) component.
```
nmanu1, line 18: It looks like only the last remark block is added to the docs, so we should include all of the remarks with only one tag

oshi97 (approved):

nmanu1 (approved):

PR #224, opened 2022-07-07
Fix static range filter bug

nmanu1:
This PR fixes a bug where range input static filters were removed when selecting/de-selecting facet options with different `fieldId`s. The expected behavior is that these static range filters are removed when conducting a new search (existing behavior), or only when a numerical facet option with the same `fieldId` is applied (fixed behavior).

J=SLAP-2213
TEST=manual

In the test-site, see that
- applying a static range filter, clearing the input, and then applying a numerical facet option for the same filter group removes the static range filter (existing behavior)
- applying a static range filter and then selecting or de-selecting a facet option from a different filter group leaves the static range filter as is (new behavior)
- conducting a new search removes facet options and static range filters, but not regular static filters (existing behavior)

### DIFF
--- a/src/components/NumericalFacets.tsx
+++ b/src/components/NumericalFacets.tsx
@@ -1,9 +1,9 @@
 import { FacetsProvider, RangeInput, RangeInputCssClasses } from './Filters';
 import { FilterGroup, FilterGroupCssClasses } from './FilterGroup';
 import { Fragment } from 'react';
-import { DisplayableFacet, NumberRangeValue } from '@yext/answers-headless-react';
+import { NumberRangeValue } from '@yext/answers-headless-react';
 import { StandardFacetsProps } from './StandardFacets';
-import { isNumberRangeFilter } from '../models/NumberRangeFilter';
+import { isNumericalFacet } from '../utils/filterutils';
 
 /**
  * The CSS class interface for {@link NumericalFacets}.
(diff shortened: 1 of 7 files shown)


src/utils/filterutils.tsx
```diff
@@ -84,16 +85,31 @@ function parseNumber(num: string) {
 }
 
 /**
- * Deselects the selected static number range filters in state.
+ * Deselects the selected static number range filters in state. If fieldIds are
+ * provided, only filters corresponding to one of those fieldIds are deselected.
+ * Otherwise, all selected filters are deselected.
  */
-export function clearStaticRangeFilters(answersActions: AnswersActions){
+export function clearStaticRangeFilters(answersActions: AnswersActions, fieldIds?: string[]) {
```
oshi97, line 92: maybe `getSelectedNumericalFacetFields` and the fieldIds param could be a Set instead of an array? Array is also good I don't think there's really much difference for what we're doing

nmanu1: sure, I can change to Set

src/components/NumericalFacets.tsx
```diff
@@ -88,7 +88,7 @@ export function NumericalFacets({
   );
 }
 
-function isNumericalFacet(facet: DisplayableFacet): boolean {
+export function isNumericalFacet(facet: DisplayableFacet): boolean {
```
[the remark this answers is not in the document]
nmanu1: moved!

oshi97 (approved):

PR #225, opened 2022-07-08
SearchBar Screenreader Fixes

ElemelonWind:
Added aria-label to `DropdownItem` to set screenreader text for result preview without user interference in `VisualAutocompleteConfig`. Changed some screenreader text to make it more consistent. Added `readAutocomplete` state so the autocomplete results don't get read when the user first toggles into the `SearchBar`.

J=SLAP-2234
TEST=auto, manual

- If initially focusing/clicking on the search bar: Reads the screenreader instructions AND
  - If no text is entered:
    - If there are no autocomplete options in the dropdown: No other announcement is made
    - If there are autocomplete options in the dropdown: Announces the number of options (always a number greater than 0)
  - If text is entered: Announces the number of options (could be 0 or any larger number)
- Typing or deleting characters should always announce the number of options (>= 0), even if the options (and their number) haven’t changed since the last time something was typed
- Clearing the search input should announce the new number of options (>= 0), even though there is no longer text in the input
- Using arrow keys or Tabs to focus on dropdown options should not announce the number of options available, it should only identify the option focused
- Selecting a dropdown option or clicking the search button should not announce the number of options available


src/components/Dropdown/DropdownItem.tsx
```diff
@@ -19,7 +19,9 @@ export type DropdownItemProps = PropsWithChildren<{
   /** Data associated with the dropdown item which is passed to the onClick and the onSelect handlers. */
   itemData?: Record<string, unknown> | undefined,
   /** A function which is fired when the item is clicked. */
-  onClick?: (value: string, index: number, focusedItemData: FocusedItemData | undefined ) => void
+  onClick?: (value: string, index: number, focusedItemData: FocusedItemData | undefined ) => void,
```
alextaing, line 22: nit: extra space before close parenthesis

oshi97:
could you add a description to your PR and also how you tested?

src/components/SearchBar.tsx
```diff
@@ -242,16 +246,20 @@ export function SearchBar({
...
   const handleInputFocus = useCallback((value = '') => {
+    setReadAutocomplete(false);
     answersActions.setQuery(value);
     updateEntityPreviews(value);
     autocompletePromiseRef.current = executeAutocomplete();
   }, [answersActions, autocompletePromiseRef, executeAutocomplete, updateEntityPreviews]);
 
   const handleInputChange = useCallback((value = '') => {
+    if (!readAutocomplete) {
+      setReadAutocomplete(true);
```
oshi97, line 257: Is there some way we can put this logic inside of the ScreenReader/Dropdown components? Ideally we don't have to add this logic to current and future components that use Dropdown. The SearchBar already has a LOT of stuff in it

ElemelonWind:
> could you add a description to your PR and also how you tested?

done :)

src/components/SearchBar.tsx
```diff
@@ -499,4 +509,17 @@ export function calculateEntityPreviewsCount(children: ReactNode): number {
     return c;
   });
   return count;
+}
+
+/**
+ * Sets screenreader text for navigable entity previews from a ReactNode containing DropdownItems.
+ */
+export function setPreviewScreenreader(children: ReactNode): ReactNode {
```
oshi97, line 517: Is there no other way to set the ariaLabel? We want to avoid low level methods like cloneElement and our recursivelyMapChildren (ideally we never have to use recurisvelymapChildren, there's also some issues with it currently)

ElemelonWind: With the way the `VisualAutocompleteConfig` is set up, the developer has to pass in their own `renderEntityPreviews` function and Nidhi and I decided we didn't want to leave it up to them to add an aria-label, so I go back and dynamically change the prop after the EntityPreviews have been rendered. Do you have any ideas on how to get around that?

nmanu1: It might also be worth checking with product on whether the developer should be in charge of this or if we should add it "magically"

oshi97: I was thinking we could update the onSubmit param in renderEntityPreview to instead be an object with default props people could use.

```tsx
{
  renderEntityPreviews: (isLoading, verticalKeyToResults, props) => {
    return verticalKeyToResults.people.map((r, index) => <DropdownItem {...props}>blah</DropdownItem>)
  }
}
```

oshi97: We could also add a default value to DropdownItem, though I didn't like that since that's putting SearchBar details into the Dropdown

ElemelonWind: both recent searches and autocomplete options have the "magical" screenreader text set up so I think what they wanted was for it to be consistent with entity previews as well. I can ask though!

oshi97: This is similar to a pattern react-bootstrap/restart-ui use, where they export a hook like `useDropdownItem` that provides props that people can use in whatever container element they want https://github.com/react-bootstrap/react-bootstrap/blob/af953c20361d91a50c06b75c8bc7d6bb13da3165/src/DropdownItem.tsx#L5

oshi97: oh I only just saw the other comments, but I imagine they will like things to be magical by default as well!

ElemelonWind: The default aria-label should probably just be the value of the item anyways; I'll add that in

nmanu1: I think, by default, the screenreader will read the value without a prefix even if no aria-label is specified

nmanu1 (commented):
When making changes to screenreader behavior, here are some expected behaviors that I like to test for with the regular search bar (vertical pages), the search bar with visual autocomplete (universal page), and filter search to make sure the new behavior is correct and to ensure there are no regressions.

- If initially focusing/clicking on the search bar: Reads the screenreader instructions AND
    - If no text is entered:
        - If there are no autocomplete options in the dropdown: No other announcement is made
        - If there are autocomplete options in the dropdown: Announces the number of options (always a number greater than 0)
    - If text is entered: Announces the number of options (could be 0 or any larger number)
- Typing or deleting characters should always announce the number of options (>= 0), even if the options (and their number) haven’t changed since the last time something was typed
- Clearing the search input should announce the new number of options (>= 0), even though there is no longer text in the input
- Using arrow keys or Tabs to focus on dropdown options should not announce the number of options available, it should only identify the option focused
- Selecting a dropdown option or clicking the search button should not announce the number of options available

Just to note, sometimes, even when there is no text entered in the search input, we can still have autocomplete options returned. And, sometimes on page load, we automatically populate the search bar with some input text by default, even though the user hasn’t typed anything

src/components/SearchBar.tsx
```diff
@@ -387,6 +389,11 @@ export function SearchBar({
         screenReaderText={screenReaderText}
         parentQuery={query}
         onToggle={handleToggleDropdown}
+        numItems={
+          (autocompleteResponse?.results.length || 0) +
```
oshi97, line 393: We already do this arithmetic above so I was thinking we could use a variable?

oshi97: Oh actually does Dropdown already know the numItems?

src/components/Dropdown/Dropdown.tsx
```diff
@@ -55,6 +57,7 @@ export function Dropdown(props: PropsWithChildren<DropdownProps>): JSX.Element {
   const containerRef = useRef<HTMLDivElement>(null);
   const screenReaderUUID: string = useMemo(() => uuid(), []);
   const [screenReaderKey, setScreenReaderKey] = useState<number>(0);
+  const [readAutocomplete, setReadAutocomplete] = useState<boolean>(false);
   const [childrenWithDropdownItemsTransformed, items] = getTransformedChildrenAndItemData(children);
```
oshi97, line 61: Could we use `items.length` instead of a new `numItems` prop? I think in the future we want to get rid of `getTransformedChildrenAndItemData` but until we do that I think we don't need an extra prop

src/components/SearchBar.tsx
ElemelonWind: I don't think so? It has `children` but I'm not sure how to find the size of that lol

src/components/Dropdown/Dropdown.tsx
ElemelonWind: ooh ok

src/components/Dropdown/Dropdown.tsx
```diff
@@ -136,7 +142,7 @@ export function Dropdown(props: PropsWithChildren<DropdownProps>): JSX.Element {
 
       <ScreenReader
         announcementKey={screenReaderKey}
-        announcementText={isActive ? screenReaderText : ''}
+        announcementText={isActive && (readAutocomplete || items.length) ? screenReaderText : ''}
```
oshi97, line 145: What cases is this `readAutocomplete` boolean necessary? Trying to wrap my head around it but having a hard time

ElemelonWind: So basically the screenreader isn't supposed to read the number of autocomplete options when the user first clicks into the searchbar, so readAutocomplete is set to true when the user starts typing and false when the user clicks out of the searchbar

ElemelonWind: If there are autocomplete suggestions then it should still read it even if the user hasn't started typing, hence the `(readAutocomplete || items.length > 0)` logic

nmanu1: The screenreader instructions are separate from the autocomplete announcement. We definitely want the instructions to read regardless of `readAutocomplete` when the user first starts interacting with the search bar. But, if there is some default input text before the user starts typing, the autocomplete announcement should also be made

src/components/Dropdown/Dropdown.tsx
```diff
@@ -55,6 +55,7 @@ export function Dropdown(props: PropsWithChildren<DropdownProps>): JSX.Element {
   const containerRef = useRef<HTMLDivElement>(null);
   const screenReaderUUID: string = useMemo(() => uuid(), []);
   const [screenReaderKey, setScreenReaderKey] = useState<number>(0);
+  const [readAutocomplete, setReadAutocomplete] = useState<boolean>(false);
```
oshi97, line 58: I think the name `readAutocomplete` is a little misleading. Even if `readAutocomplete` is true, the autocomplete won't necessarily be read, the same for the reverse `readAutocomplete` being false can still result in the autocomplete being read.
I would either
1. find a way to remove this variable (this option is preferred!! but might not be possible)
2. rename this variable
3. change the way this variable is set

src/components/SearchBar.tsx
```diff
@@ -95,7 +95,8 @@ export interface SearchBarCssClasses extends AutocompleteResultCssClasses {
 export type RenderEntityPreviews = (
   autocompleteLoading: boolean,
   verticalKeyToResults: Record<string, VerticalResultsData>,
-  onSubmit: (value: string, _index: number, itemData?: FocusedItemData) => void
+  handleSubmit: (value: string, _index: number, itemData?: FocusedItemData) => void,
+  returnEntityAria: (value: string) => string
```
oshi97, line 99: I might call this `getEntityAria` or `getAriaLabel` since I'm not sure what an entity aria is

src/components/Dropdown/Dropdown.tsx
```diff
@@ -202,6 +208,7 @@ function useDropdownContextInstance(
   index: number,
   focusedItemData: Record<string, unknown> | undefined,
   screenReaderUUID: string,
+  setReadAutocomplete: (boolean) => void,
```
oshi97, line 211: rename here too!

src/components/Dropdown/Dropdown.tsx
```diff
@@ -121,9 +123,13 @@ export function Dropdown(props: PropsWithChildren<DropdownProps>): JSX.Element {
       } else {
         toggleDropdown(false);
       }
+    } else if (!hasTyped) {
+      setHasTyped(true);
     }
   });
 
+  console.log(items.length);
```
oshi97, line 131: extra console.log

src/components/Dropdown/Dropdown.tsx
nmanu1: Not sure I understand, what would the other prop do?

nmanu1: Sometimes, when users navigate to a page, we have automatically populated text present in the search bar even before they have started typing, which runs a default search on the page. So when they click on the search bar for the first time, there's already text there

nmanu1: I think that should work!

src/components/Dropdown/Dropdown.tsx
```diff
@@ -136,7 +140,7 @@ export function Dropdown(props: PropsWithChildren<DropdownProps>): JSX.Element {
 
       <ScreenReader
         announcementKey={screenReaderKey}
-        announcementText={isActive ? screenReaderText : ''}
+        announcementText={isActive && (hasTyped || items.length || value) ? screenReaderText : ''}
```
oshi97, line 143: is it possible for `hasTyped` to be false but `value` to be true? I'm thinkin we can remove value here but you just added it so I might have missed something

src/components/SearchBar.tsx
```diff
@@ -233,7 +234,12 @@ export function SearchBar({
     executeEntityPreviewsQuery
   ] = useEntityPreviews(entityPreviewSearcher, entityPreviewsDebouncingTime);
   const { verticalKeyToResults, isLoading: entityPreviewsLoading } = entityPreviewsState;
-  const entityPreviews = renderEntityPreviews?.(entityPreviewsLoading, verticalKeyToResults, handleSubmit);
+  const entityPreviews = renderEntityPreviews?.(
+    entityPreviewsLoading,
+    verticalKeyToResults,
+    handleSubmit,
```
oshi97, line 240: Would you be ok with moving `handleSubmit` and `getAriaLabel` into an object? then somebody could just spread them into their DropdownItem, and we could also add more props here in the future instead of continually growing the number of arguments

src/components/Dropdown/Dropdown.tsx
ElemelonWind: Yeah Nidhi pointed out that in some cases there might already be something put in the search bar input by default before the user interacts with it, and in that case we still want the autocomplete to be read

src/components/SearchBar.tsx
ElemelonWind: In that case, would I have to change the names of it so it can be properly spread? (`handleSubmit` -> `onClick`, `getAriaLabel` -> `ariaLabel`)

oshi97: yes!

src/components/SearchBar.tsx
```diff
@@ -95,7 +95,10 @@ export interface SearchBarCssClasses extends AutocompleteResultCssClasses {
 export type RenderEntityPreviews = (
   autocompleteLoading: boolean,
   verticalKeyToResults: Record<string, VerticalResultsData>,
-  onSubmit: (value: string, _index: number, itemData?: FocusedItemData) => void
+  props: {
```
oshi97, line 98: Maybe we can call this `dropdownItemProps` instead of just `props`?

oshi97 (approved):
🔥 🔥 🔥

PR #228, opened 2022-07-11
Show pagination info in results count

alextaing:
Currently blocked by a larger issue.  Wrong number of results are showing on the test site.  Limit should be 10, but page 1 shows results 1-20, and page 2 shows results 11-30.  Showing duplicate results.

### DIFF
--- a/src/components/ResultsCount.tsx
+++ b/src/components/ResultsCount.tsx
@@ -7,6 +7,7 @@ import {
 import classNames from 'classnames';
 import { processTranslation } from './utils/processTranslation';
 import { useComposedCssClasses } from '../hooks/useComposedCssClasses';
+import { min } from 'lodash';
 
 /**
  *  The CSS class interface for {@link ResultsCount}.
@@ -57,6 +58,8 @@ export function ResultsCount({ customCssClasses }: ResultsCountProps): JSX.Eleme
 function useResultsCount() {
   const isVertical = useAnswersState(state => state.meta.searchType) === SearchTypeEnum.Vertical;
   const results = useAnswersState(state => isVertical ? state.vertical : state.universal.verticals);
+  const offset = useAnswersState(state => state.vertical.offset) || 0;
+  const limit = useAnswersState(state => state.vertical.limit) || 10;
   let resultsCount = 0;
   if (results) {
     if (isUniversalSearchResults(results)) {
@@ -73,7 +76,16 @@ function useResultsCount() {
     pluralForm: `${resultsCount} Results`,
     count: resultsCount
   });
-  return resultsCountText;
+
+  if (resultsCount > (limit)){
+    const paginateStart = offset + 1;
+    const paginateEnd = min([(offset + limit), resultsCount]);
+    const paginateRange = paginateStart + '-'+ paginateEnd;
+    const resultCountWithPaginationText = paginateRange + ' of ' + resultsCountText;
+    return resultCountWithPaginationText;
+  } else {
+    return resultsCountText;
+  }
 }
 
 function isUniversalSearchResults(data: VerticalResults[] | VerticalSearchState): data is VerticalResults[] {


alextaing:
When there are no results on the test site, it returns all possible "results," but no result count message.  Is there a reason why we don't want to return a `No results found.` message?

src/components/ResultsCount.tsx
```diff
@@ -73,7 +76,16 @@ function useResultsCount() {
     pluralForm: `${resultsCount} Results`,
     count: resultsCount
   });
-  return resultsCountText;
+
+  if (resultsCount > (limit)){
+    const paginateStart = offset + 1;
+    const paginateEnd = min([(offset + limit), resultsCount]);
```
oshi97, line 82: why do we need lodash for this? Math.min exists

src/components/ResultsCount.tsx
```diff
@@ -73,7 +76,16 @@ function useResultsCount() {
     pluralForm: `${resultsCount} Results`,
     count: resultsCount
   });
-  return resultsCountText;
+
+  if (resultsCount > (limit)){
```
oshi97, line 80: unnecessary parens around limit

src/components/ResultsCount.tsx
```diff
@@ -73,7 +76,16 @@ function useResultsCount() {
...
+    const paginateStart = offset + 1;
+    const paginateEnd = min([(offset + limit), resultsCount]);
+    const paginateRange = paginateStart + '-'+ paginateEnd;
+    const resultCountWithPaginationText = paginateRange + ' of ' + resultsCountText;
```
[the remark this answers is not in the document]
oshi97: I think template strings are sometimes overrated but I prefer it here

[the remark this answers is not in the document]
alextaing: Got it!

src/components/ResultsCount.tsx
alextaing: thanks

src/components/ResultsCount.tsx
alextaing: replaced

oshi97:
tests no happy

src/components/ResultsCount.tsx
```diff
@@ -57,6 +57,8 @@ export function ResultsCount({ customCssClasses }: ResultsCountProps): JSX.Eleme
 function useResultsCount() {
   const isVertical = useAnswersState(state => state.meta.searchType) === SearchTypeEnum.Vertical;
   const results = useAnswersState(state => isVertical ? state.vertical : state.universal.verticals);
+  const offset = useAnswersState(state => isVertical ? state.vertical.offset: null) || 0;
```
oshi97, line 60: I don't think you need a ternary operator here!

oshi97: If you default it to 0 immediately after there's no point of having it set to null

alextaing: Ok!  I noticed that without the ternary operator, it still works on the test site but it fails universal search tests since `state.vertical` is `undefined`.  I assume in real use, `state.vertical` won't be undefined? In that case, should I just edit the tests to define `vertical`?

alextaing: Thanks!

alextaing: Fixed, and tests edited to conform.

alextaing:
Note that Percy change is a result of changing the default limit from 10 to 20 in the `Pagination` component. Before, the default limit was 10, but 20 results showed on each page anyways (page 1 had results 1-20, page 2 had results 11-30).  This meant that many results were shown twice.  As a temporary fix, the default was changed to 20 in the `Pagination` component to align with the number of results returned per page.  Now, half the number of pages are required to show the same about of information.

src/components/ResultsCount.tsx
```diff
@@ -73,7 +75,16 @@ function useResultsCount() {
     pluralForm: `${resultsCount} Results`,
     count: resultsCount
   });
-  return resultsCountText;
+
+  if (resultsCount > limit && isVertical){
+    const paginateStart = offset + 1;
+    const paginateEnd = Math.min((offset + limit), resultsCount);
+    const paginateRange = `${paginateStart}-${paginateEnd}`;
```
oshi97, line 82: can we get a story with this

alextaing: of course, fo sho.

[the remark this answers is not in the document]
alextaing: Personally, I like it the way it is! I feel like with too much space, it's a lot to look at, especially when the numbers get larger (which I guess is not a super common/desirable use case).  I'll make the changes and we can compare them. 

image

oshi97 (approved):

tests/components/ResultsCount.test.tsx
```diff
@@ -114,7 +122,22 @@ describe('Results count for universal search', () => {
...
+    mockAnswersState(mockedStateVerticalPaginationResult);
+    render(<ResultsCount />);
+    expect(screen.getByText('1 - 20 of 30 Results')).toBeDefined();
+  });
+
+  it('Does not render pagination range if there is no pagination', () => {
+    mockAnswersState(mockedStateVerticalMultiple);
     render(<ResultsCount />);
-    expect(screen.queryByText('0 Results')).toBeNull();
+    expect(screen.queryByText('1-2 of 2 Results')).toBeNull();
```
oshi97, line 140: this text needs to be updated right?

oshi97 (changes requested):

tests/components/ResultsCount.test.tsx
oshi97: this is the danger with checking that something is null

oshi97: can we check that the page has no text instead?

tests/components/ResultsCount.test.tsx
```diff
@@ -114,7 +122,22 @@ describe('Results count for universal search', () => {
 
   it('Renders nothing if there is no result', () => {
     mockAnswersState(mockedStateUniversalNoResult);
+    const { container } = render(<ResultsCount />);
+    expect(container.childNodes[0]).toBeEmptyDOMElement();
```
yen-tt, line 126: Is there no other way to check that there's no text or if it's an empty div beside drilling into the container? I would rather not going into the specifics of a component's dom element if possible. If not, can we do container.firstChild instead?

alextaing: I tried firstChild, but I think it was upset since it wasn't a Testing Library method.  It said it didn't like direct node access. I can look into any other ways!

tests/components/ResultsCount.test.tsx
alextaing: yeh ty

alextaing: In this case, we are looking that the page has "2 Results" and not "1 - 2 of 2 Results" so the page would not be empty

tests/components/ResultsCount.test.tsx
alextaing: I tried an alternative method using regex, let me know if its OK!

tests/components/ResultsCount.test.tsx
alextaing: Instead, I found that there was an element that had text "2 Results" and I checked that the text content of that element was exactly "2 Results" and nothing else.

tests/components/ResultsCount.test.tsx
alextaing: Unfortunately after looking around, I've had no success finding a better way.  It generally does not like firstChild and gives a red squiggly, but it is OK with childNodes[0].  Is that OK, or should I continue the hunt?

oshi97 (approved):

yen-tt (approved):

PR #230, opened 2022-07-11
fix layout shift issue with filter/facet components

yen-tt:
Per product's feedback, update default width styling to be `w-full` and have user use a container element to control sizing instead. Update pages in test-site to have the div container of the filters/facets specify a fixed width.

J=SLAP-2214
TEST=manual

spin up test-site, toggle collapse/expand button on employee department facet on different screen size and see that layout remain fixed (not shifting when toggle).

### DIFF
--- a/test-site/src/pages/PeoplePage.tsx
+++ b/test-site/src/pages/PeoplePage.tsx
@@ -28,7 +28,7 @@ export function PeoplePage() {
     <div>
       <SearchBar />
       <div className='flex'>
-      <div className='w-56 shrink-0 mr-5'>
+        <div className='w-56 shrink-0 mr-5'>
           <NumericalFacets searchOnChange={false} />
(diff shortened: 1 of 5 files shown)


test-site/src/pages/PeoplePage.tsx
```diff
@@ -28,7 +28,7 @@ export function PeoplePage() {
     <div>
       <SearchBar />
       <div className='flex'>
-        <div className='min-w-fit pr-4'>
+      <div className='w-56 shrink-0 mr-5'>
```
oshi97, line 31: this new indentation is off right?

oshi97 (dismissed):

test-site/src/pages/PeoplePage.tsx
yen-tt: oops

PR #232, opened 2022-07-12
fix inputPrefix collision issue 

juliannzhou:
This PR fixes inputPrefix collision issue in RangeInput. The padding with inputPrefix is increased to avoid potential collision of max(min) input with the inputPrefix for up to three characters. The max/min input is also aligned to the right as approved by product.

J=SLAP-2212
TEST=manual

Tested the change by adding an inputPrefix of three characters (USD) to NumericalFacets. With padding increased, the max/min input no longer collides with inputPrefix. The min/max input was aligned to the right because of aesthetic reason approved by product - so without any input, the min/max placeholder and inputPrefix would not appear cramped together - input numbers appeared as expected with the right alignment.

### DIFF
--- a/src/components/Filters/RangeInput.tsx
+++ b/src/components/Filters/RangeInput.tsx
@@ -62,8 +62,8 @@ export interface RangeInputCssClasses {
 
 const builtInCssClasses: Readonly<RangeInputCssClasses> = {
   rangeInputContainer: 'flex flex-col',
-  input: 'w-full h-9 form-input cursor-pointer border rounded-md focus:ring-0 text-neutral-dark text-sm appearance-none leading-9',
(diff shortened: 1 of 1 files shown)


nmanu1 (commented):
in the Percy snapshots, it looks like the tooltips are overlapping with the range input starting a few PRs ago. is this the desired behavior?

juliannzhou:
> in the Percy snapshots, it looks like the tooltips are overlapping with the range input starting a few PRs ago. is this the desired behavior?

That was because the width styling for the facet and filter components were changed to `w-full` and became customizable, so the relative position for the tooltip no longer works when the user makes changes to the width. The tooltip positioning will be fixed in Yen's pr.

yen-tt (approved):

PR #234, opened 2022-07-13
fix tooltip location

yen-tt:
with rangeInput having `w-full`, dynamic width sizing, the manual top/right position styling adjustment no longer work if user change the width of rangeInput. The tooltip would overlap with the input boxes. Updated to ensure the tooltip always snap to the end of the rangeInput box
Screen Shot 2022-07-13 at 4 45 03 PM

J=none
TEST=manual&auto

see that tooltip is placed at the end of the rangeInput box no matter the sizing
Screen Shot 2022-07-13 at 4 47 54 PM
Screen Shot 2022-07-13 at 4 48 11 PM

### DIFF
--- a/src/components/Filters/RangeInput.tsx
+++ b/src/components/Filters/RangeInput.tsx
@@ -70,14 +70,14 @@ const builtInCssClasses: Readonly<RangeInputCssClasses> = {
   input___valid: 'border-gray-300 focus:border-primary',
   input___invalid: 'border-red-700 focus:border-red-700',
   inputContainer: 'relative',
-  inputRowContainer: 'flex flex-row items-center space-x-3 peer',
+  inputRowContainer: 'flex flex-row items-center space-x-3 group',
   buttonsContainer: 'flex flex-row items-center justify-between pt-2',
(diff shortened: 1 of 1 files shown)


oshi97 (approved):

src/components/Filters/RangeInput.tsx
```diff
@@ -202,14 +202,14 @@ export function RangeInput(props: RangeInputProps): JSX.Element | null {
         {renderInput(minRangeInput, handleMinChange, 'Min')}
         <div className='w-2.5 text-sm text-neutral'>-</div>
         {renderInput(maxRangeInput, handleMaxChange, 'Max')}
-      </div>
-      {isDisabled &&
+        {isDisabled &&
         <div className={cssClasses.tooltipContainer}>
```
oshi97, line 206: I would have thought eslint would want this indented, but I guess not?

yen-tt: nope, I guess this is technically a sibling node of the div above so it works either way if you want it indented or not

PR #235, opened 2022-07-13
update VerticalConfig to Accept a Generic Type Param

juliannzhou:
This PR makes changes to VerticalConfig to accept a generic type parameter for `CardComponent `

- UniversalResults and VerticalResults have to be updated to properly infer the typing of the provided CardComponent field
- The use of `as any` is due to typing issues with result type from `State` using `useSearchState`
- The use of `any` as the generic type for StandardCard is to avoid type conflict when using it as a default value for CardComponent field in places like StandardSection
- Added some custom components for testing purposes
- Updated useCard[Blah] hooks to accept generic types for Result

J=SLAP-2233
TEST=manual

Tested VerticalResults in PeoplePage:
- use CustomCard for CardComponent field. See that the typing works as expected and no error when starting the site
- see that using StandardCard still works
Tested UniversaResults in UniversalPage:
- See that using different CustomCard of different generic type for Result in `verticalConfigMap` produce the right mapping. The typing works as expected and no error when starting the site
- See that sectionComponent pass the generic type to the CardComponent works properly using CustomSection
- see that the useCard hooks work with Result of different type in CustomCard


tmeyer2115 (approved):
When merging this in, please update the commit message so that it doesn't say the `CardComponent` interface was updated.

src/models/verticalConfig.ts
```diff
@@ -7,11 +7,11 @@ import { VerticalLink } from './verticalLink';
  *
  * @public
  */
-export interface VerticalConfig {
+export interface VerticalConfig<T = Record<string, unknown>> {
```
tmeyer2115, line 10: Do we want `Record<string, any>` as the default instead? By having `unknown`, will it put a lot of burden on the user? Specifically, I'm wondering if they'll be forced to do manual, run-time checks in their cards. cc: @yen-tt

nmanu1 (commented):
Is the goal of this item to be able to use custom card components with type params specified inside our `UniversalResults` or `StandardSection` components? cc: @tmeyer2115

src/models/verticalConfig.ts
tmeyer2115: After talking with Yen and Davish, we do want `unknown`.

src/models/sectionComponent.ts
```diff
@@ -24,4 +24,4 @@ export interface SectionProps {
  *
  * @public
  */
-export type SectionComponent = (props: SectionProps) => JSX.Element | null;
+export type SectionComponent<T> = (props: SectionProps<T>) => JSX.Element | null;
```
nmanu1, line 27: if a user writes a custom section, will they be required to either specify a type or make it a generic?

src/components/cards/standard/StandardCard.tsx
```diff
@@ -8,7 +8,8 @@ import { StandardCardDisplay, StandardCardCssClasses } from './StandardCardDispl
  *
  * @public
  */
-export interface StandardCardProps extends CardProps {
+/* eslint-disable @typescript-eslint/no-explicit-any */
+export interface StandardCardProps extends CardProps<any> {
```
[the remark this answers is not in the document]
yen-tt: There's are type conflict issues when using StandardCard is a default for a `CardComponent` prop. For instance, in StandardSection, the generic type T would apply to `CardComponent` prop and `results` in props to be T. When CardComponent is not given, if StandardCard is used as the default with a type of DefaultRawDataType, we would have `results` with type T and a `CardComponent` with type DefaultRawDataType, causing a conflict since Type 'T' is not assignable to type 'DefaultRawDataType'. So, the generic type for StandardCardProps needs to be any for the card to be use as default.

Screen Shot 2022-07-25 at 9 27 52 AM

src/models/sectionComponent.ts
yen-tt: Updated to provide default typing!

src/components/cards/standard/StandardCard.tsx
```diff
@@ -2,13 +2,14 @@ import { useCardAnalyticsCallback } from '../../../hooks/useCardAnalyticsCallbac
...
 import { StandardCardDisplay, StandardCardCssClasses } from './StandardCardDisplay';
+import { DefaultRawDataType } from '../../../models';
 
 /**
  * Props for a StandardCard.
  *
  * @public
  */
-export interface StandardCardProps extends CardProps {
+export interface StandardCardProps<T = DefaultRawDataType> extends CardProps<T> {
```
nmanu1, line 12: does this need a default if we're specifying type `any` in `StandardCard` below? is it in case people want to extend this interface? 

should the same be done for `StandardSectionProps` to be consistent?

src/hooks/useCardAnalytics.ts
```diff
@@ -28,14 +29,17 @@ function isDirectAnswer(data: unknown): data is DirectAnswerData {
     (data as DirectAnswerData)?.type === DirectAnswerType.FieldValue;
 }
 
-export function useCardAnalytics(): (
-  cardResult: Result | DirectAnswerData, analyticsEventType: CardAnalyticsType
+export function useCardAnalytics<T = DefaultRawDataType>(): (
```
nmanu1, line 32: is the reason a default is needed here for the type param because of when the hook is used with `DirectAnswerData`?

src/hooks/useCardAnalytics.ts
```diff
@@ -28,14 +29,17 @@ function isDirectAnswer(data: unknown): data is DirectAnswerData {
...
+  cardResult: Result<T> | DirectAnswerData, analyticsEventType: CardAnalyticsType
 ) => void {
   const analytics = useAnalytics();
   const verticalKey = useSearchState(state => state.vertical.verticalKey);
   const queryId = useSearchState(state => state.query.queryId);
 
-  const reportCtaEvent = useCallback((result: DirectAnswerData | Result, eventType: CardCtaEventType) => {
+  const reportCtaEvent = useCallback((
+    result: DirectAnswerData | Result<T>,
```
nmanu1, line 40: can this union type be defined somewhere so we don't have to repeat it in multiple places?

src/components/cards/standard/StandardCard.tsx
yen-tt: yea, in case people extend it, but I don't think anyone really would. No harm to having a default though

src/hooks/useCardAnalytics.ts
yen-tt: thought we do export this so it would be good to have a default here, but we only export the callback wrapper functions (i.e. useCardAnalyticsCallback) which already have default so this is not needed. I can remove

src/hooks/useCardAnalytics.ts
yen-tt: sure

src/models/DefaultRawDataType.ts
```diff
@@ -0,0 +1,7 @@
+
+/**
+ * The default type for "rawData" field of type Result.
```
nmanu1, line 3: nit: can we link to `Result` or `rawData` here?

yen-tt: nope it would say something like `Unable to resolve reference "Result": The member reference "Result" was not found` since we don't export Result from component lib

nmanu1: ah right

nmanu1 (approved):

PR #242, opened 2022-07-18
Separate recent searches in verticals

alextaing:
This PR allows recent searches to show in verticals and keeps their recent search histories separate.

J=SLAP-2241
TEST=manual, auto

The test site was used to check that recent search histories in different verticals (and in universal) were all kept separate. Jest tests were written for the `useRecentSearches` hook, testing that recent searches appear, that they stay on their own verticals, and that they can be cleared using the returned `clearRecentSearches` function.

### DIFF
--- a/src/hooks/useRecentSearches.ts
+++ b/src/hooks/useRecentSearches.ts
@@ -1,57 +1,65 @@
-import { useCallback, useEffect, useRef, useState } from 'react';
+import { useCallback, useState } from 'react';
 import RecentSearches, { ISearch } from 'recent-searches';
-
-// export const RECENT_SEARCHES_KEY = '__yxt_recent_searches__';
 
 export function useRecentSearches(
   recentSearchesLimit: number,
   verticalKey: string|null,
   isVertical: boolean
-): [ISearch[]|undefined, (input: string) => void, () => void] {
-  const recentSearchesKey = getRecentSearchesKey();
-  const recentSearchesLimitRef = useRef(recentSearchesLimit);
-  const [ recentSearches, setRecentSeaches ] = useState<RecentSearches>(
-    new RecentSearches({
-      limit: recentSearchesLimit,
-      namespace: recentSearchesKey
-    })
-  );
-  console.log(recentSearches.getRecentSearches());
-
+): [ISearch[]|undefined, (input: string, verticalKey: string|null, isVertical: boolean) => void, () => void] {
+  const [recentSearches, setRecentSeaches] = useState<{ [key: string]: RecentSearches }>({});
+  const [recentVerticalKeys, setRecentVerticalsKeys] = useState<string[]>([]);
   const clearRecentSearches = useCallback(() => {
-    localStorage.removeItem(recentSearchesKey);
-    setRecentSeaches(new RecentSearches({
-      limit: recentSearchesLimit,
-      namespace: recentSearchesKey
-    }));
-    localStorage.removeItem(recentSearchesKey);
-  }, [recentSearchesKey, recentSearchesLimit]);
-
-  const setRecentSearch = useCallback((input: string) => {
-    recentSearches?.setRecentSearch(input);
-  }, [recentSearches]);
-
-  useEffect(() => {
-    if (recentSearchesLimit !== recentSearchesLimitRef.current) {
-      setRecentSeaches(new RecentSearches({
-        limit: recentSearchesLimit,
-        namespace: recentSearchesKey
-      }));
-      recentSearchesLimitRef.current = recentSearchesLimit;
+    setRecentSeaches({});
+    for (const verticalKey in recentVerticalKeys){
+      localStorage.removeItem(verticalKey);
     }
-  }, [recentSearchesKey, recentSearchesLimit]);
+  }, [recentVerticalKeys]);
 
-  function getRecentSearchesKey(): string {
-    if (isVertical) {
-      if (verticalKey) {
-        return `__yxt_recent_searches_${verticalKey}__`;
+  const setRecentSearch = useCallback((input: string, verticalKey: string|null, isVertical: boolean) => {
+    const recentSearchesKey = getRecentSearchesKey(verticalKey, isVertical);
+    if (recentVerticalKeys && recentSearches) {
+      if (recentSearchesKey in recentVerticalKeys) {
+        recentSearches[recentSearchesKey].setRecentSearch(input);
       } else {
-        console.error('No vertical key found on this vertical');
-        return '';
+        const recentSearchesKey = getRecentSearchesKey(verticalKey, isVertical);
+        const newRecentVerticalKeys = recentVerticalKeys.concat(recentSearchesKey);
+        setRecentVerticalsKeys(newRecentVerticalKeys);
+        setRecentSeaches({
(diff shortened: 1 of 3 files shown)


alextaing:
Are there any suggestions on writing automated tests for this item?  Is it something that is best done with unit tests, or is its scope a little too wide?

src/hooks/useRecentSearches.ts
```diff
@@ -1,41 +1,49 @@
...
-  recentSearchesLimit: number
-): [ISearch[]|undefined, (input: string) => void, () => void] {
+  recentSearchesLimit: number,
+  verticalKey: string | null
+): [ISearch[] | undefined, (input: string) => void, () => void] {
   const recentSearchesLimitRef = useRef(recentSearchesLimit);
-  const [ recentSearches, setRecentSeaches ] = useState<RecentSearches>(
+  const [recentSearchesKey, setRecentSearchesKey] = useState(getRecentSearchesKey(verticalKey));
+  const [recentSearches, setRecentSeaches] = useState<RecentSearches>(
     new RecentSearches({
```
oshi97, line 11: I feel maybe like this RecentSearches instance should be in a ref instead of in state?

src/hooks/useRecentSearches.ts
```diff
@@ -1,41 +1,49 @@
...
-
 export function useRecentSearches(
-  recentSearchesLimit: number
-): [ISearch[]|undefined, (input: string) => void, () => void] {
+  recentSearchesLimit: number,
+  verticalKey: string | null
+): [ISearch[] | undefined, (input: string) => void, () => void] {
   const recentSearchesLimitRef = useRef(recentSearchesLimit);
-  const [ recentSearches, setRecentSeaches ] = useState<RecentSearches>(
+  const [recentSearchesKey, setRecentSearchesKey] = useState(getRecentSearchesKey(verticalKey));
```
oshi97, line 9: can you look into removing the `useState` for recentSearchesKey and also the `useRef` for recentSearchesLimit? I think since the effect already updates the RecentSearches instance in state we probably don't need them?

alextaing: removed! they work fine without them!

src/hooks/useRecentSearches.ts
```diff
@@ -1,41 +1,49 @@
...
+    localStorage.removeItem(recentSearchesKey);
+  }, [recentSearchesKey, recentSearchesLimit]);
 
   const setRecentSearch = useCallback((input: string) => {
-    recentSearches?.setRecentSearch(input);
+    recentSearches.setRecentSearch(input);
   }, [recentSearches]);
 
   useEffect(() => {
-    if (recentSearchesLimit !== recentSearchesLimitRef.current) {
```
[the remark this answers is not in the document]
alextaing: removed

tests/hooks/useRecentSearches.test.tsx
```diff
@@ -0,0 +1,86 @@
...
+    useRecentSearches(10, verticalKey)
+  );
+  const [, setRecentSearch, clearRecentSearches] = result.current;
+  setRecentSearch('bob');
+  setRecentSearch('carrie');
+  setRecentSearch('williard');
+
+  rerender();
+  const [recentSearches] = result.current;
+  expect(recentSearches?.length).toBe(3);
```
oshi97, line 16: let's also check the contents of the recent searches, not just that the length is 3. It's probably fine if this is the only test for the actual contents

alextaing:
Current coveralls error is due to 504 gateway error, will try to rerun coveralls in an hour

tests/hooks/useRecentSearches.test.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+  clearPeopleSearches();
+});
+
+it('preserves recent searches when returning to a vertical', () => {
+  let verticalKey = 'people';
+  const { result, rerender } = renderHook(() =>
+    useRecentSearches(10, verticalKey)
+  );
+  const [, setRecentPeopleSearch, clearPeopleSearches] = result.current;
```
oshi97, line 45: we can probably remove the expects from the first half of this test, since the first half seems to be a copy paste of the test above it?

oshi97: alternatively you could remove the test above this one

tests/hooks/useRecentSearches.test.tsx
```diff
@@ -0,0 +1,86 @@
...
+  rerender();
+  recentSearches = result.current[0];
+  expect(recentSearches?.length).toBe(1);
+
+  verticalKey = 'people';
+  rerender();
+  recentSearches = result.current[0];
+  expect(recentSearches?.length).toBe(2);
+
+  clearPeopleSearches();
```
oshi97, line 68: are we clearing recent searches at the end of each test to avoid having them influence each other?

oshi97: if that's the reason could we clear localStorage directly in a beforeEach()?

src/hooks/useRecentSearches.ts
```diff
@@ -1,41 +1,45 @@
...
-        namespace: RECENT_SEARCHES_KEY
-      }));
-      recentSearchesLimitRef.current = recentSearchesLimit;
-    }
-  }, [recentSearchesLimit]);
+    setRecentSeaches(new RecentSearches({
+      limit: recentSearchesLimit,
+      namespace: recentSearchesKey
+    }));
+  }, [recentSearchesKey, recentSearchesLimit, verticalKey]);
```
oshi97, line 34: We can remove verticalKey as a dep right?

src/hooks/useRecentSearches.ts
```diff
@@ -1,41 +1,45 @@
...
 
-export const RECENT_SEARCHES_KEY = '__yxt_recent_searches__';
-
 export function useRecentSearches(
-  recentSearchesLimit: number
-): [ISearch[]|undefined, (input: string) => void, () => void] {
-  const recentSearchesLimitRef = useRef(recentSearchesLimit);
-  const [ recentSearches, setRecentSeaches ] = useState<RecentSearches>(
+  recentSearchesLimit: number,
+  verticalKey: string | null
```
oshi97, line 6: It might be sliightly simpler to use undefined instead of null here? I like that null is more "clear" than undefined, but accepting undefined here instead let's us get rid of defaulting verticalKey to null above in `SearchBar`(currently line 190)

tests/hooks/useRecentSearches.test.tsx
alextaing: done!

tests/hooks/useRecentSearches.test.tsx
alextaing: That makes a lot of sense!

src/hooks/useRecentSearches.ts
alextaing: thanks for catching that!

src/hooks/useRecentSearches.ts
alextaing: I see, that looks better

oshi97 (approved):
💯 💯 💯

PR #245, opened 2022-07-19
Add showMoreLimit prop

ElemelonWind:
Add a `showMoreLimit` prop to `StandardFacet`. When the number of options exceeds this limit, display only that many options, a searchable facets input, and a show more button. The show more button, when clicked displays all the options. If the number of options is less than the limit, show all the options and no search bar

J=SLAP-2252
TEST=auto

Added two new stories: one to show the prop at work, and the second to show what happens when the show more button is pressed

Note: Coveralls decreased by 0.3% because it doesn't count stories as test coverage


src/components/StandardFacets.tsx
```diff
@@ -1,6 +1,6 @@
 import { FacetsProvider } from './Filters';
 import { FilterGroup, FilterGroupCssClasses } from './FilterGroup';
-import { Fragment } from 'react';
+import { Fragment, useState } from 'react';
```
alextaing, line 3: already noted by github, but looks like `useState` is unused

tests/components/StandardFacets.stories.tsx
```diff
@@ -34,3 +34,23 @@ export const NoOptionCounts = (args: StandardFacetsProps) => {
...
+
+export const ShowMoreLimit = (args: StandardFacetsProps) => {
+  return (
+    <SearchHeadlessContext.Provider value={generateMockedHeadless(mockedHeadlessState)}>
+      <StandardFacets {...args} showMoreLimit={1} />
+    </SearchHeadlessContext.Provider>
+  );
+};
+
+export const ShowMoreLimitClicked = (args: StandardFacetsProps) => {
```
nmanu1, line 46: since the returned JSX is the same as for `ShowMoreLimit`, I think you can just bind to that and add a `play` function. we do this in a couple other stories if you want to see an example

src/components/FilterGroup.tsx
```diff
@@ -85,15 +95,27 @@ export function FilterGroup({
...
+          ? filterOptions.map(o => {
+            return (
+              <CheckboxOption
+                {...o}
+                key={o.displayName || o.value.toString()}
+                customCssClasses={cssClasses}
+              />
+            );
+          })
+          : filterOptions.slice(0, showMoreLimit).map(o => {
```
nmanu1, line 108: since the `map` function's body is the same for both of these, can we consolidate them by moving the ternary operator within the `slice` function's params? (e.g. something like `filterOptions.slice(0, showAll ? filterOptions.length : showMoreLimit).map(...)`)

ElemelonWind: ooooh smart ty!

src/components/FilterGroup.tsx
```diff
@@ -94,6 +104,7 @@ export function FilterGroup({
             />
           );
         })}
+        {limited && <button className='text-blue-500 py-1 text-sm' onClick={toggleShowAll}>{showAll ? 'Show Less' : 'Show More'}</button>}
```
yen-tt, line 107: instead of text-blue-500, let's update it to use the corresponding custom `primary` color

src/components/FilterGroup.tsx
```diff
@@ -69,6 +72,13 @@ export function FilterGroup({
     };
   }, [customCssClasses]);
 
+  const limited = filterOptions.length > showMoreLimit;
+  const [showAll, setShowAll] = useState<boolean>(!limited);
+
+  function toggleShowAll() {
```
oshi97, line 78: At first I thought we might want to wrap this in a `useCallback`, if only for consistency, but after thinking about it more we can probably omit it.

Out of curiosity, if you use an arrow function instead like `onClick={e => setShowAll(!showAll)}` does a react-perf eslint rule complain? I was thinking that it might be better to use an arrow function and an eslint-ignore comment, just to be explicit here. 

Since toggleShowAll is just passed into a regular `button` and not another component I don't think there's benefits of wrapping this in useCallback

src/components/FilterGroup.tsx
```diff
@@ -104,7 +100,8 @@ export function FilterGroup({
             />
           );
         })}
-        {limited && <button className='text-blue-500 py-1 text-sm' onClick={toggleShowAll}>{showAll ? 'Show Less' : 'Show More'}</button>}
+        {/* eslint-disable-next-line react-perf/jsx-no-new-function-as-prop */}
+        {limited && <button className='text-primary py-1 text-sm' onClick={_e => setShowAll(!showAll)}>{showAll ? 'Show Less' : 'Show More'}</button>}
```
oshi97, line 104: Can this be multiple lines? You can also replace the `_e` with `()`

ElemelonWind: not sure if I formatted it right lol, could you take a look?

src/components/FilterGroup.tsx
ElemelonWind: looks like that failed the percy build, could you approve it?

src/components/FilterGroup.tsx
```diff
@@ -85,7 +91,7 @@ export function FilterGroup({
       {renderTitle()}
       <CollapsibleSection className={cssClasses.optionsContainer}>
         {searchable && <SearchInput className={cssClasses.searchInput} />}
-        {filterOptions.map(o => {
+        {filterOptions.slice(0, showAll ? filterOptions.length : showMoreLimit).map(o => {
```
nmanu1, line 94: did you test if this works properly when typing in the search input? it seems like the search would only be performed on this slice of filter options instead of searching all of the options and then displaying the first few

src/components/FilterGroup.tsx
```diff
@@ -94,6 +100,11 @@ export function FilterGroup({
             />
           );
         })}
+        {limited &&
+        /* eslint-disable-next-line react-perf/jsx-no-new-function-as-prop */
+        <button className='text-primary py-1 text-sm' onClick={() => setShowAll(!showAll)}>
```
alextaing, line 105: Instead of using the eslint disable, can we extract the function and call it in onClick?

src/components/FilterGroup.tsx
```diff
@@ -69,6 +72,9 @@ export function FilterGroup({
     };
   }, [customCssClasses]);
 
+  const limited = filterOptions.length > showMoreLimit;
```
alextaing, line 75: Super subjective, but since this is a boolean it might be a tad more legible for the future if it were called `isLimited`-- but its tiny haha

src/components/FilterGroup.tsx
oshi97: I wanted to specifically use the eslint disable instead of using useCallback.

Extracting it to a function doesn't actually give it a stable reference, you have to use useCallback or pull the function out to the top level.
I'm fine with useCallback as well I just think it hurts performance a little because we don't actually need the memoization. But performance is not really a concern here!

src/components/FilterGroup.tsx
```diff
@@ -94,6 +100,11 @@ export function FilterGroup({
             />
           );
         })}
+        {limited &&
+        /* eslint-disable-next-line react-perf/jsx-no-new-function-as-prop */
```
oshi97, line 104: I think it's a little nicer if we indent everything after the && in this conditional (lines 104-107) and move the last brace to the next line

src/components/FilterGroup.tsx
alextaing: ohhh I see-- I totally missed that comment earlier. Thanks

oshi97: @yen-tt @nmanu1 do you guys prefer `useCallback` here? I'm really 50/50 and it's probably not a decision worth stressing over but just wondering

ElemelonWind:
I'll be removing the `shouldRenderOption` logic from `CheckboxOption` and putting it into `CheckboxOptions` after demos but wanted to push in case people had any feedback for the new internal component

src/components/FilterGroup.tsx
nmanu1: I think I have a slight preference for not using `useCallback` here. just because it's not really needed and I'm in favor of not wrapping in `useMemo`/`useCallback` unless it adds some value

src/components/FilterGroup.tsx
```diff
@@ -85,17 +94,57 @@ export function FilterGroup({
...
+          </button>
+        }
         {children}
       </CollapsibleSection>
     </FilterGroupProvider>
   );
 }
+
+function CheckboxOptions({
+  filterOptions,
```
oshi97, line 115: Do these parameters have type `any` if you don't specify a type?

src/components/FilterGroup.tsx
```diff
@@ -85,17 +94,57 @@ export function FilterGroup({
...
+    if (!searchUtilities.isCloseMatch(option.displayName, searchValue)) {
+      return false;
+    }
+
+    return true;
+  }, [searchUtilities, searchValue]);
+
+  return (
+    <>
+      {filterOptions.filter(o => {
```
oshi97, line 137: I think this can just be `filterOptions.filter(shouldRenderOption)`

src/components/FilterGroup.tsx
```diff
@@ -85,17 +94,57 @@ export function FilterGroup({
...
+
+function CheckboxOptions({
+  filterOptions,
+  showMoreLimit,
+  cssClasses,
+  showAll
+}) {
+  const searchUtilities = useSearchUtilities();
+  const { searchValue } = useFilterGroupContext();
+  const shouldRenderOption = useCallback((option: { displayName: string, value: unknown }) => {
```
oshi97, line 122: does this need a useCallback? We're not passing this function anywhere

src/components/FilterGroup.tsx
```diff
@@ -85,17 +94,57 @@ export function FilterGroup({
...
+            {showAll ? 'Show Less' : 'Show More'}
+          </button>
+        }
         {children}
       </CollapsibleSection>
     </FilterGroupProvider>
   );
 }
+
+function CheckboxOptions({
```
oshi97, line 114: Would it make sense to put this component in a separate file?

ElemelonWind: Tom said since it's an internal component we can just keep it in the same file

src/components/FilterGroup.tsx
ElemelonWind: Should I make an interface for it?

oshi97: yeah

oshi97: You could also specify it inline

oshi97: like  `{ myParam }: { myParam: string }`

src/components/FilterGroup.tsx
```diff
@@ -119,7 +120,12 @@ function CheckboxOptions({
 }) {
   const searchUtilities = useSearchUtilities();
   const { searchValue } = useFilterGroupContext();
-  const shouldRenderOption = useCallback((option: { displayName: string, value: unknown }) => {
+  const shouldRenderOption = (option:
+  { displayName?: unknown, value: string | number | boolean | NumberRangeValue }) => {
```
tmeyer2115: Instead of `unknown`, could we do `string | undefined` here? I can't remember off the top of my head, but we may also be able to do `displayName?: string`. The `string | undefined` may be redundant because of the optional nature of `displayName.

src/components/FilterGroup.tsx
```diff
@@ -119,7 +120,12 @@ function CheckboxOptions({
 }) {
   const searchUtilities = useSearchUtilities();
   const { searchValue } = useFilterGroupContext();
-  const shouldRenderOption = useCallback((option: { displayName: string, value: unknown }) => {
+  const shouldRenderOption = (option:
```
tmeyer2115: @oshi97 do we not want to wrap this in a `useMemo` or a `useCallback`? I'm wondering if, as is, we could erroneously re-render if `searchValue` and `searchUtilities` remained the same.

src/components/FilterGroup.tsx
```diff
@@ -85,17 +96,62 @@ export function FilterGroup({
...
+  ) => {
+    if (isUndefined(option.displayName)) {
+      option.displayName = option.value.toString();
+    }
+
+    if (!searchUtilities.isCloseMatch(option.displayName, searchValue)) {
+      return false;
+    }
+
+    return true;
```
yen-tt, line 141: I think we can directly return `searchUtilities.isCloseMatch(option.displayName, searchValue)` here?

src/components/FilterGroup.tsx
```diff
@@ -85,17 +96,62 @@ export function FilterGroup({
...
+  };
+
+  return (
+    <>
+      {filterOptions.filter(shouldRenderOption).map(o => {
+        return (
+          <CheckboxOption
+            {...o}
+            key={o.displayName || o.value.toString()}
```
yen-tt, line 150: since we did `option.displayName = option.value.toString();` in `shouldRenderOption`, I think it can just be `key={o.displayName}` here

src/components/FilterGroup.tsx
```diff
@@ -85,17 +96,62 @@ export function FilterGroup({
...
+  filterOptions: FilterOptionConfig[],
+  showMoreLimit: number,
+  cssClasses: CheckboxCssClasses,
+  showAll: boolean
+}) {
+  const searchUtilities = useSearchUtilities();
+  const { searchValue } = useFilterGroupContext();
+
+  const shouldRenderOption = (
+    option: { displayName?: string, value: FilterOptionConfig['value'] }
```
nmanu1, line 131: can `option` just have type `FilterOptionConfig` instead of defining each field again?

src/components/FilterGroup.tsx
```diff
@@ -85,17 +96,57 @@ export function FilterGroup({
...
-        })}
+        <CheckboxOptions
+          filterOptions={filterOptions}
+          showMoreLimit={showMoreLimit}
+          cssClasses={cssClasses}
+          showAll={showAll} />
+        {isLimited &&
+          /* eslint-disable-next-line react-perf/jsx-no-new-function-as-prop */
+          <button className='text-primary py-1 text-sm' onClick={() => setShowAll(!showAll)}>
+            {showAll ? 'Show Less' : 'Show More'}
```
nmanu1, line 107: would this scenario be handled correctly?
- `filterOptions` > `showMoreLimit`, so the options are searchable and have a Show More button displayed
- user types in the input which narrows down the matching options to fewer than `showMoreLimit`
- there probably shouldn't be a Show More/Less button displayed anymore

src/components/FilterGroup.tsx
```diff
@@ -85,17 +96,69 @@ export function FilterGroup({
...
+      <CheckboxOption
+        {...o}
+        key={o.displayName}
+        customCssClasses={cssClasses}
+      />
+    );
+  });
+
+  if (displayedOptions.length > showMoreLimit) {
+    displayedOptions = displayedOptions.slice(0, showAll ? filterOptions.length : showMoreLimit);
```
nmanu1, line 153: nit: it would be more accurate for the ternary operator to resolve to `displayedOptions.length` rather than `filterOptions.length` when `showAll` is true

src/components/FilterGroup.tsx
```diff
@@ -85,17 +93,61 @@ export function FilterGroup({
...
+  showMoreLimit: number,
+  cssClasses: CheckboxCssClasses
+}) {
+  const searchUtilities = useSearchUtilities();
+  const { searchValue } = useFilterGroupContext();
+
+  const shouldRenderOption = (
+    option: FilterOptionConfig
+  ) => {
+    if (isUndefined(option.displayName)) {
```
nmanu1, line 121: nit: can we just have `option.displayName = option.displayName || option.value.toString()` instead of this check?

tmeyer2115 (approved):

PR #247, opened 2022-07-20
Update Snyk transitive dependencies.

tmeyer2115:
Multiple transitive dependencies of Storybook were flagged as Snyk vulnerabilities: unset-value, glob-parent, and file-system-cache. I manually edited the package-lock.json to bump the versions of these dependencies. The new versions do not have the Snyk flagged vulnerabilities. In both cases, I had to bump the version to a new major. I read through the release notes for both transitive dependencies and the breaking changes should not impact us. 

TEST=manual

Built and served the Storybook site. Everything worked as expected.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10378,7 +10378,7 @@
         "set-value": "^2.0.0",
         "to-object-path": "^0.3.0",
         "union-value": "^1.0.0",
-        "unset-value": "^1.0.0"
+        "unset-value": "^2.0.1"
       },
       "engines": {
         "node": ">=0.10.0"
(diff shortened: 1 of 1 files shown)


package-lock.json
```diff
@@ -11327,7 +11327,7 @@
       "dependencies": {
         "@mrmlnc/readdir-enhanced": "^2.2.1",
         "@nodelib/fs.stat": "^1.1.2",
-        "glob-parent": "^3.1.0",
+        "glob-parent": "^5.1.2",
```
yen-tt, line 11330: I see that glob-parent have v6 as the latest major, do we want to bump to that? it got some bug fix to resolve a ReDoS vulnerability but I guess Snyk doesn't care about this particular issue

tmeyer2115: Snyk recommended we switch to 5.1.2. Maybe they don't yet know about v6. If it's okay, I'll stick with v5 for now since that's what I vetted through glob-parent's release notes.

yen-tt: ah ok, yeah sounds good!

yen-tt (approved):

PR #248, opened 2022-07-20
make option counts optional

alextaing:
This PR provides users the option to turn off option counts on standard filters.  An optional prop, `showOptionCounts` was added to the `StandardFacets` prop interface.  If unspecified, it defaults to true.

J=SLAP-2251
TEST=auto, manual

It was tested that when `showOptionCounts` was set to false, option counts were not displayed. When it was set to true, or omitted entirely, option counts were displayed.
A jest test was written to show that when `setOptionsFilters` is false, then the counts do not display. 
A Storybook story was added that displayed standard facets without option counts.


src/components/StandardFacets.tsx
```diff
@@ -32,6 +32,8 @@ export interface StandardFacetsProps {
   searchOnChange?: boolean,
   /** List of filter ids that should not be displayed. */
   excludedFieldIds?: string[],
+  /** Whether or not to show the option counts for each filter */
```
yen-tt, line 35: nit: period

yen-tt, line 35: nit: period

tests/components/StandardFacets.test.tsx
```diff
@@ -75,6 +75,17 @@ describe('StandardFacets', () => {
     expectFacetOptionSet(actions, productFacet.fieldId, productFacet.options[0], true);
   });
 
+  it('does not display option counts if setOptionCounts is set to false', () => {
```
yen-tt, line 78: nit: capitalize

src/components/StandardFacets.tsx
```diff
@@ -50,7 +52,12 @@ export interface StandardFacetsProps {
  * @public
  */
 export function StandardFacets(props: StandardFacetsProps) {
-  const { searchOnChange, excludedFieldIds = [], customCssClasses = {}, ...filterGroupProps } = props;
+  const {
+    searchOnChange,
+    excludedFieldIds = [],
+    customCssClasses = {},
+    showOptionCounts = true,
+    ...filterGroupProps } = props;
```
oshi97, line 60: line break after filterGroupProps

tests/components/StandardFacets.test.tsx
```diff
@@ -75,6 +75,17 @@ describe('StandardFacets', () => {
...
+    render(<StandardFacets showOptionCounts={false}/>);
+
+    const facets = DisplayableFacets[0];
+    const coffeeLabel = screen.queryByLabelText(facets.options[0].displayName);
+    const coffeeCount = screen.queryByLabelText(facets.options[0].count);
+
+    expect(coffeeLabel).toBeDefined();
+    expect(coffeeCount).toBeNull();
```
yen-tt, line 86: I believeee this assertion wouldn't help since screen.queryByLabelText doesn't do for partial text match, so `stuff (0)` vs `stuff` for the label would still pass for this assertion since it will always be null. You may want to build the string as if showOptionCounts is true and check that instead

src/components/StandardFacets.tsx
```diff
@@ -61,7 +68,7 @@ export function StandardFacets(props: StandardFacetsProps) {
               <FilterGroup
                 fieldId={f.fieldId}
                 filterOptions={f.options.map(o => {
-                  return { ...o, resultsCount: o.count };
+                  return showOptionCounts ? { ...o, resultsCount: o.count } : { ...o };
```
yen-tt, line 71: can we just return `o` if it's false? no need to perform spread assignment (or maybe do mapping only if showOptionCounts is true)

src/components/StandardFacets.tsx
alextaing: got it

src/components/StandardFacets.tsx
alextaing: thx

tests/components/StandardFacets.test.tsx
alextaing: thank

tests/components/StandardFacets.test.tsx
alextaing: OK! I changed it for safely-- but I think that queryByLabelText does do partial text match!

src/components/StandardFacets.tsx
alextaing: True!  I changed it to return `o` on false

tests/components/StandardFacets.test.tsx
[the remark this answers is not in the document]
alextaing: good catch, thanks!

yen-tt (approved):

PR #250, opened 2022-07-22
Rename restrictVerticals to includedVerticals

nmanu1:
This PR renames `restrictVerticals` to `includedVerticals` in the `VisualAutocompleteConfig`.

J=SLAP-2255
TEST=manual

Spin up the test-site and see that VA still works on the universal page.


alextaing:
I'm thinking that it would make more sense to have this variable name be a noun instead of a verb since it's an array-- a verb tends to make me think of a function. `includedVerticals` doesn't quite roll off the tongue, though.  Is having the name a noun instead on the table? Of course, not a big deal.

nmanu1:
> I'm thinking that it would make more sense to have this variable name be a noun instead of a verb since it's an array-- a verb tends to make me think of a function. `includedVerticals` doesn't quite roll off the tongue, though. Is having the name a noun instead on the table? Of course, not a big deal.

I'd prefer a noun as well. `includeVerticals` is what product wanted, but I don't think they would have a problem with `includedVerticals`. I'll update it and run it by Tom on Monday just to make sure

tmeyer2115 (approved):

PR #251, opened 2022-07-22
Fix search bar overlap

alextaing:
This PR changes the search bar component such that its dropdown menu is not overlapped by other search bars when it is active.  This is done by keeping an inactive search bar on `z-0` with all other components. But when it is activated by focusing on it, the search bar and the dropdown menu change to `z-10`, "lifting" them from the page.

J=SLAP-2253
TEST=manual

It was tested that when two search bars were placed one over another, that the dropdown menu of the higher search bar displayed over the lower one.

image


test-site/src/pages/PeoplePage.tsx
```diff
@@ -28,6 +28,7 @@ export function PeoplePage() {
 
   return (
     <div>
+      <SearchBar />
```
nmanu1, line 31: nit: I think this was just for testing purposes?

alextaing: ah! Thanks.

nmanu1 (approved):

PR #252, opened 2022-07-23
remove duplicate filters

juliannzhou:
This PR removes duplicate filters when same filter exists in both applied facet and applied static filters.

SLAP=J-2248
TEST=manual


src/components/AppliedFiltersDisplay.tsx
```diff
@@ -109,6 +112,18 @@ export function AppliedFiltersDisplay(props: AppliedFiltersDisplayProps): JSX.El
 
   const handleRemoveStaticFilterOption = (filter: DisplayableFilter) => {
```
nmanu1, line 113: what happens if the filter is removed by clicking on the checkbox option for the static filter or the facet? should the duplicate filter also be removed in that case?

src/components/AppliedFilters.tsx
```diff
@@ -70,6 +70,19 @@ export function AppliedFilters(props: AppliedFiltersProps): JSX.Element {
     hierarchicalFacetsFieldIds
   } = props;
 
+  const duplicateFacets: DisplayableFilter[] = useMemo(() => {
```
tmeyer2115, line 73: In this PR, are we only looking for Facets that duplicate selected Static Filters? I'm wondering if that's enough. Could there be other scenarios besides this that result in duplication? For instance, could we have an NLP Filter and a Static Filter that are the same?

juliannzhou: There shouldn't be any duplicate nlp filters since they are checked against static filters and facets in the prune nlp function of `appliedfilterutils`

src/components/AppliedFiltersDisplay.tsx
oshi97: Right now we're assuming that static filters have higher priority over facets in AppliedFilters when we dedupe. I'm not super happy making this assumption but looking into it

PR #253, opened 2022-07-25
set default showVerticalLinks to false

juliannzhou:
This PR changes `hideVerticalLinks` for `SearchBar` to `showVerticalLinks` and sets the default to false.

SLAP=J-2254
TEST=auto


ElemelonWind:
You'll need to fix `SearchBar`'s tests (there are a few that include `hideVerticalLinks`)

tests/components/SearchBar.test.tsx
```diff
@@ -147,34 +147,34 @@ describe('SearchBar', () => {
         .mockResolvedValue(mockedUniversalAutocompleteResult);
     });
 
-    it('displays vertical links as part of the query suggestions', async () => {
+    it('does not display vertical links on default', async () => {
```
oshi97, line 150: I think it would have been less code changes to flip the props in the test instead of the order of the tests but this also works!

oshi97 (approved):

src/components/SearchBar.tsx
```diff
@@ -142,8 +142,8 @@ export interface SearchBarProps {
   customCssClasses?: SearchBarCssClasses,
   /** {@inheritDoc VisualAutocompleteConfig} */
   visualAutocompleteConfig?: VisualAutocompleteConfig,
-  /** Hides vertical links if true. */
-  hideVerticalLinks?: boolean,
+  /** Shows vertical links if true. */
```
yen-tt, line 145: can we also mention the default value in the doc?

yen-tt:
lets update storybook for SearchBar to show vertical links as well

yen-tt (approved):

PR #256, opened 2022-07-25
FilterSearch: always have option selected

ElemelonWind:
The active text in the input should be the first, matching Filter from the `Dropdown`. If there are no Filters, the active text clears to an empty string. 

Arrowing/tabbing no longer changes the text in the input- only selecting will.

Ran into a race condition, my workaround is in the `useEffect` call in `FocusContext`. It makes it look like I'm setting everything twice, which is basically what's happening. 

J=SLAP-2256
TEST=auto

Wrote new jests:
- type something and click out => shouldn't select anything
- type and enter without navigating => selects first option

Note: investigate storybook `await`/`delay` instances later - sometimes it works and sometimes it doesn't?


ElemelonWind:
I'm not sure why the changes from my previous PR are here even though it's been merged already... ;-;

src/components/Dropdown/Dropdown.tsx
```diff
@@ -218,6 +220,9 @@ function useDropdownContextInstance(
   const toggleDropdown = (willBeOpen: boolean) => {
     if (!willBeOpen) {
       setHasTyped(false);
+      if (index === -1) {
+        updateFocusedItem(0);
```
nmanu1, line 224: these updates would change the behavior for the `SearchBar` dropdown too, right, which we don't want?

src/components/FilterGroup.tsx
```diff
@@ -117,15 +117,14 @@ function CheckboxOptions({
   const shouldRenderOption = (
     option: FilterOptionConfig
   ) => {
-    option.displayName = option.displayName || option.value.toString();
-    return searchUtilities.isCloseMatch(option.displayName, searchValue);
+    return searchUtilities.isCloseMatch(option.displayName || option.value.toString(), searchValue);
```
nmanu1, line 120: are these changes from your last PR?

src/components/Dropdown/Dropdown.tsx
ElemelonWind: hmm you're right, do you think making a new prop for `Dropdown` on whether or not it's `FilterSearch` would make sense? or should I try modifying `FilterSearch`'s `onToggle` method

yen-tt: I think it's best if we avoid adding new props, especially if its usage is to going to be for specific component implementation only. If we can update onToggle method for this behavior, that would be great!

tests/components/SearchBar.stories.tsx
```diff
@@ -82,4 +83,5 @@ function conductRecentSearches(textboxEl: HTMLElement) {
   userEvent.clear(textboxEl);
   userEvent.type(textboxEl, 'recent search 2');
   userEvent.keyboard('{enter}');
+  userEvent.clear(textboxEl);
```
yen-tt, line 86: lets revert the changes in SearchBar stories since it's not behaving as expected with the await approach. We can revisit this later

src/components/Dropdown/Dropdown.tsx
```diff
@@ -162,20 +172,33 @@ function useFocusContextInstance(
...
+
   function updateFocusedItem(updatedFocusedIndex: number, value?: string) {
     const numItems = items.length;
     let updatedValue;
     if (updatedFocusedIndex === -1 || updatedFocusedIndex >= numItems || numItems === 0) {
       updatedValue = value ?? lastTypedOrSubmittedValue;
-      setFocusedIndex(-1);
-      setFocusedItemData(undefined);
-      setScreenReaderKey(screenReaderKey + 1);
+      if (alwaysSelectOption && numItems !== 0) {
```
nmanu1, line 194: does this prevent us from arrowing up to loop backwards through the dropdown? I know that's the desired behavior for the search bar, but not sure if that's an issue for filter search or not

src/components/Dropdown/DropdownInput.tsx
```diff
@@ -37,15 +37,15 @@ export function DropdownInput(props: {
...
     updateFocusedItem(-1, e.target.value);
     setLastTypedOrSubmittedValue(e.target.value);
-    onChange?.(e.target.value);
   }, [onChange, setLastTypedOrSubmittedValue, toggleDropdown, updateFocusedItem]);
 
   const handleKeyDown = useCallback((e: KeyboardEvent<HTMLInputElement>) => {
     if (e.key === 'Enter' && (!submitCriteria || submitCriteria(focusedIndex))) {
       toggleDropdown(false);
-      updateFocusedItem(-1, value);
+      updateFocusedItem(focusedIndex);
```
yen-tt, line 48: why do we need this change?

src/components/Dropdown/Dropdown.tsx
```diff
@@ -162,20 +172,33 @@ function useFocusContextInstance(
...
+        setFocusedIndex(0);
+        setFocusedItemData(items[0].itemData);
+      } else {
```
yen-tt, line 197: do we still need the if else here for alwaysSelectOption? Is there a case where the useEffect logic above wouldn't catch?

src/components/Dropdown/Dropdown.tsx
```diff
@@ -162,20 +172,33 @@ function useFocusContextInstance(
...
+  alwaysSelectOption: boolean
 ): FocusContextType {
   const [focusedIndex, setFocusedIndex] = useState(-1);
   const [focusedValue, setFocusedValue] = useState<string | null>(null);
   const [focusedItemData, setFocusedItemData] = useState<Record<string, unknown> | undefined>(undefined);
 
+  useEffect(() => {
+    if (alwaysSelectOption && items.length > 0) {
+      setFocusedIndex(0);
+      setFocusedItemData(items[0].itemData);
```
nmanu1, line 185: if someone types something in the filter search input, we set the first option as active/focused. but it seems like that would cause the filter to be applied if the user clicks out of the dropdown without explicitly selecting or focusing on the option, but then clicks the Apply Filters button. is that the desired behavior?

src/components/Dropdown/Dropdown.tsx
```diff
@@ -162,20 +172,33 @@ function useFocusContextInstance(
...
     const numItems = items.length;
     let updatedValue;
     if (updatedFocusedIndex === -1 || updatedFocusedIndex >= numItems || numItems === 0) {
       updatedValue = value ?? lastTypedOrSubmittedValue;
-      setFocusedIndex(-1);
-      setFocusedItemData(undefined);
-      setScreenReaderKey(screenReaderKey + 1);
+      if (alwaysSelectOption && numItems !== 0) {
+        setFocusedIndex(0);
+        setFocusedItemData(items[0].itemData);
```
nmanu1, line 196: I think the screenreader key needs to be set here as well as in the `else`. otherwise, the screenreader may not make announcements if the user types but the number of dropdown options stays the same

tests/components/FilterSearch.stories.tsx
```diff
@@ -90,11 +90,12 @@ DropdownHighlight.parameters = {
     autoCompleteService: generateMockedAutocompleteService(undefined, labeledFilterSearchResponse)
   }
 };
-DropdownHighlight.play = ({ canvasElement }) => {
+DropdownHighlight.play = async ({ canvasElement }) => {
```
nmanu1, line 93: I think we can just delete this story now since the other 2 stories with expanded dropdowns will already show what a highlighted value looks like

src/components/Dropdown/DropdownInput.tsx
```diff
@@ -37,15 +37,15 @@ export function DropdownInput(props: {
...
   const handleChange = useCallback((e: ChangeEvent<HTMLInputElement>) => {
     toggleDropdown(true);
+    onChange?.(e.target.value);
     updateFocusedItem(-1, e.target.value);
     setLastTypedOrSubmittedValue(e.target.value);
-    onChange?.(e.target.value);
   }, [onChange, setLastTypedOrSubmittedValue, toggleDropdown, updateFocusedItem]);
 
   const handleKeyDown = useCallback((e: KeyboardEvent<HTMLInputElement>) => {
     if (e.key === 'Enter' && (!submitCriteria || submitCriteria(focusedIndex))) {
```
nmanu1, line 46: is this concept of `submitCriteria` needed anymore? if `alwaysSelectOption` is true, then we wouldn't need to check if it meets the criterion of having `focusedIndex >= 0` because that would always be the case, right?

src/components/Dropdown/Dropdown.tsx
nmanu1: Not sure if you're already working on this, but it looks like both of these are still a problem

src/components/Dropdown/Dropdown.tsx
nmanu1: The very last scenario I listed here still seems a little odd. If the user types something, navigates through the dropdown, but then changes the input text before clicking out, would we want a filter to be selected? since the last time they changed the input, they haven't interacted with the dropdown. so, maybe the desired behavior would be to not select the filter unless the user interacts with the dropdown after the last time they type? this might be a question for product

src/components/Dropdown/Dropdown.tsx
nmanu1: I tested this out again and the looping/tabbing works! The only thing now is that when a user types something, arrows/tabs through the dropdown and then back to the input, and then clicks outside, the first filter in the dropdown is no longer selected and applied on the next search

src/components/Dropdown/Dropdown.tsx
```diff
@@ -103,7 +118,11 @@ export function Dropdown(props: PropsWithChildren<DropdownProps>): JSX.Element {
     if (e.key === 'ArrowDown') {
       updateFocusedItem(focusedIndex + 1);
     } else if (e.key === 'ArrowUp') {
-      updateFocusedItem(focusedIndex - 1);
+      if (alwaysSelectOption && value !== focusedValue) {
```
yen-tt, line 121: if the input box's value is equal to the focused value (e.g. selected something, click out of the text box, and click back in), then arrowing up wouldn't loop to the the last option like you can normally. I'm not sure if that's the behavior you are expecting, but it may be better to have this as `alwaysSelectOption && focusedIndex === 0` so it would still loop regardless

src/components/Dropdown/Dropdown.tsx
```diff
@@ -103,7 +118,11 @@ export function Dropdown(props: PropsWithChildren<DropdownProps>): JSX.Element {
...
+        updateFocusedItem(focusedIndex - 2);
+      } else {
+        updateFocusedItem(focusedIndex - 1);
+      }
     } else if (e.key === 'Tab' && !e.shiftKey) {
```
yen-tt, line 126: if a user typed something and tab out, this wouldn't be selecting anything at the moment. We may need to update this part as well to select the first option or something

src/components/Dropdown/Dropdown.tsx
```diff
@@ -162,20 +181,49 @@ function useFocusContextInstance(
...
-      setScreenReaderKey(screenReaderKey + 1);
+      if (alwaysSelectOption && numItems !== 0 && selectedBeforeClose) {
+        setFocusedIndex(0);
+        setFocusedItemData(items[0].itemData);
+        setScreenReaderKey(screenReaderKey + 1);
+        updateFocusedValue = false;
+      } else {
+        setFocusedIndex(-1);
+        setFocusedItemData(undefined);
+        setScreenReaderKey(screenReaderKey + 1);
```
yen-tt, line 225: is there a reason why we only sometimes call `setScreenReaderKey` in various cases in updateFocusedItem? as well as not calling it in the useEffect above?

src/components/Dropdown/Dropdown.tsx
```diff
@@ -162,20 +181,49 @@ function useFocusContextInstance(
...
   function updateFocusedItem(updatedFocusedIndex: number, value?: string) {
     const numItems = items.length;
     let updatedValue;
+    let updateFocusedValue = true;
     if (updatedFocusedIndex === -1 || updatedFocusedIndex >= numItems || numItems === 0) {
       updatedValue = value ?? lastTypedOrSubmittedValue;
-      setFocusedIndex(-1);
-      setFocusedItemData(undefined);
-      setScreenReaderKey(screenReaderKey + 1);
+      if (alwaysSelectOption && numItems !== 0 && selectedBeforeClose) {
```
yen-tt, line 217: what is the purpose of selectedBeforeClose?

src/components/Dropdown/DropdownInput.tsx
```diff
@@ -37,15 +37,15 @@ export function DropdownInput(props: {
...
+      updateFocusedItem(focusedIndex);
```
yen-tt, line 47: why do we need this change?

src/components/Dropdown/DropdownInput.tsx
nmanu1, line 46: is this concept of `submitCriteria` needed anymore? if `alwaysSelectOption` is true, then we wouldn't need to check if it meets the criterion of having `focusedIndex >= 0` because that would always be the case, right?

src/components/Dropdown/Dropdown.tsx
```diff
@@ -101,9 +116,17 @@ export function Dropdown(props: PropsWithChildren<DropdownProps>): JSX.Element {
     }
 
     if (e.key === 'ArrowDown') {
-      updateFocusedItem(focusedIndex + 1);
+      if (alwaysSelectOption && focusedIndex === items.length-1) {
```
nmanu1, line 119: if there were only one item in the dropdown, would arrowing down set it as the focused item and the filter would be applied on the next search? from this, it looks like you wouldn't be able to arrow down to set the filter as selected

src/components/Dropdown/Dropdown.tsx
yen-tt: best to ask products thought on this, perhaps it would just be clearing the input text and not select anything

src/components/Dropdown/Dropdown.tsx
```diff
@@ -186,7 +238,7 @@ function useFocusContextInstance(
       setFocusedIndex(updatedFocusedIndex);
       setFocusedItemData(items[updatedFocusedIndex].itemData);
     }
-    setFocusedValue(updatedValue);
+    if (updateFocusedValue) setFocusedValue(updatedValue);
```
yen-tt, line 241: what is the case where we would not want to update here?

ElemelonWind: When toggling out of the dropdown, I check if the user has selected anything by comparing `value` and `focusedValue` - if they aren't the same that means the user's latest action was not selecting a filter. But in `updateFocusedItem`, `focusedValue` is always set to the same thing as `value` - the thing that usually is consistent with what is actually selected is `itemData`. So I added a that logic to make the checking a bit easier

src/components/Dropdown/DropdownInput.tsx
yen-tt: what happen if user hit enter in the input box when there are no options? Should it clear the search or leave it as is with nothing selected?

src/components/Dropdown/Dropdown.tsx
nmanu1: oh whoops I think I was looking at the ArrowUp case instead of ArrowDown. I think the scenario where there's only one option in the dropdown might still be a little weird though. because if the user types something, and then arrows to the dropdown option, I think they wouldn't be able to arrow back to the input where they get what they last typed instead of the dropdown item value. is that the expected behavior?

src/components/Dropdown/Dropdown.tsx
nmanu1: does that mean if the user types some text that exactly matches a filter value, even if they don't interact with the dropdown before clicking out, that filter would be applied?

src/components/Dropdown/DropdownInput.tsx
ElemelonWind: Ooh yeah it shouldn't trigger a search. I'll need to revert my commit where I removed the `submitCriteria`

src/components/Dropdown/Dropdown.tsx
ElemelonWind: So would we want arrowing up on index 0 to make the value the last inputted value? Or should it work for arrowing down on the last index too

src/components/Dropdown/DropdownInput.tsx
nmanu1: I think this wouldn't be a problem related to removing `submitCriteria`, because the text wouldn't match any filter value since there are none in the dropdown? but not sure if it should clear the search or leave it as is without selecting anything

src/components/Dropdown/Dropdown.tsx
ElemelonWind: It would. I'm not sure how to remove that because the filter auto-focuses on a filter now, and afaik there'd be no way to check if there's any difference between a typed filter and a selected one except the text.

src/components/Dropdown/Dropdown.tsx
nmanu1: I think it should be consistent for both of them, but, I think it would be good to check with product on the behavior they want here. maybe they're fine with not letting people arrow back to their typed text, although I find that a little odd

src/components/Dropdown/DropdownInput.tsx
ElemelonWind: I reverted it anyways because if there are no filters in the dropdown, the focused index is -1, which means that the check is still needed. So if the user hits enter on an invalid filter nothing would happen

src/components/Dropdown/Dropdown.tsx
nmanu1: would it be possible to just keep track with a boolean if the user has pressed some arrow key (or tab) and use that to determine whether to select a filter or not? it could reset whenever the user changes the input text

src/components/Dropdown/Dropdown.tsx
yen-tt: chatted with cindy, the cases taken care of by this logic (and `shouldSelect`) can be done through FilterSearch onToggle handler method.

src/components/Dropdown/Dropdown.tsx
yen-tt: chatted with Cindy, removing it since the new logic added in FilterSearch should take care of it

src/components/Dropdown/Dropdown.tsx
```diff
@@ -162,20 +192,41 @@ function useFocusContextInstance(
...
 ): FocusContextType {
   const [focusedIndex, setFocusedIndex] = useState(-1);
   const [focusedValue, setFocusedValue] = useState<string | null>(null);
   const [focusedItemData, setFocusedItemData] = useState<Record<string, unknown> | undefined>(undefined);
+  useEffect(() => {
+    if (alwaysSelectOption) {
+      if (items.length > 0) {
+        const index = focusedIndex === -1 || focusedIndex >= items.length ? 0 : focusedIndex;
+        setFocusedIndex(index);
+        setFocusedValue(items[index].value);
```
yen-tt, line 206: can we refactor all the setBlah into one helper function and invoke that instead?

src/components/Dropdown/Dropdown.tsx
yen-tt: chatted with Cindy, wouldn't select anything. Ack

src/components/Dropdown/Dropdown.tsx
ElemelonWind: I just looked at the [loom](https://www.loom.com/share/647127a38f0b4cba8e3ae7d0e3b82029) again- in the example, looping back around doesn't actually exist: see time 0:49. Should I ask about that?

ElemelonWind:
Also, it's now failing coveralls. Are there any tests you'd suggest writing?

src/components/Dropdown/Dropdown.tsx
nmanu1: I looked at the loom, and it seems like the text displayed in the input element shouldn't be updated when the user navigates through the options. The user-typed text always stays there. If we match that behavior, we won't have the problem I brought up about users not being able to go back to their typed text

src/components/Dropdown/Dropdown.tsx
```diff
@@ -48,13 +50,17 @@ export function Dropdown(props: PropsWithChildren<DropdownProps>): JSX.Element {
...
     parentQuery,
+    alwaysSelectOption = false
   } = props;
 
   const containerRef = useRef<HTMLDivElement>(null);
   const screenReaderUUID: string = useMemo(() => uuid(), []);
   const [screenReaderKey, setScreenReaderKey] = useState<number>(0);
   const [hasTyped, setHasTyped] = useState<boolean>(false);
-  const [childrenWithDropdownItemsTransformed, items] = getTransformedChildrenAndItemData(children);
+  const [lastActionSelect, setLastActionSelect] = useState<boolean>(false);
```
yen-tt, line 60: Maybe `hasNavigated` for the name?

yen-tt (commented):
can you add a couple jest tests of the cases you went through for this pr?
such as..
- type something and click out => shouldn't select anything
- type and enter without navigating => selects first option
- tab/shift+tab => shouldn't select anything
etc.

src/components/FilterSearch.tsx
```diff
@@ -130,7 +130,8 @@ export function FilterSearch({
     handleDropdownEvent(value, itemData, true);
   }, [handleDropdownEvent]);
 
-  const handleToggleDropdown = useCallback((isActive, _prevValue, value, _index, itemData) => {
+  const handleToggleDropdown =
+  useCallback((isActive, _prevValue, value, _index, itemData) => {
```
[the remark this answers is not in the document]
ElemelonWind: Yep, will fix one sec

src/components/Dropdown/DropdownInput.tsx
```diff
@@ -37,15 +37,15 @@ export function DropdownInput(props: {
...
     toggleDropdown(true);
+    onChange?.(e.target.value);
     updateFocusedItem(-1, e.target.value);
     setLastTypedOrSubmittedValue(e.target.value);
-    onChange?.(e.target.value);
   }, [onChange, setLastTypedOrSubmittedValue, toggleDropdown, updateFocusedItem]);
 
   const handleKeyDown = useCallback((e: KeyboardEvent<HTMLInputElement>) => {
     if (e.key === 'Enter' && (!submitCriteria || submitCriteria(focusedIndex))) {
+      updateFocusedItem(focusedIndex, typeof focusedItemData?.displayName === 'string' ? focusedItemData.displayName : undefined);
```
yen-tt, line 47: how come we are using the display name of the focusedItemData and not the value field (which `updateFocusedItem` uses by default if the second arg here is not specified)?

nmanu1 (commented):
it looks like clicking/pressing Enter on a dropdown option applies the filter. but, then when you remove the filter by clicking "Clear All" or removing the specific applied filter pill, the input text is no longer cleared like it used to be

src/components/Dropdown/DropdownInput.tsx
ElemelonWind: oop good catch

ElemelonWind: I have to specify it here because now arrowing/tabbing doesn't change the value in the input by default

oshi97:
Since this stuff is really tricky I was thinking it might be worth looking into factoring out this logic into some kind of `<Select>` or `<DropdownSelect>` element in the future! Completely out of scope here just wanted to plant the idea for the future.

src/components/Dropdown/Dropdown.tsx
```diff
@@ -101,28 +111,39 @@ export function Dropdown(props: PropsWithChildren<DropdownProps>): JSX.Element {
...
-      if (focusedIndex >= 0) {
+      if (focusedIndex > 0 || (!alwaysSelectOption && focusedIndex === 0)) {
         updateFocusedItem(focusedIndex - 1);
         e.preventDefault();
       } else {
         toggleDropdown(false);
       }
-    } else if (!hasTyped) {
-      setHasTyped(true);
+    } else {
```
oshi97, line 143: think we can change this back to an else if

src/components/FilterSearch.tsx
```diff
@@ -130,7 +132,8 @@ export function FilterSearch({
     handleDropdownEvent(value, itemData, true);
   }, [handleDropdownEvent]);
 
-  const handleToggleDropdown = useCallback((isActive, _prevValue, value, _index, itemData) => {
+  const handleToggleDropdown =
```
nmanu1, line 135: when clicking out of the dropdown without selecting a filter, should we reset the focused index? not sure if it's the desired behavior to have the dropdown remember where you were last when arrowing through

src/components/Dropdown/DropdownInput.tsx
```diff
@@ -37,15 +37,19 @@ export function DropdownInput(props: {
 
   const handleChange = useCallback((e: ChangeEvent<HTMLInputElement>) => {
     toggleDropdown(true);
+    onChange?.(e.target.value);
```
oshi97, line 40: For my edification was there a reason we moved the onChange?

src/components/Dropdown/DropdownInput.tsx
```diff
@@ -37,15 +37,19 @@ export function DropdownInput(props: {
...
     updateFocusedItem(-1, e.target.value);
     setLastTypedOrSubmittedValue(e.target.value);
-    onChange?.(e.target.value);
   }, [onChange, setLastTypedOrSubmittedValue, toggleDropdown, updateFocusedItem]);
 
   const handleKeyDown = useCallback((e: KeyboardEvent<HTMLInputElement>) => {
     if (e.key === 'Enter' && (!submitCriteria || submitCriteria(focusedIndex))) {
+      let newValue;
+      if (focusedItemData?.filter && typeof focusedItemData?.filter === 'object' && typeof focusedItemData?.filter['value'] === 'string') {
```
oshi97, line 48: Why does DropdownInput need to know the contents of `focusedItemData`?

src/components/FilterSearch.tsx
```diff
@@ -115,13 +115,15 @@ export function FilterSearch({
...
         searchActions.setFilterOption({ ...currentFilter, selected: false });
       }
-      searchActions.setFilterOption({ ...newFilter, displayName: newDisplayName, selected: true });
-      setCurrentFilter(newFilter);
-      setFilterQuery(value);
-      if (select && searchOnSelect) {
-        searchActions.setOffset(0);
-        searchActions.resetFacets();
-        executeSearch(searchActions);
+      if (select) {
```
oshi97, line 118: `select` is always true right? I might be missing something but I think `handleDropdownEvent` is only called within `handleSelectDropdown` which always has the 3rd param as true. Could we move `handleDropdownEvent` into `handleSelectDropdown`? There's no need to have an extra wrapper function

oshi97: Probably we used to have `handleDropdownEvent` called in 2+ places and we removed the others

src/components/Dropdown/DropdownInput.tsx
ElemelonWind: We wanted to call the `onChange` earlier so it could fetch the new items in the dropdown before updating the focused item

src/components/Dropdown/DropdownInput.tsx
ElemelonWind: We changed it so that arrowing/tabbing doesn't change the value in the input anymore, so pressing enter sets the value in the input to the focused value

src/components/FilterSearch.tsx
ElemelonWind: `handleDropdownEvent` is also called by `handleToggleDropdown`

tests/components/FilterSearch.stories.tsx
```diff
@@ -78,25 +78,6 @@ DropdownSectioned.play = ({ canvasElement }) => {
   userEvent.type(canvas.getByRole('textbox'), 'name');
 };
 
-export const DropdownHighlight = (args: FilterSearchProps) => {
```
oshi97, line 81: We don't want this snapshot anymore?

ElemelonWind: Yeah, since the first option is auto-highlighted the snapshot is the same as the normal dropdown story

src/components/FilterSearch.tsx
oshi97: I think we want to refactor things to remove this `select` boolean but we can procrastinate on that! It's normally not very readable to have a boolean flag to "do something"

src/components/Dropdown/DropdownInput.tsx
oshi97: DropdownInput shouldn't know what your focusedItemData is. Could add some kind of `defaultItem` or `defaultValue` prop to Dropdown? This value/item would be used for the `-1` index

oshi97: Alternatively instead of a new prop, is you have `alwaysSelectOption` true the -1 index would just point to the first element

nmanu1 (commented):
now that a dropdown item is always focused, can you test if the screenreader behavior has changed? I think it might announce the highlighted value now instead of the typed text when the user changes what's in the input, which could be confusing.

generally, I think the screenreader should only announce a filter value once the user explicitly arrows on to it. but, that would mean the first filter value wouldn't be announced, which is not great for accessibility. maybe in this case, the first value should be announced along with the text change and number of options selected?

tests/components/FilterSearch.stories.tsx
oshi97: sgtm! I was thinking I wouldn't want to lose coverage on the {Tab} interactions but we have unit tests for it

src/components/Dropdown/DropdownInput.tsx
ElemelonWind: What exactly is default thing be for? To clarify, this is only setting the value in the input when the user presses enter; arrowing down would keep the value the same which is why I had to change the `setValue` in `updateFocusedItem`

oshi97: I was just trying to brainstorm solutions. Regardless how you do it, it's important DropdownInput doesn't access `focusedItemData`.

ElemelonWind: gotcha. Would it be ok to add a new param to toggleDropdown and tell it to set the value?

oshi97: Are you saying you would pass a `setValue` callback down to `handleToggleDropdown`? That's not ideal but probably ok for now. If you wanted to do it that way couldn't you call `setFilterQuery` instead of adding a new param?

ElemelonWind: No I meant the `toggleDropdown` method in `Dropdown`. If I had a param that was like `selected` I could probably call `setValue` there- and if the user clicked out instead of clicking a filter/entering then `selected` would be `false` and wouldn't change the input value

oshi97: Sure I'm not exactly sure what you're describing but give it a go. If you want I can get on zoom

nmanu1 (commented):
I'm still seeing some weird screenreader behavior:

https://user-images.githubusercontent.com/88398086/182682968-3559cfd0-2dec-4928-8489-ea298b0794e5.mov

Only the first typed character has the correct announcement. After that, it announces the first value in the previous dropdown when you type, instead of the newly typed character. Additionally, the value after the options count isn't always correct.

oshi97:
I noticed that the FilterSearch clears out its input if I already have a value selected, then click in and click away.
It might be good to practice test driven development here, i.e. each time you make a specific case work writing a test for it! That way you'll know that fixing one specific case won't break a separate one.

https://user-images.githubusercontent.com/23005393/182689851-657ec45c-7ece-4491-9376-a673aa1d76f0.mov

src/components/Dropdown/Dropdown.tsx
```diff
@@ -203,9 +203,7 @@ function useFocusContextInstance(
         setFocusedIndex(index);
         setFocusedValue(items[index].value);
         setFocusedItemData(items[index].itemData);
-        if (focusedIndex === -1) {
-          setAlwaysSelectScreenReaderText(screenReaderText + ' ' + items[index].value);
-        }
+        setAlwaysSelectScreenReaderText(screenReaderText + ' ' + items[index].value);
```
nmanu1, line 205: I think this would read the options count screen reader text even when arrowing onto dropdown options instead of just when the input text changes. when navigating with keyboard, we would only want the dropdown item's value to be announced

oshi97:
I also noticed a weird flicker that happens when I click into the filtersearch with a filter already selected

https://user-images.githubusercontent.com/23005393/182690473-428fcbb6-bb38-413e-89a9-6d9e361e59c3.mov

ElemelonWind:
> I also noticed a weird flicker that happens when I click into the filtersearch with a filter already selected. For instance if I have "Randall Turner" selected and click into it, it will first display 3 options, then 1. The initial 3 options are probably from the previous filtersearch request, and the 1 option is from the request for "Randall Turner"
> 
>  Screen.Recording.2022-08-03.at.3.12.44.PM.mov

I think that might just be the promise that's being called taking time to fetch data. I looked at main branch and it also does this

oshi97:
> > I also noticed a weird flicker that happens when I click into the filtersearch with a filter already selected. For instance if I have "Randall Turner" selected and click into it, it will first display 3 options, then 1. The initial 3 options are probably from the previous filtersearch request, and the 1 option is from the request for "Randall Turner"
> > Screen.Recording.2022-08-03.at.3.12.44.PM.mov
> 
> I think that might just be the promise that's being called taking time to fetch data. I looked at main branch and it also does this

Could you see if this can be fixed by calling `executeFilterSearch` inside `handleDropdownEvent`, in the same block where we call `setFilterQuery`? I think it makes sense to hit the filtersearch endpoint when filterQuery is updated and should fix this

ElemelonWind:
> > > I also noticed a weird flicker that happens when I click into the filtersearch with a filter already selected. For instance if I have "Randall Turner" selected and click into it, it will first display 3 options, then 1. The initial 3 options are probably from the previous filtersearch request, and the 1 option is from the request for "Randall Turner"
> > > Screen.Recording.2022-08-03.at.3.12.44.PM.mov
> > 
> > 
> > I think that might just be the promise that's being called taking time to fetch data. I looked at main branch and it also does this
> 
> Could you see if this can be fixed by calling `executeFilterSearch` inside `handleDropdownEvent`, in the same block where we call `setFilterQuery`? I think it makes sense to hit the filtersearch endpoint when filterQuery is updated and should fix this

this seemed to work!

src/components/Dropdown/Dropdown.tsx
```diff
@@ -162,20 +183,42 @@ function useFocusContextInstance(
...
+        setFocusedIndex(index);
+        setFocusedValue(items[index].value);
+        setFocusedItemData(items[index].itemData);
+      } else {
+        setFocusedIndex(-1);
+        setFocusedValue(null);
+        setFocusedItemData(undefined);
+      }
+    }
+  // eslint-disable-next-line react-hooks/exhaustive-deps
```
oshi97, line 205: Could we remove this eslint-disable?

oshi97: We don't want to "lie" to react about deps (there's a good number of articles on this that just parrot Dan Abramov's twitter)

ElemelonWind: ooh yup i had it there when i was experimenting w some stuff lol

src/components/Dropdown/Dropdown.tsx
```diff
@@ -25,10 +25,11 @@ export interface DropdownProps {
     prevValue: string,
     value: string,
     index: number,
-    focusedItemData: Record<string, unknown> | undefined
+    focusedItemData: Record<string, unknown> | undefined,
```
nmanu1, line 28: nit: extra comma

src/components/Dropdown/Dropdown.tsx
```diff
@@ -64,7 +68,8 @@ export function Dropdown(props: PropsWithChildren<DropdownProps>): JSX.Element {
     lastTypedOrSubmittedValue,
     setValue,
     screenReaderKey,
-    setScreenReaderKey
+    setScreenReaderKey,
+    alwaysSelectOption,
```
nmanu1, line 72: nit: extra comma

src/components/Dropdown/Dropdown.tsx
```diff
@@ -85,37 +90,53 @@ export function Dropdown(props: PropsWithChildren<DropdownProps>): JSX.Element {
...
+  ]);
 
   useRootClose(containerRef, () => {
     toggleDropdown(false);
   }, { disabled: !isActive });
 
   useGlobalListener('keydown', e => {
-    if (e.key === 'ArrowDown' || e.key === 'ArrowUp') {
-      e.preventDefault();
+    if (e.key === 'Tab' || e.key === 'ArrowDown' || e.key === 'ArrowUp') {
```
nmanu1, line 105: maybe I'm missing something, but what does this `if` statement add?

src/components/Dropdown/Dropdown.tsx
```diff
@@ -85,37 +90,53 @@ export function Dropdown(props: PropsWithChildren<DropdownProps>): JSX.Element {
...
+      }
     }
 
     if (!isActive) {
       return;
     }
 
     if (e.key === 'ArrowDown') {
-      updateFocusedItem(focusedIndex + 1);
+      if (alwaysSelectOption && focusedIndex === items.length-1) {
```
nmanu1, line 116: nit: spaces around `-`

src/components/Dropdown/Dropdown.tsx
```diff
@@ -85,37 +90,53 @@ export function Dropdown(props: PropsWithChildren<DropdownProps>): JSX.Element {
...
+        updateFocusedItem(0);
+      } else {
+        updateFocusedItem(focusedIndex + 1);
+      }
     } else if (e.key === 'ArrowUp') {
-      updateFocusedItem(focusedIndex - 1);
+      if (alwaysSelectOption && focusedIndex === 0) {
+        updateFocusedItem(items.length-1);
```
nmanu1, line 123: nit: spaces around `-`

src/components/Dropdown/Dropdown.tsx
```diff
@@ -85,37 +90,53 @@ export function Dropdown(props: PropsWithChildren<DropdownProps>): JSX.Element {
...
+      } else {
+        updateFocusedItem(focusedIndex - 1);
+      }
     } else if (e.key === 'Tab' && !e.shiftKey) {
       if (items.length !== 0) {
         if (focusedIndex >= items.length - 1) {
           toggleDropdown(false);
           updateFocusedItem(-1);
+          toggleDropdown(false);
```
nmanu1, line 130: looks like this is a duplicate call to `toggleDropdown`

src/components/Dropdown/Dropdown.tsx
ElemelonWind: oh whoops that was also from when i was experimenting

src/components/Dropdown/Dropdown.tsx
```diff
@@ -48,13 +49,16 @@ export function Dropdown(props: PropsWithChildren<DropdownProps>): JSX.Element {
     className,
     activeClassName,
     parentQuery,
+    alwaysSelectOption = false
```
nmanu1, line 52: do we need to explicitly set the default as false? I think it would have the same behavior whether `alwaysSelectOption` is false or undefined

src/components/Dropdown/Dropdown.tsx
```diff
@@ -162,20 +183,41 @@ function useFocusContextInstance(
   lastTypedOrSubmittedValue: string,
   setValue: (newValue: string) => void,
   screenReaderKey: number,
-  setScreenReaderKey: (newKey: number) => void
+  setScreenReaderKey: (newKey: number) => void,
+  alwaysSelectOption: boolean,
```
nmanu1, line 185: nit: extra comma

src/components/Dropdown/Dropdown.tsx
```diff
@@ -204,13 +246,13 @@ function useDropdownContextInstance(
...
   screenReaderUUID: string,
-  setHasTyped: (boolean) => void,
+  setHasTyped: (hasTyped: boolean) => void,
   onToggle?: (
     isActive: boolean,
     prevValue: string,
     value: string,
     index: number,
-    focusedItemData: Record<string, unknown> | undefined
+    focusedItemData: Record<string, unknown> | undefined,
```
nmanu1, line 253: nit: extra comma

src/components/Dropdown/DropdownInput.tsx
```diff
@@ -80,7 +86,7 @@ export function DropdownInput(props: {
       id={screenReaderUUID && generateDropdownId(screenReaderUUID, -1)}
       autoComplete='off'
       aria-describedby={screenReaderUUID}
-      aria-activedescendant={screenReaderUUID && generateDropdownId(screenReaderUUID, focusedIndex)}
+      aria-activedescendant={!isTyping ? (screenReaderUUID && generateDropdownId(screenReaderUUID, focusedIndex)) : ''}
```
nmanu1, line 89: can this just be `!isTyping && screenReaderUUID && generateDropdownId(screenReaderUUID, focusedIndex)`?

src/components/Dropdown/Dropdown.tsx
```diff
@@ -101,22 +111,31 @@ export function Dropdown(props: PropsWithChildren<DropdownProps>): JSX.Element {
     }
 
     if (e.key === 'ArrowDown') {
-      updateFocusedItem(focusedIndex + 1);
+      if (alwaysSelectOption && focusedIndex === items.length-1) {
+        updateFocusedItem(0);
+      } else {
+        updateFocusedItem(focusedIndex+1);
```
nmanu1, line 117: nit: spacing here and a few more places in this file

src/components/Dropdown/DropdownInput.tsx
ElemelonWind: when i try this I get an error: 

Types of property ''aria-activedescendant'' are incompatible.
      Type 'string | false | undefined' is not assignable to type 'string | undefined'.
        Type 'boolean' is not assignable to type 'string'.

nmanu1: oh can you try switching the order of `screenReaderUUID` and `!isTyping`?

src/components/Dropdown/Dropdown.tsx
oshi97: I made an item to add https://typescript-eslint.io/rules/space-infix-ops to our eslint

ElemelonWind: the spacing rule is just no spaces between operators right? or is there other spacing i also need to fix

src/components/Dropdown/DropdownInput.tsx
ElemelonWind: this: `screenReaderUUID && !isTyping && generateDropdownId(screenReaderUUID, focusedIndex)`? it still doesn't work :c

src/components/Dropdown/Dropdown.tsx
oshi97: we want the spacing between operators so like `focusedIndex + 1` here instead of `focusedIndex+1`
I'm definitely ok ignoring adjusting these for now since this is an autofixable eslint rule.

src/components/Dropdown/DropdownInput.tsx
nmanu1: ok nvm then, the ternary operator makes it look a little uglier, but it's fine

oshi97: the && operator returns the first falsy value, so if screenReaderUUID is falsy it will return blank string since screenReaderUUID is a string, but if screenReaderUUID is truthy but !isTyping is falsy it will return false since isTyping is a boolean.

Maybe we can tweak `generateDropdownId` to accept an undefined screenReaderUUID, and return blank string if it's undefined? Normally it's good to keep utility methods "dumb" but this is an internal only method, and we do this `screenReaderUUID && generateDropdownId` check literally every time we call `generateDropdownId` so it's fine to bake in.
then for aria-activedescendant we can do
```ts
!isTyping ? generateDropdownId(screenReaderUUID, focusedIndex) : ''
// or
!isTyping && generateDropdownId(screenReaderUUID, focusedIndex) || ''
```

src/components/Dropdown/generateDropdownId.ts
```diff
@@ -5,6 +5,7 @@
  * @param screenReaderUUID - the shared UUID for the {@link Dropdown} instance
  * @param index - the index of the dropdown item in question
  */
-export function generateDropdownId(screenReaderUUID: string, index: number): string {
+export function generateDropdownId(screenReaderUUID: string | undefined, index: number): string {
+  if (!screenReaderUUID) return '';
```
oshi97, line 9: can we change the other `screenReaderUUID && generateDropdownId()` usages to just `generateDropdownId()`?

src/components/Dropdown/DropdownInput.tsx
```diff
@@ -80,7 +86,7 @@ export function DropdownInput(props: {
       id={screenReaderUUID && generateDropdownId(screenReaderUUID, -1)}
       autoComplete='off'
       aria-describedby={screenReaderUUID}
-      aria-activedescendant={screenReaderUUID && generateDropdownId(screenReaderUUID, focusedIndex)}
+      aria-activedescendant={!isTyping ? (generateDropdownId(screenReaderUUID, focusedIndex)) : ''}
```
oshi97, line 89: small nit but can we flip the boolean to just `isTyping ? generateDropdownId : ''`? a negated boolean is very slightly harder to read

juliannzhou:
I noticed a bug while testing that when you click an option after typing something other than the option selected it will select two filters. See below for the behavior:

https://user-images.githubusercontent.com/65748924/182893608-65f44d65-25ed-48f4-b953-6ead72dfb56a.mov

ElemelonWind:
> I noticed a bug while testing that when you click an option after typing something other than the option selected it will select two filters. See below for the behavior:
> 
>  two-options.mov

i've tried it and i'm not able to replicate this?

juliannzhou:
> > I noticed a bug while testing that when you click an option after typing something other than the option selected it will select two filters. See below for the behavior:
> > two-options.mov
> 
> i've tried it and i'm not able to replicate this? Is this just on refresh or do you do something beforehand

This seems to be fixed in the recent commit. When I pulled the issue disappeared, but it was an issue when I do `git reset --hard dev/filtersearch-option@{"30 minutes ago"}`

oshi97 (approved):
🔥 🔥 🔥

juliannzhou:
There was also another behavior I noticed which is when you enter an option and apply the filter, then delete the input partially and then clear filter, the input does not clear, which is different from the behavior in main. I am not sure if this is something we want to address.


https://user-images.githubusercontent.com/65748924/182909371-6e317046-ae61-4bdf-80ba-fb250d71c3be.mov

https://user-images.githubusercontent.com/65748924/182909379-8eeb5840-9ade-4dfc-ba27-d90d53707009.mov

juliannzhou:
You might also want to check the interaction with apply button. If you select an option and delete partial or all input, and then click the apply button, it will get rid of the filter but the results still remain on the page.

https://user-images.githubusercontent.com/65748924/182932199-fa034f80-87c7-479f-abe5-5a55df1ef5e6.mov

https://user-images.githubusercontent.com/65748924/182932210-b25c8f3d-5daf-4afd-af1e-f9acaf21187a.mov




.

PR #259, opened 2022-07-26
use useComposedCssClasses in ApplyFiltersButton

yen-tt:
Previously, custom styling would replace default styling. use `useComposedCssClasses` to properly merge default tailwind styling with user custom styling.

J=SLAP-2284
TEST=manual

update ApplyFiltersButton in PeoplePage to include `customCssClasses={{ button: "mt-2 text-red-600" }}` and see that the the custom styling is merged with the default instead of replacing it.


nmanu1 (commented):
are there any other places where this was a problem?

yen-tt:
> are there any other places where this was a problem?

Did a quick check across our components, everything else looks alright. I believe this is the only one that is not using useComposedCssClasses

nmanu1 (approved):

PR #260, opened 2022-07-27
Dedupe filters in AppliedFilters

oshi97:
This PR adds deduping logic to the `AppliedFilters` component.

In order to do this, a refactor of `AppliedFiltersDisplay` was done so that it no longer differentiates
between facets, hierarchical facets, static filters, and nlp filters.
Now, it only knows whether a filter is a `RemovableFilter`, or an nlp filter display name (a string).
Each `RemovableFilter` has a callback, `handleRemove`, which is called when it is removed.
Duplicate filters are deduped, and when the deduped version is removed, it's `handleRemove`
will be called, and then all of its duplicate siblings' will also be called.

I tried to separate logic out into hooks that are single hook per file, and deleted displayfilterutils and appliedfilterutils

J=SLAP-2248
TEST=manual

test that if I have a duplicate hierarchical facet and static filter, only a single RemovableFilter will be rendered, and removing it by clicking on it removes both the hierarchical facet and static filter from their respective components

test that if I have a duplicate regular facet and a static filter (I used c_employeeDepartment="Consulting") it also dedupes and can remove both simultaneously


src/components/AppliedFilters.tsx
```diff
@@ -58,42 +58,29 @@ export interface AppliedFiltersProps {
...
-    );
-  }, [
-    hasResults,
-    filters,
-    hiddenFields,
-    hierarchicalFacetsDelimiter,
-    hierarchicalFacetsFieldIds,
-    nlpFilters
-  ]);
+  const removableFilters = useRemovableFilters(
```
tmeyer2115, line 70: The verb `use` seems a bit funny here and below. We're not using anything really, we're getting the values. Do these need to be React Hooks or can they just be normal functions? The benefit of writing your own Hook is that you can share stateful logic between Components. Are we doing that with these functions?

src/components/AppliedFilters.tsx
```diff
@@ -58,42 +58,29 @@ export interface AppliedFiltersProps {
...
+    hierarchicalFacetsFieldIds, hierarchicalFacetsDelimiter, hiddenFields);
+  const nlpFilterDisplayNames = useNlpFilterDisplayNames(removableFilters.map(f => f.filter), hiddenFields);
```
tmeyer2115, line 72: It feels a bit like we're making these functions Hooks out of necessity. A Hook should allow us to reuse stateful logic between Components. We don't invoke these methods anywhere else, right? They use Hooks internally, so they have to be Hooks at the top-level. 

Instead, what do you think about making these just normal functions? Any memo-izing would happen in this file. Any `answersActions` or `answersState` dependencies would be injected.

oshi97: I like how making these hooks lets us call hooks from within them. This results in more concise code. I think whether or not we invoke these methods in other components doesn't effect whether or not these functions are hooks or not - by that same logic regular utility functions are also meant to be shared between components. I would actually argue that because these functions aren't shared between components that it's a little simpler to just make these hooks and not worry about passing parameters down to these functions. Having to pass tons of parameters down was something I really disliked about appliedfilterutils, especially because those parameters were only going to be used within those utility functions and not anywhere else in the component.

oshi97: I'm definitely open to turning these into regular functions if we need to share them with other components, though! I think doing that refactor isn't something that will get harder with time, so there's no need to jump start it before it's needed.

tmeyer2115 (approved):

PR #261, opened 2022-07-27
Add StaticFilters stories

nmanu1:
Add 2 stories for `StaticFilters`: one with the default props with an option selected and one with `searchable = true`. Also, fix errors in `UniversalResults` and `VerticalResults` stories now that the props require a type param.

J=SLAP-2286
TEST=manual

Check that the new Percy snapshots look as expected.

### DIFF
--- a/tests/components/UniversalResults.stories.tsx
+++ b/tests/components/UniversalResults.stories.tsx
@@ -6,6 +6,7 @@ import { generateMockedHeadless } from '../__fixtures__/search-headless';
 import { UniversalResults, UniversalResultsProps } from '../../src/components/UniversalResults';
 import { RecursivePartial } from '../__utils__/mocks';
 import { verticalResults } from '../__fixtures__/data/universalresults';
+import { DefaultRawDataType } from '../../src/models/DefaultRawDataType';
 
 const meta: ComponentMeta<typeof UniversalResults> = {
   title: 'UniversalResults',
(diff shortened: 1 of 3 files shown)


tests/components/UniversalResults.stories.tsx
```diff
@@ -26,15 +27,15 @@ const verticalConfigMap = {
   }
 };
 
-export const Primary = (args: UniversalResultsProps) => {
+export const Primary = (args: UniversalResultsProps<DefaultRawDataType>) => {
```
oshi97, line 30: Do we need to specify the default? I thought it would default to the default

alextaing:
Do you think it would be good to add another story for a collapsed static filter?

tests/components/UniversalResults.stories.tsx
nmanu1: Only StandardSectionProps, StandardCardProps, SectionComponent, and CardComponent have defaults because those are the ones we would expect users may want to write/extend. Vertical/UniversalResults don't because they would generally be able to infer the type from the props. Here we have to specify the type param because we're using the prop interface directly as a type

nmanu1:
> Do you think it would be good to add another story for a collapsed static filter?

I didn't include a story for that because we already test the collapsed behavior in the Jest tests and there isn't much more to the UI beyond that. It didn't seem particularly useful from a documentation perspective either since users can go into the Storybook site and play around with the collapsing behavior easily. But, if you think it would still be helpful, I can add it. I don't have a strong preference

yen-tt (approved):

alextaing:
> > Do you think it would be good to add another story for a collapsed static filter?
> 
> I didn't include a story for that because we already test the collapsed behavior (e.g. `defaultExpanded = false`) in the Jest tests and there isn't much more to the UI beyond that. It didn't seem particularly useful from a documentation perspective either since users can go into the Storybook site and play around with the collapsing behavior easily. But, if you think it would still be helpful, I can add it. I don't have a strong preference

I see-- that makes sense.  Then, LGTM!

PR #262, opened 2022-07-27
Update version to v1.0.0-beta.

tmeyer2115:


package.json
```diff
@@ -1,6 +1,6 @@
 {
```
nmanu1, line 1: can we point to the new Headless-React version?

nmanu1 (commented):
does the PR description need release notes?

package.json
tmeyer2115: Updated!

tmeyer2115:
> does the PR description need release notes?

I'm going to add a brief note when I make the official GH Release. It's basically going to just point to the docs. I don't think the PR/Commit needs a message.

package.json
```diff
@@ -1,6 +1,6 @@
...
   "name": "@yext/search-ui-react",
-  "version": "0.2.0-beta.243",
+  "version": "1.0.0-beta",
```
oshi97, line 3: I think I still prefer it if we have the PR number here.

tmeyer2115: Product asked for this to be the version. I can see if they're okay with us adding the PR number.

oshi97 (approved):

PR #263, opened 2022-07-28
unique keys for applied filters

oshi97:
I assumed that displayNames would be unique for NLP filters coming from
the API. This is not necessarily the case, and results in duplicate keys
which can result in UI bugs.

Including an underscore and index at the end makes the keys effectively unique barring
very weird edge cases.

J=none
TEST=manual

searching for Richmond on a locations page (does not exist yet, will be introduced in Alex's maps pr)
then searching for texas, does not result in an incorrect UI state

### DIFF
--- a/src/components/AppliedFiltersDisplay.tsx
+++ b/src/components/AppliedFiltersDisplay.tsx
@@ -51,7 +51,7 @@ export function AppliedFiltersDisplay(props: AppliedFiltersDisplayProps): JSX.El
     return null;
   }
 
-  const dedupedNlpFilterDisplaynames = nlpFilterDisplayNames.filter(displayName => {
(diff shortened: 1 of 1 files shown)


src/components/AppliedFiltersDisplay.tsx
```diff
@@ -105,10 +105,11 @@ function getDedupedRemovableFilters(filters: RemovableFilter[]) {
 function renderRemovableFilter(
   displayName: string | undefined,
   handleRemove: () => void,
-  cssClasses: AppliedFiltersCssClasses
+  index: number,
+  cssClasses: AppliedFiltersCssClasses,
```
nmanu1, line 109: nit: extra comma at the end

nmanu1 (dismissed):
besides the nit, lgtm!

PR #264, opened 2022-07-28
add visual regression coverage

juliannzhou:
This PR adds visual regression coverage tests by using add-on storyshot from storybook. Storyshot renders a test UI with jest upon testing and compares the previously generated jest snapshot from stories with the render.

SLAP=J-2269
TEST=manual

Tests were able to run after executing the following command to show visual regression
`npm run coverage:visual` for visual-regression coverage

Snyk issues: approved by legal

Screen Shot 2022-07-28 at 4 52 12 PM


package.json
```diff
@@ -121,7 +125,8 @@
       "/node_modules/(?!(@yext/search-headless-react)/)"
     ],
     "moduleNameMapper": {
-      "./SearchCore": "<rootDir>/tests/__fixtures__/core/SearchCore.ts"
+      "./SearchCore": "<rootDir>/tests/__fixtures__/core/SearchCore.ts",
+      "\\.(css|less|scss|sass)$": "identity-obj-proxy"
```
nmanu1, line 129: for my own understanding, why is this needed?

tests/components/storyshot.test.ts
```diff
@@ -0,0 +1,15 @@
+import initStoryshots from '@storybook/addon-storyshots';
```
nmanu1, line 1: it seems a little odd for the `storyshot.test.ts` file to live in the components directory because it doesn't match the structure of our `src` folder and test a single component. maybe it could be moved a level up?

tests/components/__snapshots__/storyshot.test.ts.snap
```diff
@@ -0,0 +1,8687 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
nmanu1, line 1: do we use this snapshot file somewhere? if not and it's just auto-generated, maybe we could remove it

package.json
juliannzhou: This is added because storyshot needs to read .storybook folder for configuration and could not read the index.css file

alextaing: I ran into this issue on my own PR-- but I had to come to a similar solution.  When I imported a CSS file into one of my components, Jest tests started to fail because it seemed like Jest did not know what to make out of the CSS class imports, so you have to configure Jest to know what to do with them.

Some more information!
https://jestjs.io/docs/webpack 
https://github.com/facebook/jest/issues/3094

alextaing: But, I'm not clear on why it's needed here in this situation, so Juliann could probably answer more on that

alextaing:
Are we planning on having more than one storyshot test, or is it one storyshot test that tests all components?  If it's the latter, would it be a good idea to shorten the command to maybe `npm run storyshot`?

package.json
```diff
@@ -40,7 +40,8 @@
     "prepublishOnly": "npm run build",
     "test": "jest --coverage",
     "storybook": "start-storybook -p 6006",
-    "build-storybook": "build-storybook"
+    "build-storybook": "build-storybook",
+    "storyshot": "jest --coverage --coverageDirectory=coverage/visual-regression"
```
alextaing, line 44: Are we planning on having more than one storyshot test, or is it one storyshot test that tests all components? If it's the latter, would it be a good idea to shorten the command to maybe npm run storyshot?

tests/components/__snapshots__/storyshot.test.ts.snap
juliannzhou: Yes, the storyshot add-on renders generates a jest snapshot. It should be committed so it will compare the future renders in pr with the snapshot as part of the testing. The snapshot file needs to be updated by running the storyshot command and committed every time there's an approved change in code implementation, so we might need a github action to do that. More information here: https://jestjs.io/docs/snapshot-testing

juliannzhou:
> Are we planning on having more than one storyshot test, or is it one storyshot test that tests all components? If it's the latter, would it be a good idea to shorten the command to maybe `npm run storyshot`?

It's the one storyshot test that tests all components. The `npm run storyshot storyshot` command is to show the storyshot coverage instead of the regular jest test coverage.

tmeyer2115:
When we're collecting the coverage, what files under `src` are we checking? Are we generating coverage numbers for all of the files in `src`? If so, we shouldn't. We're only explicitly testing the contents of the `src/components` directory. We don't (and won't) have Storybook tests for the contents of `src/models` for instance.

package.json
```diff
@@ -40,7 +40,8 @@
     "prepublishOnly": "npm run build",
     "test": "jest --coverage",
     "storybook": "start-storybook -p 6006",
-    "build-storybook": "build-storybook"
+    "build-storybook": "build-storybook",
+    "storyshot": "jest --coverage --coverageDirectory=coverage/visual-regression --collectCoverageFrom='[\"src/components/**\"]' --testMatch=\"<rootDir>/tests/storyshot.test.ts\""
```
yen-tt, line 44: can we rename this to be `test:visual-coverage` or something similar?

tests/hooks/useSynchronizedRequest.test.tsx
```diff
@@ -1,5 +1,5 @@
 import { useSynchronizedRequest } from '../../src/hooks/useSynchronizedRequest';
-import { renderHook } from '@testing-library/react-hooks';
+import { renderHook } from '@testing-library/react-hooks/dom';
```
yen-tt, line 2: how come we need `/dom`?

package.json
tmeyer2115, line 44: I think it's a bit confusing to have a whole `jest` configuration in this file, and then override many of its values with command flags here. I understand why we're doing that: we need to run Jest with two separate configurations. I think we should do the following:

- Look for a best practice for this type of situation. 
- If we can't find a consistent, best practice, we should strip the `jest` configuration down to only what's common between the two runs.

tests/hooks/useSynchronizedRequest.test.tsx
juliannzhou: `/dom` is needed to resolve the warning:
`Warning: It looks like you're using the wrong act() around your test interactions.
Be sure to use the matching version of act() corresponding to your renderer:

// for react-dom:
import {act} from 'react-dom/test-utils';
// ...
act(() => ...);

// for react-test-renderer:
import TestRenderer from react-test-renderer';`

package.json
juliannzhou: I found a `--projects ` config option which allow specifying different jest configs in separate files and calling them in the command. This seems to be the best practice.

jest.unit.config.js
```diff
@@ -0,0 +1,39 @@
+module.exports = {
+  bail: 0,
```
oshi97, line 2: is it possible for the jest config files to go in the tests folder?

jest.unit.config.js
```diff
@@ -0,0 +1,39 @@
+module.exports = {
```
oshi97, line 1: Is it possible for these jest config files to be typescript? Also is there some way to pull out the shared config between the two of them? Even just a base config file that gets imported

tests/hooks/useRecentSearches.test.tsx
```diff
@@ -1,6 +1,5 @@
 import { useRecentSearches } from '../../src/hooks/useRecentSearches';
```
oshi97, line 1: (putting the comment here so we can have a comment chain, but asking about storyshot.test.ts.snap) What is storyshot.test.ts.snap?

tests/storyshot.test.ts
```diff
@@ -0,0 +1,15 @@
+import initStoryshots from '@storybook/addon-storyshots';
+import { render } from '@testing-library/react';
+import { resolve } from 'path';
+
+jest.mock('uuid', () => {
```
oshi97, line 5: why do we need this mock?

tests/hooks/useRecentSearches.test.tsx
juliannzhou: storyshot.test.ts.snap is a jest snapshot generated with stories using the storyshot add-on and would be used to compare against future renders of test UI in pr. (https://jestjs.io/docs/snapshot-testing) It needs to be updated by running `npm run coverage:visual` and committed if there is an approved change for the tests to pass.

tests/storyshot.test.ts
juliannzhou: We need this mock because uuid generated each time with the test rendering is different, so it is recognized by storyshot add-on as a code implementation change.

package.json
```diff
@@ -40,9 +40,13 @@
     "prepublishOnly": "npm run build",
     "storybook": "start-storybook -p 6006",
     "build-storybook": "build-storybook",
-    "test": "jest --projects jest.unit.config.js --coverage",
-    "coverage:visual": "jest --projects jest.visual.config.js --coverage"
-  },
+    "test": "jest --coverage --coverageDirectory=coverage/unit",
+    "coverage:visual": "jest --coverage --coverageDirectory=coverage/visual-regression --collectCoverageFrom='[\"src/components/**\"]' --testMatch=\"<rootDir>/tests/storyshot.test.ts\"",
+    "coverage:clean": "rm -rf .nyc_output && rm -rf coverage",
+    "coverage:merge": "istanbul-merge --out coverage/merged/coverage-final.json ./coverage/unit/coverage-final.json  ./coverage/visual-regression/coverage-final.json",
```
tmeyer2115, line 46: Is the coverage merging supposed to be part of this PR? If I recall, there was a separate PR for that.

jest.unit.config.js
juliannzhou: The current test files do not support typescript config. I created a branch `dev/ts-jest-config` with the config files in typescript, but many syntax errors within tests need to be resolved to accommodate typescript configuration for future reference.

package.json
juliannzhou: Yes, because the combined-coverage PR needs to use coverage:merge scripts from main in github action coverage:base which is testing the main branch. So the scripts need to be merged in for the action to work.

.storybook/main.js
```diff
@@ -1,12 +1,13 @@
 module.exports = {
   stories: [
-    '../tests/**/*.stories.tsx'
+    '../tests/**/*.stories.ts?(x)'
```
tmeyer2115, line 3: Do we need this change? I think our stories should always be TSX files, since they're using React Components.

package.json
```diff
@@ -38,10 +38,15 @@
...
     "storybook": "start-storybook -p 6006",
-    "build-storybook": "build-storybook"
-  },
+    "build-storybook": "build-storybook",
+    "test": "jest --projects tests/jest.unit.config.js --coverage",
+    "coverage:visual": "jest --projects tests/jest.visual.config.js --coverage",
+    "coverage:clean": "rm -rf .nyc_output && rm -rf coverage",
+    "coverage:merge": "istanbul-merge --out coverage/merged/coverage-final.json ./coverage/unit/coverage-final.json  ./coverage/visual-regression/coverage-final.json",
+    "coverage:merge-report": "nyc report --reporter=lcov --reporter=text --temp-dir=./coverage/merged --report-dir=./coverage/merged",
+    "coverage": "npm run coverage:clean && npm run test && npm run coverage:visual && npm run coverage:merge && npm run coverage:merge-report"
```
tmeyer2115, line 48: Having all these new `coverage` scripts is a bit confusing. The names `merge` and `merge-report` are pretty similar, I think we should do more to distinguish the two. Is `clean` scoped to just the visual regression coverage? If so, its name should indicate that.

tests/hooks/useRecentSearches.test.tsx
tmeyer2115: Is it possible to use the Percy snapshots instead of a local one? If not, no worries. But, I think we need to include a GH Action that will make a PR updating the Snapshot when appropriate. Without that action, people will inevitably forget to generate a new Snapshot and check it in.

tests/hooks/useSynchronizedRequest.test.tsx
tmeyer2115: I checked with a couple other Slapshot folks, we weren't able to see this warning. It may be related to your NPM/Node version. I'd rather we avoid using the `dom` version of `renderHook` if possible. Depending on the `dom` version can lead to other, nuanced issues.

package.json
juliannzhou: merge command is to combine the visual and jest coverage lcov files and the merge-report command is to display them. I was wondering if I should combine them or rename them to something else? The clean command removes the previous coverage folder which includes the visual, jest, and merge coverage files.

juliannzhou:
> Is it possible to use the Percy snapshots instead of a local one? If not, no worries. But, I think we need to include a GH Action that will make a PR updating the Snapshot when appropriate. Without that action, people will inevitably forget to generate a new Snapshot and check it in.

I couldn't find anything on coverage report with Percy snapshot, but there is another option [codecov](https://about.codecov.io/) which works in github but not locally and supports visual, jest, and merge coverage reports. We would need to sign up for an account with them to use the service.

package.json
oshi97: can we have a bash script instead? I don't think anybody is ever going to call something like `coverage:clean` on its own

yen-tt: I agree! We can probably move some stuff to bash script, and in package.json, have three general commands: `test:unit`, `test:visual`, `test:all` or something similar

package.json
```diff
@@ -38,10 +38,15 @@
...
-    "test": "jest --coverage",
     "storybook": "start-storybook -p 6006",
-    "build-storybook": "build-storybook"
-  },
+    "build-storybook": "build-storybook",
+    "test": "jest --projects tests/jest.unit.config.js --coverage",
+    "coverage:visual": "jest --projects tests/jest.visual.config.js --coverage",
+    "coverage:clean": "rm -rf .nyc_output && rm -rf coverage",
+    "coverage:merge": "istanbul-merge --out coverage/merged/coverage-final.json ./coverage/unit/coverage-final.json  ./coverage/visual-regression/coverage-final.json",
+    "coverage:merge-report": "nyc report --reporter=lcov --reporter=text --temp-dir=./coverage/merged --report-dir=./coverage/merged",
```
oshi97, line 47: For my edification what is coverage:merge-report doing? Which lcov files is it merging together?

juliannzhou: merge-report is only displaying the coverage on screen with the merge json file generated with the merge command

tests/hooks/useRecentSearches.test.tsx
oshi97: Do we need to add actual snapshot testing here? I thought all we wanted was the coverage, we don't need to see if the snapshot has changed.

If we do need this file and DO need it up to date I would rather generate it on the fly instead of checking it into the repo. I'd like to avoid adding additional noise to the repo if possible.

oshi97: Also is `storyshot.test.ts.snap` the default name for the file? It feels a little wordy

oshi97:
> I couldn't find anything on coverage report with Percy snapshot, but there is another option [codecov](https://about.codecov.io/) which works in github but not locally and supports visual, jest, and merge coverage reports. We would need to sign up for an account with them to use the service.

I think Codecov does basically the same thing as Coveralls. The reason we ended up choosing Coveralls over Codecov was a security breach that happened a little bit before we started looking at cloud coverage providers https://blog.gitguardian.com/codecov-supply-chain-breach/

yen-tt:
After digging into visual regression coverage test approaches with Julianne, looks like storybook is moving away from storyshots and to @storybook/test-runner with a code coverage generation option. This requires less setup compare to storyshot (no snapshots needed). We made another PR here: https://github.com/yext/search-ui-react/pull/275

PR #267, opened 2022-07-29
Mapbox Component

alextaing:
This PR introduces the Mapbox Component. This lays the basic foundation for the component, with the interfaces constructed through discussion with Product.

Here are two relevant slack threads:
https://yext.slack.com/archives/C032CKFARGS/p1658322867803349 (inception)
https://yext.slack.com/archives/C032CKFARGS/p1659449732357219 (more recent)

J=SLAP-2221
TEST=manual

Tested through LocationsPage

Later items will add jest tests and storybook for further testing

NOTE: as of now, local test-site requires an `env` variable REACT_APP_MAPBOX_API_KEY for map to work properly
For github workflow to work (in future test items), some changes may be needed in the workflow: https://github.com/facebook/create-react-app/discussions/9064#discussioncomment-19594


src/components/LocationBias.tsx
```diff
@@ -4,6 +4,7 @@ import { getUserLocation } from '../utils/location-operations';
 import { twMerge, useComposedCssClasses } from '../hooks/useComposedCssClasses';
 import { useState } from 'react';
 import LoadingIndicator from '../icons/LoadingIndicator';
+import React from 'react';
```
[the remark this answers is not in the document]
alextaing: Oops that import React shouldn't even be there

src/components/Mapbox.tsx
```diff
@@ -0,0 +1,111 @@
...
+  mapboxOptions,
+  generateMarkerOptions = defaultGenerateMarkerOptions
+}: MapboxProps) {
+  mapboxgl.accessToken = mapboxApiKey;
+  const mapContainer = useRef(null);
+  const map = useRef<Map | null>(null);
+  const markers = useRef<Marker[]>([]);
+
+  const locationResults = useSearchState(state => state.vertical.results);
+  const hasLocationResults = locationResults && locationResults.length !== 0;
```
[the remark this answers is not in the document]
alextaing: unfortunately empty arrays are truthy :0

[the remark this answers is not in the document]
oshi97: you can use optional chaining here i.e. `locationResults?.length !== 0`

src/components/Mapbox.tsx
```diff
@@ -0,0 +1,130 @@
+import React, { useRef, useEffect } from 'react';
```
oshi97, line 1: can remove the React import here too

src/components/Mapbox.tsx
```diff
@@ -0,0 +1,130 @@
...
+import mapboxgl from 'mapbox-gl';
+import { Map, Marker, MapboxOptions, MarkerOptions, LngLatBounds } from 'mapbox-gl';
```
oshi97, line 3: why do we have two imports for mapbox-gl?

src/components/Mapbox.tsx
```diff
@@ -0,0 +1,130 @@
...
+import { Result, useSearchState } from '@yext/search-headless-react';
+import 'mapbox-gl/dist/mapbox-gl.css';
```
oshi97, line 5: This doesn't load the css multiple times if you have multiple maps right? I feel like it wouldn't but if you could check that would be great!

src/components/Mapbox.tsx
```diff
@@ -0,0 +1,130 @@
...
+    const bounds = new LngLatBounds();
+    markers.current = [];
+
+    if (map.current && hasLocationResults) {
+      for (const result of locationResults) {
+        const markerLocation = getCoordinate(result);
+        if (markerLocation) {
+          const { latitude, longitude } = markerLocation;
+          const el = document.createElement('div');
+          PinComponent && ReactDOM.render(<PinComponent result={result} />, el);
```
oshi97, line 98: why are you using ReactDOM.render here?

src/components/Mapbox.tsx
alextaing: I think I need to keep the general import for the accessToken on ln. 62-- is there another way to do it?  (if I keep it, ill move it down so they're on the same line

src/components/Mapbox.tsx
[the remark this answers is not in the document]
alextaing: oo OK, I wasn't sure if that was subtly different or not

src/components/Mapbox.tsx
```diff
@@ -0,0 +1,130 @@
...
+ */
+export function Mapbox<TEntity>({
+  mapboxApiKey,
+  mapboxOptions,
+  markerOptions,
+  PinComponent,
+  defaultCenter: center,
+  getCoordinate
+}: MapboxProps<TEntity>) {
+  mapboxgl.accessToken = mapboxApiKey;
```
oshi97, line 62: should this be in a useEffect? is this the normal way of setting the mapbox api key?

src/components/Mapbox.tsx
alextaing: This was one of the things that I wanted to go back and revisit-- I took this small 2 line snippet from davish's example since he implemented it in a way that we can use imported pin components-- Looks like renders the pin to a div, then we insert that div into the Map.  Is there a better way to do that?  I'll of course look around.

src/components/Mapbox.tsx
```diff
@@ -0,0 +1,130 @@
...
+  const mapContainer = useRef(null);
+  const map = useRef<Map | null>(null);
+  const markers = useRef<Marker[]>([]);
+
+  const locationResults = useSearchState(state => state.vertical.results) as Result<TEntity>[];
+  const hasLocationResults = locationResults && locationResults.length !== 0;
+
+  const mapboxStyle = mapboxOptions?.style ?? 'mapbox://styles/mapbox/streets-v11';
+
+  useEffect(() => {
```
oshi97, line 72: should these effects have dependency arrays, or do they really need to run every single render?

src/components/Mapbox.tsx
alextaing: Yeah, its the way that do it in their React [example](https://docs.mapbox.com/help/tutorials/use-mapbox-gl-js-with-react/)-- they don't seem to put in in a hook, but I'll give it some thought

src/components/Mapbox.tsx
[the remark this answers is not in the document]
oshi97: nope it's the exact same thing as the && check you do here! besides being a little more concise

src/components/Mapbox.tsx
oshi97: ahh yeah makes sense! we can keep it, let's just combine it like `import mapboxgl, { blah } from 'mapbox-gl'`

src/components/Mapbox.tsx
oshi97: To check this you could try looking at the css that gets loaded onto the page

src/components/Mapbox.tsx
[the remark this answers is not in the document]
alextaing: turns out they are slightly different! 

If locationResults was `undefined`, the && way would return `undefined` while optional chaining would return `true`.
The component needs to follow the former behavior to work, so I changed it back.

src/components/Mapbox.tsx
alextaing: Talked with yen and it looks like the CSS is bundled in

src/components/Mapbox.tsx
alextaing: ahh no it definitely does not need to run every render

src/components/Mapbox.tsx
oshi97: it's not bundled in multiple times right?

alextaing:
Hello!! For whoever is working on this after I leave-- thank you! Also, I was thinking about another aspect users might want to configure is the area on the map where pins can show.  For example, if for some reason, the user wanted to have a floating window with results, they may want pins to only show on the right side of the window since the pins on the left would be covered by the overlay.  

image

Or, if users had different sized pins, they might need to configure the marker margins to make sure no markers get cut off. Overall, I feel like if the marker margins are off, it makes the map look off.

src/components/Mapbox.tsx
alextaing: Nope! Just once

alextaing:
Another thing that I might want to consider is adding some sort of size styling?  Right now, I have it such that the map fills whatever space it has-- so if the user wants to make a map of a certain size, they have to place it in a div with a specified size.  But not necessary, since we're trying to prune prop interfaces.

src/components/Mapbox.tsx
alextaing: Since we only need to set the API key once and not at every render, I wrapped it. Just out of curiosity, is there some sort of overhead/trade-off when using the useEffect hook, or is it still worth using for small things like var assignments?

oshi97: There's some overhead for each new effect you have. In this case that overhead is probably greater than any work you would save by not setting the apiKey extra times (setting property values is extremely fast). This is more on principal of clearly defining side effects. Generally speaking putting effects in useEffect/useLayoutEffect will have more consistent behavior than putting it in the render.

alextaing:
The pin component was just something that I added in to test compatibility with a custom React component.  It is a simplified version of the pin component in Davish's search-locator-starter.  It can be removed, and the location page can be refactored when the final PR is ready.

alextaing:
The part with the generic `TEntity` and the coordinate getter is still a little fuzzy, and I think that there is probably a much better way to implement that idea than what I have at the moment.  Also the code organization is weird: I moved the options interfaces to the bottom since they took up a lot of space near the top of the file and looked messy, but they also looks strange no matter where I put them-- maybe they can be extracted into another file?

src/components/MapboxMap.tsx
```diff
@@ -0,0 +1,157 @@
...
+  }, [mapboxOptions, onDragDebounced]);
+
+  useEffect(() => {
+    markers.current.forEach(marker => marker.remove());
+    markers.current = [];
+
+    if (map.current && hasLocationResults) {
+      const bounds = new LngLatBounds();
+
+      for (const [index, result] of locationResults.entries()) {
```
nmanu1, line 108: can this use declarative forEach instead?

src/hooks/useDebouncedFunction.ts
```diff
@@ -5,10 +5,13 @@ type Func = (...args: any[]) => any;
 type DebouncedFunction<F extends Func> = (...args: Parameters<F>) => Promise<undefined | ReturnType<F>>;
 
 export function useDebouncedFunction<F extends Func>(
-  func: F,
+  func: F | undefined,
```
nmanu1, line 8: instead of letting this be undefined, could we just check in the Map component if onDrag is defined before using this hook?

src/components/MapboxMap.tsx
```diff
@@ -0,0 +1,157 @@
...
+  const locationResults = useSearchState(state => state.vertical.results) as Result<T>[];
+  const hasLocationResults = locationResults && locationResults.length !== 0;
+  const onDragDebounced = useDebouncedFunction(onDrag, 100);
+
+  useEffect(() => {
+    if (mapContainer.current && !map.current) {
+      const options: MapboxOptions = {
+        container: mapContainer.current,
+        style: 'mapbox://styles/mapbox/streets-v11',
+        center: [-73.935242, 40.730610],
```
nmanu1, line 86: where is this value from?

src/components/MapboxMap.tsx
```diff
@@ -0,0 +1,157 @@
...
+          const { latitude, longitude } = markerLocation;
+          const el = document.createElement('div');
+          PinComponent && ReactDOM.render(<PinComponent result={result} index={index} />, el);
+          const markerOptions: MarkerOptions = {
+            element: PinComponent ? el : undefined
+          };
+          const marker = new Marker(markerOptions)
+            .setLngLat({ lat: latitude, lng: longitude })
+            .addTo(map.current);
+          markers.current = [marker, ...markers.current];
```
nmanu1, line 120: is this order important? or could we just push to the array?

src/components/MapboxMap.tsx
```diff
@@ -0,0 +1,157 @@
...
+          bounds.extend([longitude, latitude]);
+        }
+      }
+
+      if (!bounds.isEmpty()){
+        map.current.fitBounds(bounds, {
+          padding: { top: 50, bottom: 25, left: 50, right: 50 },
```
nmanu1, line 127: out of curiosity, is there a reason less padding is needed at the bottom?

test-site/src/components/MapPin.tsx
```diff
@@ -0,0 +1,20 @@
+import { PinComponent } from "@yext/search-ui-react";
+import { Location } from '../pages/LocationsPage'
+
+export const MapPin: PinComponent<Location> = () => {
```
nmanu1, line 4: do we need this export if we already export the MapPin as the default below?

src/components/MapboxMap.tsx
```diff
@@ -0,0 +1,157 @@
...
+  getCoordinate?: CoordinateGetter<T>,
+  /** A function which is called when user drag the map. */
+  onDrag?: (center: LngLat, bounds: LngLatBounds) => void
+}
+
+/**
+ * A component that renders a map with markers to show result locations using Mapbox GL.
+ *
+ * @remarks
+ * For the map to work properly, be sure to include Mapbox GL stylesheet in the application.
```
nmanu1, line 55: do we want to include the exact import statement somewhere in the docs?

src/components/MapboxMap.tsx
```diff
@@ -0,0 +1,157 @@
...
+  useEffect(() => {
+    mapboxgl.accessToken = mapboxAccessToken;
+  }, [mapboxAccessToken]);
+
+  const mapContainer = useRef<HTMLDivElement>(null);
+  const map = useRef<Map | null>(null);
+  const markers = useRef<Marker[]>([]);
+
+  const locationResults = useSearchState(state => state.vertical.results) as Result<T>[];
+  const hasLocationResults = locationResults && locationResults.length !== 0;
```
nmanu1, line 78: can we use optional chaining here, something like `locationResults?.length > 0`?

src/components/MapboxMap.tsx
```diff
@@ -0,0 +1,157 @@
...
+/**
+ * Props for the {@link MapboxMap} component.
+ *
+ * @public
+ */
+export interface MapboxMapProps<T> {
+  /** Mapbox access token. */
+  mapboxAccessToken: string,
+  /** Interface for customizing the {@link MapboxMap} map derived from Mapbox API's Map options. */
+  mapboxOptions?: Omit<MapboxOptions, 'container'>,
```
nmanu1, line 42: is styling customizable through these options?

src/components/MapboxMap.tsx
```diff
@@ -0,0 +1,157 @@
...
+
+/**
+ * Props for the {@link MapboxMap} component.
+ *
+ * @public
+ */
+export interface MapboxMapProps<T> {
+  /** Mapbox access token. */
+  mapboxAccessToken: string,
+  /** Interface for customizing the {@link MapboxMap} map derived from Mapbox API's Map options. */
```
nmanu1, line 41: nit: this doc reads a little funny: "... the MapboxMap map derived ...."

test-site/src/pages/LocationsPage.tsx
```diff
@@ -0,0 +1,80 @@
...
+          <div className='flex-grow'>
+            <div className='flex items-baseline'>
+              <ResultsCount />
+              <AppliedFilters />
+            </div>
+            <div className='flex flex-col'>
+              <div className='h-80 mb-4'> 
+                <MapboxMap<Location>
+                mapboxAccessToken='pk.eyJ1IjoieWV4dCIsImEiOiJqNzVybUhnIn0.hTOO5A1yqfpN42-_z_GuLw'
+                mapboxOptions={{zoom: 10 }}
```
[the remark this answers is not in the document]
yen-tt: looks like eslint is not running inside test-site

src/components/MapboxMap.tsx
yen-tt: typescript fail to recognize the check if(map.current && ...) when using `forEach` -- complains about `map.current` being possibly null. This works fine for a normal for loop. There is workaround for it to use forEach so I can update to that

src/hooks/useDebouncedFunction.ts
yen-tt: React Hooks can't be called conditionally

nmanu1: ah right

src/components/MapboxMap.tsx
yen-tt: It's what Alex added in his previous commits (https://github.com/yext/search-ui-react/pull/267/commits/453bc99825eaa0ce10d82a9d5ae1e74c39e77909), which I assume is through his discussion with Product. I can double check again once Product is back

src/components/MapboxMap.tsx
yen-tt: don't think so, pushing to array should be fine

src/components/MapboxMap.tsx
yen-tt: There are multiple ways to include the stylesheet so it's up to the user -- they could either import it like our test-site like `import 'mapbox-gl/dist/mapbox-gl.css';` or add a link like `<link href="https://api.mapbox.com/mapbox-gl-js/VERSION/mapbox-gl.css" rel="stylesheet">` on their html page.

I can update the doc to include both examples

test-site/src/pages/LocationsPage.tsx
```diff
@@ -0,0 +1,80 @@
...
+    }
+    searchActions.setStaticFilters([...nonLocationFilters, nearFilter])
+    searchActions.executeVerticalQuery()
+  }, [filters, searchActions])
+
+  return (
+      <div>
+        <SearchBar />
+        <div className='flex'>
+          <div className='min-w-fit pr-4'>
```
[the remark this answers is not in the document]
yen-tt: updated, removed a couple things here

test-site/src/pages/LocationsPage.tsx
```diff
@@ -0,0 +1,80 @@
...
+          </div>
+          <div className='flex-grow'>
+            <div className='flex items-baseline'>
+              <ResultsCount />
+              <AppliedFilters />
+            </div>
+            <div className='flex flex-col'>
+              <div className='h-80 mb-4'> 
+                <MapboxMap<Location>
+                mapboxAccessToken='pk.eyJ1IjoieWV4dCIsImEiOiJqNzVybUhnIn0.hTOO5A1yqfpN42-_z_GuLw'
```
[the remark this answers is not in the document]
yen-tt: it's what Alex added in his previous commit. I assume it was provided by Product, but I can double check

src/components/MapboxMap.tsx
yen-tt: don't think so, I can make it all 50

src/components/MapboxMap.tsx
yen-tt: sure

src/components/MapboxMap.tsx
yen-tt: are you referring to the style like `style: 'mapbox://styles/mapbox/streets-v11'`? If so, yup that field is included in `MapboxOptions` interface

test-site/src/components/MapPin.tsx
yen-tt: removed the default export

test-site/src/pages/LocationsPage.tsx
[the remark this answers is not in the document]
yen-tt: `eslintConfig` field in test-site was causing issues with the linter using the rules defined in `eslintrc` file. Updated now, so eslint should lint test-site properly

src/components/MapboxMap.tsx
yen-tt: the type for `style` is: ` style?: mapboxgl.Style | string | undefined;`. User can either select one of the predefined styles (https://docs.mapbox.com/api/maps/styles/) or make their own following the interface of `mapboxgl.Style`:
Screen Shot 2022-09-26 at 4 42 01 PM

src/components/MapboxMap.tsx
```diff
@@ -0,0 +1,163 @@
...
+    markers.current = [];
+    const mapbox = map.current;
+    if (mapbox && locationResults?.length > 0) {
+      const bounds = new LngLatBounds();
+      locationResults.forEach((result, i) => {
+        const markerLocation = getCoordinate(result);
+        if (markerLocation) {
+          const { latitude, longitude } = markerLocation;
+          const el = document.createElement('div');
+          PinComponent && ReactDOM.render(<PinComponent result={result} index={i} />, el);
```
nmanu1, line 119: is there a default pin component that can be used for markers if user's don't specify a custom component?

yen-tt: mapbox provide a default pin already if there's no custom one provided -- product is fine with that

src/components/MapboxMap.tsx
```diff
@@ -0,0 +1,163 @@
...
+ */
+export interface Coordinate {
+  /** The latitude of the location. */
+  latitude: number,
+  /** The longitude of the location. */
+  longitude: number
+}
+
+/**
+ * Props for the {@link MapboxMap} component.
```
nmanu1, line 34: can you update this doc to mention what the type param is for?

src/components/MapboxMap.tsx
```diff
@@ -0,0 +1,163 @@
...
+ * @public
+ */
+export interface MapboxMapProps<T> {
+  /** Mapbox access token. */
+  mapboxAccessToken: string,
+  /** Interface for map customization derived from Mapbox GL's Map options. */
+  mapboxOptions?: Omit<MapboxOptions, 'container'>,
+  /** Custom Pin component to render for markers on the map. */
+  PinComponent?: PinComponent<T>,
+  /** A function to derive a result's coordinate for the corresponding marker's location on the map. */
```
nmanu1, line 45: can we include what the default behavior is here if no function is specified?

src/components/MapboxMap.tsx
```diff
@@ -0,0 +1,163 @@
...
+    if (mapbox && locationResults?.length > 0) {
+      const bounds = new LngLatBounds();
+      locationResults.forEach((result, i) => {
+        const markerLocation = getCoordinate(result);
+        if (markerLocation) {
+          const { latitude, longitude } = markerLocation;
+          const el = document.createElement('div');
+          PinComponent && ReactDOM.render(<PinComponent result={result} index={i} />, el);
+          const markerOptions: MarkerOptions = {
+            element: PinComponent ? el : undefined
```
nmanu1, line 121: instead of checking if the custom PinComponent is specified twice, can the code be wrapped in a single `if` check?

test-site/package.json
```diff
@@ -26,12 +26,6 @@
     "test": "react-scripts test",
     "eject": "react-scripts eject"
   },
-  "eslintConfig": {
```
nmanu1, line 29: now that this config was deleted, I'm sometimes seeing an error like this in the test-site:
`ERROR in Plugin "react" was conflicted between "../.eslintrc.js » @yext/eslint-config-slapshot/typescript-react » eslint-config-react-app » ../node_modules/eslint-config-react-app/base.js" and "BaseConfig » ../test-site/node_modules/eslint-config-react-app/base.js".`

have you gotten this too?

test-site/src/pages/LocationsPage.tsx
```diff
@@ -0,0 +1,76 @@
...
+  return (
+    <div>
+      <SearchBar />
+      <div className='flex flex-col'>
+        <div className='flex items-baseline'>
+          <ResultsCount />
+          <AppliedFilters />
+        </div>
+        <div className='h-80 mb-4'>
+          <MapboxMap<Location>
```
nmanu1, line 61: do we need to specify the type param or can it be inferred from the props?

test-site/src/pages/LocationsPage.tsx
```diff
@@ -0,0 +1,76 @@
...
+  useLayoutEffect(() => {
+    searchActions.setVertical('KM');
+    searchActions.executeVerticalQuery();
+  }, [searchActions]);
+
+  const onDrag: MapboxMapProps<Location>['onDrag'] = useCallback(
+    (center: LngLat, bounds: LngLatBounds) => {
+      const radius = center.distanceTo(bounds.getNorthEast());
+      const nonLocationFilters: SelectableStaticFilter[] = filters?.filter(f => f.filter.kind !== 'fieldValue' || f.filter.fieldId !== 'builtin.location') ?? [];
+      const nearFilter: SelectableStaticFilter = {
```
nmanu1, line 40: nit: do we want to give this filter some kind of generic displayName so applied filters doesn't just display it as a blank pill with no text?

test-site/src/pages/LocationsPage.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+export function LocationsPage() {
+  const searchActions = useSearchActions();
+  const filters = useSearchState(state => state.filters.static);
+  useLayoutEffect(() => {
+    searchActions.setVertical('KM');
+    searchActions.executeVerticalQuery();
+  }, [searchActions]);
+
+  const onDrag: MapboxMapProps<Location>['onDrag'] = useCallback(
```
nmanu1, line 36: do we want to export the type of this function like we do for `getCoordinate`? then users wouldn't need to reach into the props like this to get the type

src/components/MapboxMap.tsx
```diff
@@ -0,0 +1,163 @@
...
+ * Props for the {@link MapboxMap} component.
+ *
+ * @public
+ */
+export interface MapboxMapProps<T> {
+  /** Mapbox access token. */
+  mapboxAccessToken: string,
+  /** Interface for map customization derived from Mapbox GL's Map options. */
+  mapboxOptions?: Omit<MapboxOptions, 'container'>,
+  /** Custom Pin component to render for markers on the map. */
```
nmanu1, line 43: can this doc mention that this component would replace the default pin from mapbox?

src/components/MapboxMap.tsx
yen-tt: I like to have the construction of markerOptions all in one place instead of doing `markerOptions.element = el` separately so I kept it like this. But eh, if check is cool too, I can update

test-site/package.json
yen-tt: I have not

src/components/MapboxMap.tsx
nmanu1: I see. I don't have a strong preference, so this is fine

test-site/package.json
nmanu1: oh looks like re-compiling fixed it

test-site/src/pages/LocationsPage.tsx
yen-tt: sure thing, update to have 'Near Current Area' as the display name

test-site/package.json
yen-tt: okay I just saw the error after a couple reinstall and reload of test-site. From the look of this [thread](https://stackoverflow.com/questions/70449712/create-react-app-error-in-plugin-react-was-conflicted-between-eslintrc-jso) and the error message, I think it's two versions of the react eslint plugin being used. I update test-site package.json to point eslint-config-react-app to the top level node_modules one. The issue is gone from my end now

test-site/src/pages/LocationsPage.tsx
yen-tt: It helps when I had the props inline before. After moving things out to use usecallback, it's not needed anymore. removed!

test-site/src/pages/LocationsPage.tsx
yen-tt: sure!

src/components/MapboxMap.tsx
yen-tt: Product doesn't mind what default center is set to -- I end up switching it to Yext NYC office coordinate

src/components/MapboxMap.tsx
```diff
@@ -0,0 +1,179 @@
...
+ * A function which is called when user drag the map.
+ *
+ * @public
+ */
+export type OnDragHandler = (center: LngLat, bounds: LngLatBounds) => void;
+
+/**
+ * Props for the {@link MapboxMap} component.
+ *
+ * @typeParam T - Type of "rawData" field of the results use in the map
```
nmanu1, line 43: is `typeParam` a tag that API extractor recognizes? I think it might just ignore this and not add it to the docs

yen-tt: I thought it does looking at their docs: https://api-extractor.com/pages/tsdoc/tag_typeparam/. But looking at our generated docs (seems like it never got implemented? https://github.com/microsoft/rushstack/issues/1635). I will just update this to be a normal comment then.

yen-tt:
rewrote history to remove LocationsPage. Will open another pr

nmanu1 (approved):

nmanu1:
is the failing Snyk license approved by legal?

yen-tt:
> is the failing Snyk license approved by legal?

Product approved the usage of Mapbox GL JS Version 2.0 and its pricing model, but is still working on the license approval.  They gave permission with cutting a beta version -- I will merge this into a feature branch instead for now.

src/components/Mapbox.tsx
yen-tt: I didn't find a good way to do this as Mapbox's marker option takes in HTMLElement only for the DOM element to use as a marker, so this needs to be pre-rendered before passing into new Marker()

PR #268, opened 2022-07-29
Fix bug with applying query filters

nmanu1:
This PR fixes the bug from [this Slack thread](https://yext.slack.com/archives/C032CKFARGS/p1658949929942189)  where query filters were not being applied except when hot-reloading the page. Applied query filters are not returned in the response if an empty array of facet options is sent in the request. Because facets were de-selected rather than cleared out completely when executing a new search from the search bar, no query filters were applied.

J=none
TEST=manual

Check that the correct filters are applied automatically when searching "Marty", "Bleecker", and "Marty frodo". If they correspond to a facet, then the applied filters are removable and don't get "stuck" as applied. Query filters are only applied again after a new search is run from the search bar.

### DIFF
--- a/src/components/SearchBar.tsx
+++ b/src/components/SearchBar.tsx
@@ -231,7 +231,7 @@ export function SearchBar({
   const handleSubmit = useCallback((value?: string, index?: number, itemData?: FocusedItemData) => {
     value !== undefined && searchActions.setQuery(value);
     searchActions.setOffset(0);
-    searchActions.resetFacets();
+    searchActions.setFacets([]);
     clearStaticRangeFilters(searchActions);
     if (itemData && isVerticalLink(itemData.verticalLink) && onSelectVerticalLink) {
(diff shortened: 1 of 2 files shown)


yen-tt (commented):
We may need to update other places that uses resetFacets as well? For instance, FiltersSearch would call resetFacets before executing search, which may tell backend to exclude facet options if it's an empty array in the request when it shouldn't.

nmanu1:
> We may need to update other places that uses resetFacets as well? For instance, FiltersSearch would call resetFacets before executing search, which may tell backend to exclude facet options if it's an empty array in the request when it shouldn't.

The other places we use `resetFacets` are correct, including in `FilterSearch`. An empty array of facet options only indicates that the backend shouldn't return applied query filters (aka NLP filters), not regular facets. So when a new search is conducted from `FilterSearch` or the Apply Filters button, we don't want the query filters that have been removed to be re-applied. The only time we want new query filters to be applied is when the user makes a new query, which is why the only update that's needed is in `SearchBar`.

yen-tt (approved):

PR #272, opened 2022-08-02
Add Test Site to Sites

yen-tt:
- added ci_config.json and .nvmrc to publish test site to Sites

J=SLAP-2290
TEST=manual

Temporarily point staging to dev/add-to-sites branch and see that test-site is deployed [successfully](https://devaddtosites-perpetual--illusive--cringing-pgsdemo-com.preview.pagescdn.com/).

Will update production to point to `test-site` after merging.


.github/workflows/sync-testsite-branch.yml
```diff
@@ -0,0 +1,18 @@
+name: Sync Main to Storybook Site Branch
```
nmanu1, line 1: nit: name still refers to Storybook Site

.github/workflows/sync-testsite-branch.yml
```diff
@@ -0,0 +1,18 @@
...
+
+on:
+  push:
+    branches:
+      - main
+jobs:
+  sync-branches:
```
nmanu1, line 8: is it possible to combine this workflow with the storybook site one so we don't need two?

.github/workflows/sync-testsite-branch.yml
nmanu1, line 1: I think this workflow needs to be merged into main so it runs on commits

yen-tt: oops

sites-config/ci_config.json
```diff
@@ -0,0 +1,21 @@
...
+      "assets": [
+          {
+              "root": "test-site/build"
+          }
+      ],
+      "hbsTemplates": []
+  },
+  "dependencies": {
+      "installDepsCmd": "npm install",
+      "requiredFiles": [".nvmrc"]
```
oshi97, line 12: I guess .nvmrc is not included by default but package.json and package-lock are?

sites-config/ci_config.json
```diff
@@ -0,0 +1,21 @@
+{
```
oshi97, line 1: oh was looking at the sites starter for unrelated studio stuff and noticed this file can just be ci.json now?
https://github.com/yext/pages-starter-react-locations/blob/main/sites-config/ci.json

sites-config/ci_config.json
yen-tt: This is what I found on slack:
"the specified required_files are copied into the docker image, and then the install_deps cmd is ran. This happens  before the entire repo is copied into the image. So every file which is required by the install_deps command needs to be specified as a required file."
I assume the package files would have to be default since it installed and built fine in Sites. But I will update to include it nontheless.

sites-config/ci_config.json
yen-tt: cool, I will update!

.github/workflows/sync-testsite-branch.yml
yen-tt: I kinda like it separate..but sure

.github/workflows/sync-testsite-branch.yml
yen-tt: yep

.github/workflows/sync-testsite-branch.yml
nmanu1: maybe just a callable workflow that's reused with a variable passed in if we want to keep them as separate?

.github/workflows/sync-testsite-branch.yml
yen-tt: made a pr!: https://github.com/yext/search-ui-react/pull/273

.github/workflows/sync-testsite-branch.yml
yen-tt: eh, combining it is fine

.github/workflows/sync-testsite-branch.yml
nmanu1: nice! in that case, can we delete the workflow changes from this PR? the syncing isn't needed on this branch technically, and it would be automatically merged in when the other PR gets merged into main

yen-tt: yee. Removed changes!

tmeyer2115 (approved):

PR #274, opened 2022-08-02
Add WCAG PR check

nmanu1:
Add a GH workflow to check for WCAG violations. If the check fails, the PR will be blocked from merging into main. There are some WCAG violations that we ignore, namely color contrast failures on stories where components are in a "loading" state. Using config options, that rule is disabled for those particular stories. To get the automated WCAG checks, Storybook has to be running while the `npm run wcag` command is run.

Note, Jest had to be downgraded to v27 to work with `@storybook/test-runner`. But, it looks like they will be adding support for v28 shortly (https://github.com/storybookjs/test-runner/pull/154), so we should be able to upgrade again soon. In the meantime, we haven't been using any features from v28, so the downgrade didn't affect any tests.

Also, the Snyk checks are failing, but the new license and one of new vulnerabilities are both approved by legal in Juliann's [PR](https://github.com/yext/search-ui-react/pull/264#issue-1321477086). The other new vulnerability is also introduced only through Storybook, so we wouldn't be susceptible to it.

J=SLAP-2289
TEST=auto

See that the check runs as expected on this PR and passes.


.github/workflows/wcag_test.yml
```diff
@@ -0,0 +1,14 @@
...
+    branches:
+      - main
+
+jobs:
+  call_wcag_test:
+    uses: yext/slapshot-reusable-workflows/.github/workflows/wcag_test.yml@v1
+    secrets:
+      NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+    with:
+      build_script: 'npm run storybook & sleep 10 && curl http://localhost:6006 -I'
```
oshi97, line 14: I'm sure you looked into this but can we avoid the `sleep 10`? Maybe some option in start-storybook?

nmanu1: Yeah, I tried finding a way to not use `sleep` and checked again in the `start-storybook` options. The closest option to what we want is --smoke-test, but that stops serving the site once it successfully starts. We need the site to be running when the `test-storybook` command is run. The only other way I saw to do this as part of a GH workflow was either to host the site somewhere public, or involved setting up [service containers](https://docs.github.com/en/actions/using-containerized-services/about-service-containers). Service containers seemed like they might be overkill, but I can look into it more

oshi97: is something like this possible? I don't have any experience trying stuff like this https://stackoverflow.com/questions/21475639/wait-until-service-starts-in-bash-script

.storybook/test-runner.cjs
```diff
@@ -0,0 +1,35 @@
...
+/*
+* See https://storybook.js.org/docs/react/writing-tests/test-runner#test-hook-api-experimental
+* to learn more about the test-runner hooks API.
+*/
+module.exports = {
+  async preRender(page) {
+    await injectAxe(page);
+  },
+  async postRender(page, context) {
+    const axeOptions = context.name === 'Loading'
```
alextaing, line 19: Does this mean that the name of the view must be exactly "Loading" in order for it to apply the loading config?  I'm not sure if there would be a situation where we would have more than one loading view for a story, but would it be good to make it so that it'll apply this config if "loading" or "Loading" appears anywhere in the name?

nmanu1: Yes, the story name would need to be an exact match. I thought about being more generic, but I'd prefer to err on the side of caution and only disable rules for the fewest cases right now. If we start writing other loading stories with different naming conventions, we could easily expand this matching then

.github/workflows/wcag_test.yml
nmanu1: sure! I updated it to be a while loop instead so it keeps checking every second, instead of having an arbitrary sleep time of 10 seconds

oshi97: nice!

.storybook/test-runner.cjs
```diff
@@ -0,0 +1,35 @@
+const { injectAxe, checkA11y } = require('axe-playwright');
```
oshi97, line 1: can we use typescript for this file?

.github/workflows/wcag_test.yml
```diff
@@ -0,0 +1,15 @@
...
+      - main
+
+jobs:
+  call_wcag_test:
+    uses: yext/slapshot-reusable-workflows/.github/workflows/wcag_test.yml@v1
+    secrets:
+      NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+    with:
+      build_script:
+        'npm run storybook & while ! curl http://localhost:6006 -I; do sleep 1; done'
```
oshi97, line 15: instead of downgrading jest for the whole repo would it be possible to manually install jest@27 just for the wcag tests as part of the workflow?

package.json
```diff
@@ -40,44 +40,46 @@
...
     "@babel/preset-react": "^7.16.7",
     "@babel/preset-typescript": "^7.14.5",
-    "@percy/cli": "^1.0.7",
-    "@percy/storybook": "^4.2.0",
-    "@storybook/addon-a11y": "^6.4.20",
-    "@storybook/addon-actions": "^6.4.19",
-    "@storybook/addon-essentials": "^6.4.19",
-    "@storybook/addon-interactions": "^6.4.19",
-    "@storybook/addon-links": "^6.4.19",
+    "@percy/cli": "^1.7.2",
```
oshi97, line 51: wow this is a big bump 👀 
I checked out of curiosity for what changed and it looks like 1.8.0 was released 18 hours ago?

.github/workflows/wcag_test.yml
nmanu1: I imagined it would also be nice to be able to run the wcag command locally to test, which would mean we would need to install v27 locally for that too. and I think we'd also need to do the same for the visual coverage stuff. so it seemed easier to downgrade the whole repo for now

oshi97: that's fair!

.storybook/test-runner.cjs
nmanu1: updated!

package.json
nmanu1: upgraded percy/cli and storybook because it looks like they just released 6.5.10 yesterday

oshi97: oh I like to use https://www.npmjs.com/package/npm-check-updates to see what the latest versions for things in my package are, if you aren't using it already (it only reads from the package.json and not your package-lock)

oshi97: I didn't use it earlier for percy/cli though that was just me going to their github page

.storybook/test-runner.ts
```diff
@@ -0,0 +1,29 @@
+import { injectAxe, checkA11y } from 'axe-playwright';
+import { Page } from 'playwright-core';
+
+/*
+* See https://storybook.js.org/docs/react/writing-tests/test-runner#test-hook-api-experimental
+* to learn more about the test-runner hooks API.
+*/
+module.exports = {
```
oshi97, line 8: should this be es6 syntax instead i.e. `export default`?

.github/workflows/wcag_test.yml
oshi97, line 15: is there a way to make npm run storybook output a little quieter? if you look at the github action logs there's 1000+ lines for this step https://github.com/yext/search-ui-react/runs/7676458663?check_suite_focus=true#logs

nmanu1: updated! now it's <200 lines

.github/workflows/wcag_test.yml
```diff
@@ -0,0 +1,19 @@
...
+      - main
+
+jobs:
+  call_wcag_test:
+    uses: yext/slapshot-reusable-workflows/.github/workflows/wcag_test.yml@v1
+    secrets:
+      NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+    with:
+      build_script: |
+        npm run storybook -- --quiet &
```
yen-tt, line 15: can we use the same approach as the one for visual-coverage script? with max time out and terminate if storybook is not served?

.github/workflows/wcag_test.yml
```diff
@@ -0,0 +1,19 @@
...
+        until curl http://localhost:6006 -I --silent
```
yen-tt, line 16: can we use the same approach as the one for visual-coverage script? with max time out and terminate if storybook is not served?

.storybook/test-runner.ts
```diff
@@ -0,0 +1,31 @@
...
+  async postRender(page: Page, context) {
+    await checkA11y(page, '#root', {
+      axeOptions: {
+        runOnly: {
+          type: 'tag',
+          values: ['wcag2a', 'wcag2aa', 'wcag21a', 'wcag21aa']
+        },
+        rules: {
+          'color-contrast': { enabled: context.name !== 'Loading' }
+        },
```
yen-tt, line 21: we also defined this in preview.js (excluding rules here, which we should probably add to preview.js), can we have a file containing this configuration that is use in both preview and test-runner?

.storybook/test-runner.ts
```diff
@@ -0,0 +1,29 @@
+import { injectAxe, checkA11y } from 'axe-playwright';
+import { Page } from 'playwright-core';
+import { parameters } from './preview';
+
+/*
```
oshi97, line 5: extra space here at the start of the line

.storybook/test-runner.ts
```diff
@@ -0,0 +1,29 @@
...
+* See https://storybook.js.org/docs/react/writing-tests/test-runner#test-hook-api-experimental
```
oshi97, line 6: The indentation is a little off, I think we're missing a space before these asterisks.
I think the first first line should have two asterisks like `/**`

oshi97: If you type `/**` in vscode and hit enter it should automatically format stuff nicely for you, if you only type `/*` at least on my machine you don't get the formatting

.storybook/test-runner.ts
nmanu1: I moved the common part to another file, but as you mentioned, the rules can't go in preview.js because we don't have context there for the story's name. we would have to configure that individually for each Loading story

.github/workflows/wcag_test.yml
nmanu1: moved the common part into a separate script that's used for both

.storybook/wcagConfig.ts
```diff
@@ -0,0 +1,6 @@
+import { axeOptionsConfig } from 'axe-playwright';
```
oshi97, line 1: Does this need to be a separate file? If it's only ever going to be used in test-runner.ts we can put them together

nmanu1: yen preferred it to be separate so it can be used in both test-runner.ts and preview.js

oshi97 (approved):

PR #281, opened 2022-08-05
fix edge cases from filtersearch-option pr

ElemelonWind:
fixed two edge cases that Juliann pointed out
- https://github.com/yext/search-ui-react/pull/256#issuecomment-1205532847
- https://github.com/yext/search-ui-react/pull/256#issuecomment-1205658494

deselecting the filter `onToggle` messed up these two behaviors because it didn't recognize the input in the search filter as a selected filter, and decided not to clear it / keep the filter on the page. 

TEST=auto

### DIFF
--- a/tests/components/FilterSearch.test.tsx
+++ b/tests/components/FilterSearch.test.tsx
@@ -442,6 +442,8 @@ it('clears input when old filters are removed', async () => {
     expect(setFilterOption).toHaveBeenCalledWith({ ...deselectedFilter, selected: true });
   });
 
+  userEvent.click(searchBarElement);
+
   const mockDeselectButton = screen.getByRole('button');
   userEvent.click(mockDeselectButton);
 
--- a/tests/components/FilterSearch.test.tsx
+++ b/tests/components/FilterSearch.test.tsx
@@ -442,8 +442,6 @@ it('clears input when old filters are removed', async () => {
     expect(setFilterOption).toHaveBeenCalledWith({ ...deselectedFilter, selected: true });
   });
 
-  userEvent.click(searchBarElement);
-
   const mockDeselectButton = screen.getByRole('button');
   userEvent.click(mockDeselectButton);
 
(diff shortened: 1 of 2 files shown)


oshi97 (commented):
can we add a test? generally it's good to add a test after you fix a bug.

It's a little counterintuitive, but putting TEST=auto on a PR doesn't mean you ran the unit tests, it means you added additional automated coverage. There's generally no need to say you ran the unit tests because those get run automatically on PR anyways

tests/components/FilterSearch.test.tsx
```diff
@@ -442,6 +442,8 @@ it('clears input when old filters are removed', async () => {
     expect(setFilterOption).toHaveBeenCalledWith({ ...deselectedFilter, selected: true });
   });
 
+  userEvent.click(searchBarElement);
```
oshi97, line 445: I think it's not really clear how adding this line tests this case. If somebody 6 months from now had to come in and change the tests due to a refactor, or if somebody accidentally triggers this test to fail, it might be hard to understand what isn't working. Is there some way we could make it clearer?

oshi97: Adding a comment is OK but a bit of a bandaid option, there should be better ways

ElemelonWind: Hmm. Should I add something to the title of the test?

ElemelonWind: Or I could make a new test, but it's basically the exact same thing just with clicking into the input to check for the toggle behavior

oshi97: Could you make a new test for just this case? For unit tests we generally want short tests that each only test a single thing. This makes them easier to understand and also easier to refactor in the future if we change the behavior.

oshi97: For instance this current FilterSearch test case is actually on the long side and ideally would be shorter

tests/components/FilterSearch.test.tsx
```diff
@@ -451,4 +451,39 @@ it('clears input when old filters are removed', async () => {
...
+  userEvent.type(searchBarElement, '{enter}');
+
+  await waitFor(() => {
+    expect(executeFilterSearch).toHaveBeenCalled();
+  });
+
+  await waitFor(() => {
+    expect(setFilterOption).toBeCalledTimes(1);
+  });
+
```
oshi97, line 484: is it possible to move this boilerplate to a function, and share it with the other tests that do the same thing? That 1. helps reduce repeated code and 2. helps describe what this block of code is doing by giving it a name (self documenting code)

oshi97: I highlighted the lines when writing this comment but it looks like it didn't come through, by boilerplate I mean from the start of the test down to before we click on searchBarElement

ElemelonWind: I think all of the tests do something slightly different; like in this test I'm counting how many times `executeFilterSearch` is called and in the other one it's checking what it's been called with, and so on. The rendering is also a bit different in each; in this one I need an external div to click out into. lmk if there's a specific block of code you think I can put into a function- I don't really see one though lol

oshi97 (approved):

PR #283, opened 2022-08-05
code coverage percentages job on pull_request event

yen-tt:
Add/update comment for code coverage percentages on pull_request event only.

TEST=manual

made a non dev/ test branch and push the same changes to it. see that the ` individual_coverage_percentages` job was cancelled.
https://github.com/yext/search-ui-react/runs/7694568793?check_suite_focus=true
https://github.com/yext/search-ui-react/tree/test-coverage-githubaction
Screen Shot 2022-08-05 at 11 49 18 AM


oshi97 (approved):
At first I was thinking it could be cool if we had 3 different code coverage badges on our README but that would probably be more confusing than anything else.

PR #284, opened 2022-08-05
extract FilterDivider from facet components

juliannzhou:
This PR extracts FilterDivider function from HierarchicalFacets, NumericalFacets, and StandardFacets to be a standalone component.

J=SLAP-2294
TEST=manual

Test: the divider renders correctly on test site on default and when inputing a custom divider name.


src/components/Filters/FilterDivider.tsx
```diff
@@ -0,0 +1,3 @@
+export function FilterDivider({ className ='w-full h-px bg-gray-200 my-4' }: { className?: string }): JSX.Element {
```
yen-tt, line 1: can we add docs to this? make sure to mark it as `@public`

src/components/Filters/FilterDivider.tsx
```diff
@@ -0,0 +1,3 @@
...
+  return <div className={className} />;
+}
```
yen-tt, line 3: also need to export the component

src/components/Filters/FilterDivider.tsx
```diff
@@ -0,0 +1,12 @@
+/**
+ * A component that displays numerical facets applicable to the current vertical search.
+ *
+ * @param props - A customizable className for facet divider
+ * @returns A React component for facets
+ *
+ * @public
+ */
+
+export function FilterDivider({ className ='w-full h-px bg-gray-200 my-4' }: { className?: string }): JSX.Element {
```
nmanu1, line 10: should we be using `tw-merge` for the `className` to allow deep merging? it might be a little confusing if users specify it with the `customCssClasses` of a component like `StandardFacets` where the other classes would be merged, but this one would be replaced

src/components/Filters/index.ts
```diff
@@ -2,6 +2,7 @@ export { CheckboxOption, CheckboxOptionProps, CheckboxCssClasses, FilterOptionCo
 export { CollapsibleLabel, CollapsibleLabelProps } from './CollapsibleLabel';
 export { CollapsibleSection, CollapsibleSectionProps } from './CollapsibleSection';
 export { FacetsProvider, FacetsProviderProps } from './FacetsProvider';
+export { FilterDivider } from './FilterDivider';
```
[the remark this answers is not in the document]
yen-tt: we do import certain things from this folder (`{...} from './Filters';` in components/index.ts) so you could add FilterDivider to that list. (There isn't a clear distinction of what would be part of `Filters/` or top level `components/` so I think you could argue for either way..) I don't have a preference where it should go, but make sure it is being exported by using in on a page in the test-site!

src/components/FilterDivider.tsx
```diff
@@ -0,0 +1,12 @@
...
+/**
+ * A divider component for NumericalFacets, HierarchicalFacets, and StaticFacets.
+ *
+ * @param props - A customizable className for facet divider
+ * @returns A React component for facets
+ *
+ * @public
+ */
+export function FilterDivider({ className }: { className?: string }): JSX.Element {
+  return <div className={twMerge(className, 'w-full h-px bg-gray-200 my-4')} />;
```
oshi97, line 11: should the twMerge be in the other order? I'm pretty sure we want the user defined className to have priority

src/components/FilterDivider.tsx
```diff
@@ -0,0 +1,12 @@
+import { twMerge } from "../hooks/useComposedCssClasses";
+/**
+ * A divider component for NumericalFacets, HierarchicalFacets, and StaticFacets.
```
nmanu1, line 3: this is used in `StandardFacets` (not `StaticFacets`) and should be placed between `StaticFilters` components

src/components/FilterDivider.tsx
```diff
@@ -0,0 +1,12 @@
...
+ *
+ * @param props - A customizable className for facet divider
+ * @returns A React component for facets
```
nmanu1, line 6: needs to be updated too

src/components/FilterDivider.tsx
```diff
@@ -0,0 +1,12 @@
+import { twMerge } from "../hooks/useComposedCssClasses";
+/**
+ * A divider component for NumericalFacets, HierarchicalFacets, and StaticFacets.
+ *
+ * @param props - A customizable className for facet divider
```
nmanu1, line 5: nit: can this be "filter divider" instead of "facet divider"?

nmanu1 (commented):
there are a couple linting warnings to address

src/components/FilterDivider.tsx
```diff
@@ -0,0 +1,12 @@
+import { twMerge } from '../hooks/useComposedCssClasses';
+/**
+ * A divider component for NumericalFacets, HierarchicalFacets, and StandardFacets.
```
nmanu1, line 3: can you also add that this divider should be used in between `StaticFilters` components to separate them?

yen-tt (approved):

PR #287, opened 2022-08-09
Exclude `src/models` from Jest Coverage Collection.

tmeyer2115:
This directory contains types and interfaces outlining data models.
There are no function or anything of the like to unit test.

TEST=auto

Re-ran the `test` command and saw this directory excluded from the
coverage collection.

### DIFF
--- a/package.json
+++ b/package.json
@@ -99,7 +99,8 @@
     "verbose": true,
     "collectCoverage": false,
     "collectCoverageFrom": [
-      "src/**"
+      "src/**",
+      "!src/models/**"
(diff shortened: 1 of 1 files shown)


yen-tt (approved):
I think we do have some type guard functions in some files in /models such as `isNumberRangeFilter` and `isVerticalLink`, which is placed there with the corresponding interfaces since any changes to the interface would directly affect those type guard function. But since they are not super complicated or is something we would have explicit unit tests on, it should be fine to exclude!

PR #288, opened 2022-08-09
Added Dropdown Stories

ElemelonWind:
Added stories for `Dropdown` component. Though it is used in `FilterSearch` and `SearchBar`, we figured it'd be good to have separate stories because users should be able to use `Dropdown` for their own components. `Dropdown` doesn't have any built-in css, so I used the `FilterSearch` css as a reference for styling.

On writing the tests, I found a wcag bug with `DropdownInput` that wasn't setting the `aria-activedescendant` correctly on select. I had to set `focusedIndex` back to `-1` after performing actions related to pressing 'enter'. 

J=SLAP-2320
TEST=auto

### DIFF
--- a/.storybook/preview.js
+++ b/.storybook/preview.js
@@ -34,7 +34,8 @@ export const parameters = {
         'AlternativeVerticals',
         'SpellCheck',
         'ResultsCount',
-        'LocationBias'
+        'LocationBias',
+        'Dropdown'
       ]
     }
   }
--- a/tests/components/Dropdown.stories.tsx
+++ b/tests/components/Dropdown.stories.tsx
@@ -45,12 +45,37 @@ DropdownExpanded.play = ({ canvasElement }) => {
   userEvent.click(textboxEl);
 };
 
+export const AlwaysSelectExpanded = (args: DropdownProps) => {
+  return (
+    <div className={cssClasses.filterSearchContainer}>
+      <Dropdown {...args} alwaysSelectOption={true}>
+        <DropdownInput className={cssClasses.inputElement} />
+        <DropdownMenu>
+          <div className='absolute z-10 w-full shadow-lg rounded-md border border-gray-300 bg-white pt-2 pb-2 mt-1'>
+            <DropdownItem focusedClassName={cssClasses.focusedOption} className={cssClasses.option} value='item1'>
+              item1
+            </DropdownItem>
+            <DropdownItem focusedClassName={cssClasses.focusedOption} className={cssClasses.option} value='item2'>
+              item2
+            </DropdownItem>
+          </div>
+        </DropdownMenu>
+      </Dropdown>
+    </div>
+  );
+};
+AlwaysSelectExpanded.play = ({ canvasElement }) => {
+  const canvas = within(canvasElement);
+  const textboxEl = canvas.getByRole('textbox');
+  userEvent.click(textboxEl);
+};
+
 export const DropdownHighlighted = Primary.bind({});
 DropdownHighlighted.play = ({ canvasElement }) => {
   const canvas = within(canvasElement);
   const textboxEl = canvas.getByRole('textbox');
   userEvent.click(textboxEl);
-  userEvent.keyboard('{arrowdown}');
+  userEvent.keyboard('{arrowdown}{arrowdown}');
 };
 
 export const DropdownSelected = Primary.bind({});
@@ -58,6 +83,6 @@ DropdownSelected.play = ({ canvasElement }) => {
   const canvas = within(canvasElement);
   const textboxEl = canvas.getByRole('textbox');
   userEvent.click(textboxEl);
-  userEvent.keyboard('{arrowdown}');
+  userEvent.keyboard('{arrowdown}{arrowdown}');
   userEvent.keyboard('{enter}');
 };
--- a/tests/components/Dropdown.stories.tsx
+++ b/tests/components/Dropdown.stories.tsx
@@ -18,17 +18,22 @@ const cssClasses = {
   option: 'text-sm text-neutral-dark py-1 cursor-pointer hover:bg-gray-100 px-4'
 };
 
+const dropdownItemProps = {
+  focusedClassName: cssClasses.focusedOption,
+  className: cssClasses.option
+};
+
 export const Primary = (args: DropdownProps) => {
   return (
     <div className={cssClasses.filterSearchContainer}>
-      <Dropdown {...args}>
+      <Dropdown {...args} screenReaderText='screen reader text here'>
         <DropdownInput className={cssClasses.inputElement} />
         <DropdownMenu>
           <div className='absolute z-10 w-full shadow-lg rounded-md border border-gray-300 bg-white pt-2 pb-2 mt-1'>
-            <DropdownItem focusedClassName={cssClasses.focusedOption} className={cssClasses.option} value='item1'>
+            <DropdownItem {...dropdownItemProps} value='item1'>
               item1
             </DropdownItem>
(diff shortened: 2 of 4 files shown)


ElemelonWind:
Wasn't sure where to put the story in the order (in `preview.js`) - any suggestions appreciated!

nmanu1 (commented):
looks like there's a regression on one of the `SearchBar` stories where the button for clearing the search input is missing. also, on mobile with Safari, the search input's height has changed:
Screen Shot 2022-08-10 at 11 59 51 AM

Screen Shot 2022-08-10 at 12 00 20 PM

ElemelonWind:
> Details

Yeah- I fixed it so we clear the `SearchBar` before arrowing down, which means the clear input button isn't there anymore.

tests/components/Dropdown.stories.tsx
```diff
@@ -0,0 +1,93 @@
...
+};
+
+const dropdownItemProps = {
+  focusedClassName: cssClasses.focusedOption,
+  className: cssClasses.option
+};
+
+export const Primary = (args: DropdownProps) => {
+  return (
+    <div className={cssClasses.filterSearchContainer}>
```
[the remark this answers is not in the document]
ElemelonWind: the `Dropdown` in AlwaysSelectExpanded has an extra prop `alwaysSelectOption=true`. is it still possible to reuse it?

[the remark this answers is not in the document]
nmanu1: yeah, you can take in `alwaysSelectOption` as a param to a function that returns the component with that value dynamically populated

tests/components/Dropdown.stories.tsx
```diff
@@ -0,0 +1,93 @@
...
+      </Dropdown>
+    </div>
+  );
+};
+AlwaysSelectExpanded.play = ({ canvasElement }) => {
+  const canvas = within(canvasElement);
+  const textboxEl = canvas.getByRole('textbox');
+  userEvent.click(textboxEl);
+};
+
```
juliannzhou, line 77: Is `AlwaysSelectedEpanded` expected to have the first option selected when clicking on the textbox without any inputs? Since I remember in filterSearch it will select an option only after you input something.

tests/components/Dropdown.stories.tsx
```diff
@@ -0,0 +1,93 @@
+import { ComponentMeta } from '@storybook/react';
+import { within, userEvent } from '@storybook/testing-library';
+import { DropdownItem } from '../../src/components';
+import { Dropdown, DropdownProps } from '../../src/components/Dropdown/Dropdown';
+import { DropdownInput } from '../../src/components/Dropdown/DropdownInput';
+import { DropdownMenu } from '../../src/components/Dropdown/DropdownMenu';
+
+const meta: ComponentMeta<typeof Dropdown> = {
+  title: 'Dropdown',
+  component: Dropdown,
```
juliannzhou, line 10: nit: extra comma

tests/components/Dropdown.stories.tsx
ElemelonWind: yep! in filterSearch it only selects after inputting because there aren't any autocomplete options if the input is blank

tests/components/Dropdown.stories.tsx
```diff
@@ -0,0 +1,93 @@
...
+DropdownHighlighted.play = ({ canvasElement }) => {
+  const canvas = within(canvasElement);
+  const textboxEl = canvas.getByRole('textbox');
+  userEvent.click(textboxEl);
+  userEvent.keyboard('{arrowdown}{arrowdown}');
+};
+
+export const DropdownSelected = Primary.bind({});
+DropdownSelected.play = ({ canvasElement }) => {
+  const canvas = within(canvasElement);
```
nmanu1, line 88: could we factor out the 3 lines here for clicking the textbox (88-90) into a separate function? because it looks like every play function repeats these actions

tests/components/Dropdown.stories.tsx
```diff
@@ -43,11 +43,15 @@ export const Primary = (args: DropdownProps) => {
   );
 };
 
-export const DropdownExpanded = Primary.bind({});
-DropdownExpanded.play = ({ canvasElement }) => {
+function clickTextbox(canvasElement) {
```
nmanu1, line 46: nit: can this function be moved to the bottom of the file instead of having it split up the stories?

tests/components/Dropdown.stories.tsx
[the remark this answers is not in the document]
nmanu1: I think we do something similar in the `FilterSearch` and `RangeInput` Jest tests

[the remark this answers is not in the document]
ElemelonWind: whoops i missed this- will do!

tests/components/Dropdown.stories.tsx
```diff
@@ -0,0 +1,76 @@
...
+    </div>
+  );
+};
+
+export const DropdownExpanded = Primary.bind({});
+DropdownExpanded.play = ({ canvasElement }) => {
+  clickTextbox(canvasElement);
+};
+
+export const AlwaysSelectExpanded = Primary.bind({});
```
nmanu1, line 51: could we move this story to be the last one? the others use exactly the same component, so it makes more sense to me to have them grouped together before having one with different props

.storybook/preview.js
```diff
@@ -34,7 +34,8 @@ export const parameters = {
         'AlternativeVerticals',
         'SpellCheck',
         'ResultsCount',
-        'LocationBias'
+        'LocationBias',
+        'Dropdown'
```
nmanu1, line 38: I think `Dropdown` could go higher in this order. but, it might be good to ask product where they would like it placed because I think they've structured their documentation to follow the ordering here

ElemelonWind: Asked Tom about it; he said we should keep it at the bottom since no one uses it outside of `FilterSearch` and `SearchBar` yet

nmanu1:
> Coverage decreased (-0.03%) to 85.502% when pulling **[2dcb8b4](https://github.com/yext/search-ui-react/commit/2dcb8b476206542fd3f441e083b34bcb97d96b7d) on dev/dropdown-storybook** into **[14da812](https://github.com/yext/search-ui-react/commit/14da8129f69fe0e9f23b20f1e8cbeb0c9954421e) on main**.

how come the coverage has decreased? I thought we were reporting the combined coverage to Coveralls, which should increase by adding stories? or is this just the unit test coverage?

ElemelonWind:
> how come the coverage has decreased? I thought we were reporting the combined coverage to Coveralls, which should increase by adding stories? or is this just the unit test coverage?

coveralls only gets the unit test coverage! it decreased because I added that one line in `DropdownInput` to fix the failing wcag test- Tom said not to worry about writing a new jest for it

tests/components/Dropdown.stories.tsx
```diff
@@ -0,0 +1,76 @@
...
+    </div>
+  );
+};
+
+export const DropdownExpanded = Primary.bind({});
+DropdownExpanded.play = ({ canvasElement }) => {
+  clickTextbox(canvasElement);
+};
+
+export const DropdownHighlighted = Primary.bind({});
```
nmanu1, line 51: do we need this story? `AlwaysSelectExpanded` already shows what a highlighted dropdown option looks like

ElemelonWind: ah true, i'll delete

oshi97 (approved):

PR #289, opened 2022-08-10
Add Jest Tests for the Search Operations utilities.

tmeyer2115:
No test was written for `getSearchIntents` as that's essentially a pass through to `executeAutocomplete`.

J=SLAP-2321
TEST=auto

Ensured all unit tests passed.

### DIFF
--- a/tests/operations/search-operations.test.ts
+++ b/tests/operations/search-operations.test.ts
@@ -1,0 +1,41 @@
+import { SearchHeadless } from '@yext/search-headless-react';
+import { executeAutocomplete, executeSearch } from '../../src/utils/search-operations';
+import { generateMockedHeadless } from '../__fixtures__/search-headless';
+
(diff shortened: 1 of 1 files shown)


oshi97 (approved):

tests/operations/search-operations.test.ts
```diff
@@ -0,0 +1,41 @@
+import { SearchHeadless } from '@yext/search-headless-react';
```
nmanu1, line 1: do we want the file structure for the tests to mimic the `src` directory? in which case, this should go under `tests/utils/search-operations.test.ts`

tmeyer2115: So, we already have a `__utils__` directory already under `tests`. I think having both a `utils` and `__utils__` would be confusing. That's why I chose `operations`.

PR #290, opened 2022-08-10
Added location-operations jests

ElemelonWind:
Wrote a few jests to test the call of `getUserLocation` in different cases. Also fixed a small typo in `location-operations` docs.

J=SLAP-2322
TEST=auto

### DIFF
--- a/docs/search-ui-react.getuserlocation.md
+++ b/docs/search-ui-react.getuserlocation.md
@@ -4,7 +4,7 @@
 
 ## getUserLocation() function
 
-Retrieves user's location using nagivator.geolocation API.
+Retrieves user's location using navigator.geolocation API.
 
 <b>Signature:</b>
 
--- a/tests/operations/location-operations.test.tsx
+++ b/tests/operations/location-operations.test.tsx
@@ -1,0 +1,29 @@
+import { SearchIntent, useSearchActions } from '@yext/search-headless-react';
+import { getUserLocation, updateLocationIfNeeded } from '../../src/utils/location-operations';
+
+describe('Location Operations updateLocationIfNeeded', () => {
+  it('fires getUserLocation with NearMe intent', async () => {
+    const searchActions = useSearchActions();
+    const intents: SearchIntent[] = [SearchIntent.NearMe];
+    await updateLocationIfNeeded(searchActions, intents);
+    expect(getUserLocation).toHaveBeenCalled();
+  });
+
+  it('does not fire getUserLocation without NearMe intent', async () => {
+    const searchActions = useSearchActions();
+    const intents: SearchIntent[] = [];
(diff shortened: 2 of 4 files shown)


docs/search-ui-react.getuserlocation.md
```diff
@@ -4,7 +4,7 @@
 
 ## getUserLocation() function
 
-Retrieves user's location using nagivator.geolocation API.
```
oshi97, line 7: nagivator

tests/operations/location-operations.test.tsx
```diff
@@ -0,0 +1,29 @@
+import { SearchIntent, useSearchActions } from '@yext/search-headless-react';
+import { getUserLocation, updateLocationIfNeeded } from '../../src/utils/location-operations';
+
+describe('Location Operations updateLocationIfNeeded', () => {
+  it('fires getUserLocation with NearMe intent', async () => {
+    const searchActions = useSearchActions();
```
oshi97, line 6: how is useSearchActions working outside of a react component here?

ElemelonWind: uHHh honestly I'm not sure why it worked lmao

how would you suggest approaching this instead?

oshi97: Would you mind digging a little into what's happening right now? If it works as is then that's good

ElemelonWind:
Will probably add a test for second part of if statement (`!searchActions.state.location.userLocation`) tmrw

tests/operations/location-operations.test.tsx
```diff
@@ -0,0 +1,37 @@
+import { SearchIntent } from '@yext/search-headless-react';
+import { generateMockedHeadless } from '../__fixtures__/search-headless';
+import { updateLocationIfNeeded } from '../__fixtures__/utils/location-operations';
+
+describe('Location Operations updateLocationIfNeeded', () => {
+
+  it('fires setUserLocation with NearMe intent', async () => {
+    const searchActions = generateMockedHeadless();
```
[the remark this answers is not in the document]
ElemelonWind: I referenced Tom's [search-operations PR](https://github.com/yext/search-ui-react/commit/6e2d2f39930cf517484cebe7f772445b5885db2b) bc I figured it'd be the most similar to locations-operations. I don't think it makes a difference in this case?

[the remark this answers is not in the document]
oshi97: Their purposes are a little different, `generateMockedHeadless()` makes a real AnswersHeadless instance using a mocked search-core (it doesn't make real requests) while the mock/spy methods are more lightweight. I think what we have here is fine!

oshi97 (approved):

PR #291, opened 2022-08-10
update visual test script to kill process

juliannzhou:
This PR updates visual-coverage and start-storybook test script to only kill any process started by the test at the end of the test. If there are pre-existing process on port, the script assumes the process is a storybook instance and use it for test-runner coverage instead of starting a new storybook, and leaves the pre-existing process running after the test.

J=SLAP-2323
TEST=manual

Test: tested by running `npm run test:visual` without preexisting process on port, the test runs as expected and the storybook process is killed at the end of the test; also ran the visual test with a preexisting storybook process on port, the test-runner uses the existing storybook for coverage report and leaves the storybook process running after the test as expected.


tests/scripts/start-storybook.sh
```diff
@@ -1,5 +1,6 @@
 #!/bin/bash
 
+npx kill-port 6006
```
oshi97, line 3: was the issue that both visual-coverage.sh and start-storybook.sh both started up storybook so they would port clobber each other?

tests/scripts/start-storybook.sh
```diff
@@ -1,5 +1,6 @@
...
 npm run storybook -- --quiet --ci &
```
oshi97, line 4: is there a start-storybook cli flag we can use to enforce the port number and fail if the port is already in use? for instance vite has https://vitejs.dev/config/server-options.html#server-strictport . 
I believe this is the storybook cli docs? https://storybook.js.org/docs/react/api/cli-options

tests/scripts/start-storybook.sh
juliannzhou: Since in visual-coverage.sh it is running the script ./start-storybook.sh, I don't think storybook is started two times.

oshi97: Oh gotcha, how did this issue come up? Was it during local dev?

nmanu1: yeah this is something that happened when I had storybook running locally and then also ran the test command locally

tests/scripts/start-storybook.sh
juliannzhou: Do we want the start-storybook to fail if the port is in use or kill whatever process that's running to start the test? I don't think there is a cli option in the storybook docs that does this.

oshi97: hmm I think having storybook fail instead of kill whatever was already on 6006 might be better? killing a preexisting process seems a little aggressive

tests/scripts/start-storybook.sh
oshi97: I feel like in that case the right behavior would be to allow the second storybook to start in a different port, and use that for the test command? Alternatively we could say that if something is already running on port 6006, assume that it's a preeixsting storybook server and use that instead of starting a new one?

tests/scripts/start-storybook.sh
juliannzhou: Okay I updated the test script to make the test fail and asking the user to kill process instead of killing the process automatically

tests/scripts/visual-coverage.sh
```diff
@@ -1,16 +1,24 @@
 #!/bin/bash
 
-# start Storybook in the same shell
-. ./tests/scripts/start-storybook.sh
+#check for existing process on port:6006
+process=`lsof -i :6006`
```
oshi97, line 4: Instead of guarding this script by whether 6006 is available, can we tweak start-storybook.sh to be a no-op if 6006 is not available? That way if storybook is already running we don't kill it and restart unnecessarily

tests/scripts/visual-coverage.sh
```diff
@@ -1,16 +1,24 @@
...
 
-# generate coverage from storybook test-runner
-test-storybook --coverage
+  # generate coverage from storybook test-runner
+  test-storybook --coverage
 
-# kill the locally served Storybook
-kill -9 $JOB_ID
+  # kill the locally served Storybook
+  npx kill-port 6006
```
oshi97, line 15: instead of killing a hardcoded port of 6006, can we instead of start-storybook.sh return the job id (process id? I don't know the terminology) of the storybook instance, and kill that process here instead? that way we don't have a hardcoded port of 6006

tests/scripts/visual-coverage.sh
juliannzhou: Is there a case where a process is running on port 6006 but it's not storybook? If there is, I am concerned about the `test-storybook --coverage` step in visual test script erroring out.

juliannzhou: Also to confirm, if there is already a storybook instance running on port 6006, we'd start a visual test using that process. Do we need to kill it after we run the test or leave it as it since it's running before we've started the test?

oshi97: I wouldn't worry about a case where something is running on 6006 but it's not storybook. Since this item was for local dev QOL imo it's not a big deal if we false start.

I think ideally, if storybook was started before the test we would leave it, and if storybook was started solely for the purposes of visual-coverage.sh we would kill it. I would do something like "if start-storybook is a no-op return a blank JOB_ID, and if JOB_ID is blank don't kill it". I'm not exactly sure how returning stuff works for bash scripts.

nmanu1: > I'm not exactly sure how returning stuff works for bash scripts.

because we run the start-storybook script in the same shell the variables persist between the scripts. so defining JOB_ID in one will let us use/modify it in the other

juliannzhou: > I wouldn't worry about a case where something is running on 6006 but it's not storybook. Since this item was for local dev QOL imo it's not a big deal if we false start.
> 
> I think ideally, if storybook was started before the test we would leave it, and if storybook was started solely for the purposes of visual-coverage.sh we would kill it. I would do something like "if start-storybook is a no-op return a blank JOB_ID, and if JOB_ID is blank don't kill it". I'm not exactly sure how returning stuff works for bash scripts.

I modified the behavior to kill storybook only when it's started by the test. So the current behavior is to assume existing process on port is storybook and start test using the storybook instance, then leave the storybook running if it's a preexisting process.

juliannzhou: It turns out that the storybook processes running are the child processes for the original process with the JOB_ID, so we have to kill them along with the parent process.

nmanu1 (commented):
can you update the PR description? also, do you know why the coverage decreased?

juliannzhou:
> can you update the PR description? also, do you know why the coverage decreased?

I think the coverage decrease is caused by recent changes to main by the dropdown storybook pr, since the base_coverage action is pulling from main. The three coverage numbers reported by github-actions bot increased instead of decreased.

tests/scripts/start-storybook.sh
```diff
@@ -1,21 +1,28 @@
 #!/bin/bash
 
-npm run storybook -- --quiet --ci &
-JOB_ID=$(echo $!) #get the background job ID
+#check for existing process on port:6006
+process=`lsof -i :6006`
+if [[ -z $process ]]
```
oshi97, line 5: This is probably a dumb question but what is the `-z` here?

tests/scripts/visual-coverage.sh
```diff
@@ -6,9 +6,11 @@
 # generate coverage from storybook test-runner
 test-storybook --coverage
 
-# kill the locally served Storybook
-kill -9 $JOB_ID
-
+# kill the locally served Storybook started by the test
+if [[ -z $process ]]
```
oshi97, line 10: will this always kill the local storybook, even if it wasn't started as part of this script?

tests/scripts/start-storybook.sh
juliannzhou: This is to check if the variable is undefined, and returns true here if there is nothing returned with the `lsof -i :6006` command.

tests/scripts/visual-coverage.sh
juliannzhou: This will kill the local storybook only if it is started by the test and will leave the storybook started before the test running.

oshi97 (approved):
nice!

tests/scripts/start-storybook.sh
```diff
@@ -1,21 +1,28 @@
 #!/bin/bash
 
-npm run storybook -- --quiet --ci &
-JOB_ID=$(echo $!) #get the background job ID
+#check for existing process on port:6006
+process=`lsof -i :6006`
```
oshi97, line 4: wait actually sorry one last thing, can we rename this variable? process is a little vague

oshi97 (changes requested):

tests/scripts/start-storybook.sh
oshi97: maybe preexistingPortProcess? I don't love the name I came up with so open to suggestions!

tests/scripts/start-storybook.sh
```diff
@@ -1,21 +1,28 @@
...
+if [[ -z $process ]]
+then
+  echo "port:6006 available - starting storybook"
+  npm run storybook -- --quiet --ci &
+  JOB_ID=$(echo $!)
```
oshi97, line 9: instead of JOB_ID maybe NEW_STORYBOOK_JOB_ID?

tests/scripts/start-storybook.sh
juliannzhou: sounds good to me!

nmanu1 (commented):
> Test: tested by starting a process on port 6006 and running `npm run test:visual`, the process is killed on the port before starting storybook and the storybook process is killed at the end of the test.

could you also update the description for how you tested?

oshi97 (approved):
🔥 🔥 🔥

PR #293, opened 2022-08-12
fix node v18 visual test issue

juliannzhou:
This PR fixes the visual coverage test in node v18, as Run Tests / call_run_tests-react-16 / tests (18.x) from github actions in the past pull requests have been failing. It is a temporary fix to allow the visual test in node 18 to run. It is not a recommended approach to use node option with OpenSSL 3 in the legacy mode, as it would be running with known insecure algorithms. But for testing purpose, there would not be security issue. This PR only sets node option if the node version is 17+. Since issue is caused by storybook using a lower version of its dependency webpack (with newest version supporting node 18), it will be fixed if storybook bump their webpack version.

TEST=auto


package.json
```diff
@@ -43,7 +43,7 @@
     "wcag": "test-storybook",
     "test:unit": "jest --coverage --coverageDirectory=coverage/unit",
     "test:visual": "./tests/scripts/visual-coverage.sh",
-    "test": "./tests/scripts/combined-coverage.sh"
+    "test": "unset NODE_OPTIONS && export NODE_OPTIONS=--openssl-legacy-provider && ./tests/scripts/combined-coverage.sh"
```
oshi97, line 46: what is the issue with node 18?

oshi97: What command in the script has the issue?

juliannzhou: The command that has issue is `npm run test:visual`. So for test 18x in the past pull requests, the visual coverage test has always been skipped due to storybook not starting. It is due to the webpack dependency storybook uses not not supporting with node version 17 and up
Screen Shot 2022-08-12 at 3 15 38 PM
.

tests/scripts/combined-coverage.sh
```diff
@@ -1,6 +1,13 @@
 #!/bin/bash
 
 npm run test:unit
+CURRENT_NODE_VERSION=`node -v`
+if [[ $CURRENT_NODE_VERSION =~ 17(.*)|18(.*) ]]
+then
+  echo -e "currently using node version 17+ - setting NODE_OPTIONS \n"
+  unset NODE_OPTIONS
```
oshi97, line 8: out of curiosity why do we need to unset NODE_OPTIONS? We set NODE_ICU_DATA in https://github.com/yext/answers-hitchhiker-theme/blob/master/package.json#L16 without needing to unset it

juliannzhou: There was an error `node: --openssl-legacy-provider is not allowed in NODE_OPTIONS` when using node v14 and I tried to resolve it with unset node option. I will remove it.

tests/scripts/combined-coverage.sh
```diff
@@ -1,6 +1,12 @@
 #!/bin/bash
 
 npm run test:unit
+CURRENT_NODE_VERSION=`node -v`
+if [[ $CURRENT_NODE_VERSION =~ 17(.*)|18(.*) ]]
+then
+  echo -e "currently using node version 17+ - setting NODE_OPTIONS \n"
+  export NODE_OPTIONS=--openssl-legacy-provider
```
oshi97, line 8: can we do this within the github workflow instead? https://github.com/yext/search-ui-react/blob/dev/v18-visual-test/.github/workflows/run-tests.yml

oshi97: Ideally in the workflows we would only set this in the node 18 jobs just like you have here. We would probably need to separate out 18.x from the specified node matrices

juliannzhou: Should I still keep the changes in the test script? I was thinking keeping them would allow visual test to run locally with node 17/18

oshi97: oh I see, that's fair!

oshi97: Could you go back to what you had before, but instead of putting it in combined-coverage.sh make a new script for getting the visual coverage? otherwise we won't be able to run `npm run test:visual` on its own

a separate question, if you export NODE_OPTIONS in this script, does it persist outside of the script? if it does, I think we want to specify it inline, something like `NODE_OPTIONS=blah npx jest ... `

juliannzhou: I tested it in node 16 with a test script to set option since the test would error out of NODE_OPTION is set to --openssl-legacy-provider. But after running set NODE_OPTION bash script, the test still runs. So I think setting the NODE_OPTION only works within the script. I added an unset NODE_OPTION at the end of the visual test script just to be safe but it's most likely not necessary.

.github/workflows/run-tests.yml
```diff
@@ -1,3 +1,5 @@
+
```
oshi97, line 1: extra lines

tests/scripts/visual-coverage.sh
```diff
@@ -1,5 +1,12 @@
...
 
+# set NODE_OPTIONS for node 17/18
+if [[ $CURRENT_NODE_VERSION =~ 17(.*)|18(.*) ]]
+then
+  echo -e "currently using node version 17+ - setting NODE_OPTIONS \n"
+  export NODE_OPTIONS=--openssl-legacy-provider
+fi
+
 # start Storybook in the same shell
 . ./tests/scripts/start-storybook.sh
```
oshi97, line 11: Is the issue is in start-storybook which uses webpack?

.github/workflows/run-tests.yml
```diff
@@ -1,3 +1,4 @@
...
```
oshi97, line 1: there's still an extra line here 😢

tests/scripts/start-storybook.sh
```diff
@@ -23,6 +31,8 @@ then
     attempt_counter=$(($attempt_counter+1))
     sleep 1
   done
+  # unset NODE_OPTIONS
+  unset NODE_OPTIONS
```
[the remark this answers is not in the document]
juliannzhou: Okay I will make a note of it with Yen!

PR #295, opened 2022-08-17
Static filter state updates

nmanu1:
Update the components that interact with the static filters State to work with the changes in Headless. The `StaticFilters` and `AppliedFilters` components will only work with field value filters. If someone wants to create and display more complex static filters, they should fork the components and do that themselves.

Also, because the Headless-React alpha versions include the change in Core to make direct answers generic, I updated the component to manually cast those values as strings for now. That behavior should be changed with the other direct answer work.

J=SLAP-2327
TEST=manual

Spin up the test-site and see that static filters, range input filters, and filter search filters can be successfully applied and removed.

### DIFF
--- a/src/components/FilterSearch.tsx
+++ b/src/components/FilterSearch.tsx
@@ -1,9 +1,9 @@
-import { AutocompleteResult, Filter, FilterSearchResponse, SearchParameterField, useSearchActions, useSearchState } from '@yext/search-headless-react';
+import { AutocompleteResult, FieldValueFilter, FilterSearchResponse, SearchParameterField, useSearchActions, useSearchState } from '@yext/search-headless-react';
 import { useCallback, useEffect, useMemo, useState } from 'react';
 import { useComposedCssClasses } from '../hooks/useComposedCssClasses';
 import { useSynchronizedRequest } from '../hooks/useSynchronizedRequest';
 import { executeSearch } from '../utils';
-import { isDuplicateFilter } from '../utils/filterutils';
+import { getSelectableFieldValueFilters, isDuplicateFieldValueFilter } from '../utils/filterutils';
 import { Dropdown } from './Dropdown/Dropdown';
 import { DropdownInput } from './Dropdown/DropdownInput';
 import { DropdownItem } from './Dropdown/DropdownItem';
@@ -42,7 +42,7 @@ const builtInCssClasses: Readonly<FilterSearchCssClasses> = {
 export interface FilterSearchProps {
   /** An array of fieldApiName and entityType which indicates what to perform the filter search against. */
   searchFields: Omit<SearchParameterField, 'fetchEntities'>[],
-  /** The display label for the component. Defaults to "Filter". */
+  /** The display label for the component. */
   label?: string,
   /**
    * The search input's placeholder text when no text has been entered by the user.
@@ -78,9 +78,13 @@ export function FilterSearch({
     return { ...searchField, fetchEntities: false };
   });
   const cssClasses = useComposedCssClasses(builtInCssClasses, customCssClasses);
-  const [currentFilter, setCurrentFilter] = useState<Filter>();
+  const [currentFilter, setCurrentFilter] = useState<FieldValueFilter>();
   const [filterQuery, setFilterQuery] = useState<string>();
-  const filters = useSearchState(state => state.filters.static);
+  const staticFilters = useSearchState(state => state.filters.static);
+  const fieldValueFilters = useMemo(
+    () => getSelectableFieldValueFilters(staticFilters ?? []),
+    [staticFilters]
+  );
 
   const [
     filterSearchResponse,
@@ -95,12 +99,14 @@ export function FilterSearch({
   );
 
   useEffect(() => {
-    if (currentFilter && filters?.find(f => isDuplicateFilter(f, currentFilter) && !f.selected)) {
+    if (currentFilter && fieldValueFilters?.find(f =>
+      isDuplicateFieldValueFilter(f, currentFilter) && !f.selected
+    )) {
       clearFilterSearchResponse();
       setCurrentFilter(undefined);
       setFilterQuery('');
     }
-  }, [clearFilterSearchResponse, currentFilter, filters]);
+  }, [clearFilterSearchResponse, currentFilter, fieldValueFilters]);
 
   const sections = useMemo(() => {
     return filterSearchResponse?.sections.filter(section => section.results.length > 0) ?? [];
@@ -109,14 +115,21 @@ export function FilterSearch({
   const hasResults = sections.flatMap(s => s.results).length > 0;
 
   const handleDropdownEvent = useCallback((value, itemData, select) => {
-    const newFilter = itemData?.filter as Filter;
+    const newFilter = itemData?.filter as FieldValueFilter;
     const newDisplayName = itemData?.displayName as string;
     if (newFilter && newDisplayName) {
       if (select) {
         if (currentFilter) {
-          searchActions.setFilterOption({ ...currentFilter, selected: false });
+          searchActions.setFilterOption({
+            filter: { ...currentFilter, kind: 'fieldValue' },
+            selected: false
+          });
         }
-        searchActions.setFilterOption({ ...newFilter, displayName: newDisplayName, selected: true });
+        searchActions.setFilterOption({
+          filter: { ...newFilter, kind: 'fieldValue' },
+          displayName: newDisplayName,
+          selected: true
+        });
         setCurrentFilter(newFilter);
         setFilterQuery(newDisplayName);
         executeFilterSearch(newDisplayName);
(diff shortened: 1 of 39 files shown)


tests/components/AppliedFilters.test.tsx
```diff
@@ -84,7 +93,8 @@ describe('AppliedFilters', () => {
 
   it('Static filters are rendered', () => {
     render(<AppliedFilters />);
-    const staticFilterDisplayName = mockedState.filters.static[0].value as string;
+    const staticFilter = mockedState.filters.static[0].filter as FieldValueStaticFilter;
```
oshi97, line 96: we can't do `const staticFilter: FieldValueStaticFilter` instead of the assert?

src/utils/filterutils.tsx
```diff
@@ -119,3 +124,25 @@ export function getSelectedNumericalFacetFields(searchActions: SearchActions): S
...
+ * Goes through the provided static filters and returns all that are field value
+ * filters.
+ *
+ * @param staticFilters - The list of {@link SelectableStaticFilter}s to search through
+ * @returns All filters that are {@link SelectableFieldValueFilter}s
+ */
+export function getSelectableFieldValueFilters(
+  staticFilters: SelectableStaticFilter[]
+): SelectableFieldValueFilter[] {
+  return staticFilters.reduce((fieldValueFilters: SelectableFieldValueFilter[], s) => {
```
oshi97, line 138: Can we use `filter` instead of `reduce` here?

src/components/Filters/CheckboxOption.tsx
```diff
@@ -97,14 +97,14 @@ export function CheckboxOption(props: CheckboxOptionProps): JSX.Element | null {
     handleClick(evt.target.checked);
   }, [handleClick]);
 
-  const optionFilter: Filter = useMemo(() => {
+  const optionFilter: FieldValueFilter = useMemo(() => {
```
oshi97, line 100: Is this a FieldValueFilter if it's missing the `kind: 'fieldValue'` property? I wonder if it's simpler to put the fieldValue here

src/components/FilterSearch.tsx
```diff
@@ -109,14 +115,21 @@ export function FilterSearch({
...
         if (currentFilter) {
-          searchActions.setFilterOption({ ...currentFilter, selected: false });
+          searchActions.setFilterOption({
+            filter: { ...currentFilter, kind: 'fieldValue' },
+            selected: false
+          });
         }
-        searchActions.setFilterOption({ ...newFilter, displayName: newDisplayName, selected: true });
+        searchActions.setFilterOption({
+          filter: { ...newFilter, kind: 'fieldValue' },
```
oshi97, line 129: instead of adding the fieldValue kind when we set the filter can we add it in itemDataMatrix? I think then we only have to add the `kind` in one place.

tests/components/AppliedFilters.test.tsx
nmanu1: we can't set the type like that because the static filters state could store other types of static filters too. but, I made some changes that removed the new asserts

src/utils/filterutils.tsx
nmanu1: I think `filter` would only work if we wanted to return a list of type `SelectableStaticFilter[]`. since we're parsing the static filters and returning a different type, especially one that's not one of the `StaticFilter` union types, I used `reduce`

src/components/FilterSearch.tsx
nmanu1: I chose to do it this way because it makes the comparisons between existing filters easier. it's harder to compare between `StaticFilter` types than it is between `FieldValueFilter`s, so keeping the `itemData` as a `FieldValueFilter` and adding the `kind` later seemed easier. also, I kind of like that it leaves most of our code untouched and is essentially just modifying the interactions with Headless (when we get the state and make updates to it). I feel like that would make it easier if later on we decided to make more changes related to static filters

src/utils/filterutils.tsx
oshi97: Does .filter combined with .map work? I don't think there's anything wrong with reduce it's just slightly harder to understand (because you can do more complicated things with it)

src/components/Filters/CheckboxOption.tsx
oshi97: Should we change the kind value to something other than `"fieldValue"` then? Seems a little confusing to me that a Filter with kind = fieldValue isn't a FieldValueFilter

src/components/FilterSearch.tsx
oshi97: I'm a little confused, doesn't this require us to change more code? If we add it to the itemMatrix that's one line to change and will forever be one line changed. With the current way don't we have to tack on kind: 'fieldValue' every time we call setFilterOption? (currently two times).

In terms of comparison, if we use FieldValueStaticFilter instead of FieldValueFilter can't we use the exact same comparison methods?

src/components/Filters/CheckboxOption.tsx
nmanu1: I'm not sure I follow. the "Filter" model no longer exists because it was renamed to `FieldValueFilter` (which has the same `fieldId`, `matcher`, and `value`). now, if you want to specify a static filter with a `kind` of fieldValue, that would be a `FieldValueStaticFilter`

oshi97: Can we change the "fieldValue" kind to "fieldValueStaticFilter" or something that makes it clear the associated model isn't `FieldValueFilter` but `FieldValueStaticFilter`? I didn't notice this until seeing the models used in the component lib. Seeing `kind: 'fieldValue'` makes me think "oh this object is a FieldValueFilter" when in reality it's a FieldValueStaticFilter.

oshi97: Ideally I think it would be good if we came up for different names so that it's easier to differentiate FieldValueFilter and FieldValueStaticFilter but I don't have any ideas and this could be a "me" thing

src/utils/filterutils.tsx
nmanu1: splitting the `reduce` into a `filter` and then a `map` would have type errors because `filter` still returns a type of `SelectableStaticFilter[]`, and `map` wouldn't know that it technically only has field value static filters in it after the filtering. so we would have to repeat the type checking in the `map` to access the `fieldId`, `matcher`, and `value` properties. but, we would still have to handle the case where TS thinks it might not be a field value static filter. essentially, we would have to either do a manual assertion, or allow `undefined` to be returned in the array like this:

Screen Shot 2022-08-17 at 2 05 10 PM

src/components/FilterSearch.tsx
nmanu1: actually you're right, `FieldValueStaticFilter` and `FieldValueFilter` can be easily compared. we'd only need to change the comparison if it were `StaticFilter` being compared to `FieldValueFilter`. updated to move the `kind` value into `itemDataMatrix`!

src/components/Filters/CheckboxOption.tsx
nmanu1: right now, the `kind` values are consistent so that `FieldValueStaticFilter` has `kind: 'fieldValue'`, `ConjunctionStaticFilter` has `kind: 'conjunction'`, and `DisjunctionStaticFilter` has `kind: 'disjunction'`. would it be more clear if each `kind` was the full type (e.g. `kind = 'fieldValueStaticFilter' | 'conjunctionStaticFilter' | 'disjunctionStaticFilter'`)?

src/utils/filterutils.tsx
oshi97: hmm we could do this if you like? but I see your point, either works for me!
```ts

  return staticFilters.map(s => {
    const { filter: { kind, ...filterFields }, ...displayFields } = s;
    if (kind === 'fieldValue') {
      return {
        ...displayFields,
        ...filterFields
      };
    }
    return undefined;
  }).filter((s): s is SelectableFieldValueFilter => !!s);
```

src/components/Filters/CheckboxOption.tsx
nmanu1: I was thinking about it more as what kind of static filter it is, rather than what the corresponding type is. So the other data that you'll be able to access for the filter would correspond to a 'fieldValue' or to a 'conjunction'

src/utils/filterutils.tsx
nmanu1: that works, updated!

src/components/Filters/CheckboxOption.tsx
oshi97: Ah I see. I think I can get used to the naming! I was trying to think to alternatives to `FieldValueFilter` but I could only come up with stuff like `ApiFilter` and `BaseFilter`

src/components/Filters/RangeInput.tsx
```diff
@@ -102,6 +102,10 @@ export function RangeInput(props: RangeInputProps): JSX.Element | null {
   const [minRangeInput, setMinRangeInput] = useState<string>('');
   const [maxRangeInput, setMaxRangeInput] = useState<string>('');
   const staticFilters = useSearchState(state => state.filters.static);
+  const fieldValueFilters = useMemo(
+    () => getSelectableFieldValueFilters(staticFilters ?? []),
+    [staticFilters]
+  );
   const isDisabled = filters.some(filter => filter.selected && filter.fieldId === fieldId);
 
   const rangeFilter: NumberRangeFilter = useMemo(() => {
```
oshi97, line 111: Would it be possible to also put `kind: 'fieldValue'` here?

nmanu1: updated!

oshi97 (approved):
lgtm!

PR #296, opened 2022-08-17
FieldValueDirectAnswer component

yen-tt:
Added FieldValueDirectAnswer to handle all the built in field types for field value direct answer. Note, we currently do not support internationalization so addresses will be written as if they are US addresses.

in nidhi's PR for featured snippet direct answer, she will implement the logic to handle rich text field type using react-markdown library. So, the switch case for rich_text in FieldValueDirectAnswer will be updated once that is done. Along with the tests for rich_text.

J=SLAP-2332
TEST=auto

added storybook and jest tests for the different field types in field value direct answer

### DIFF
--- a/src/components/DirectAnswer.tsx
+++ b/src/components/DirectAnswer.tsx
@@ -1,16 +1,18 @@
-import { useSearchState, DirectAnswerType, DirectAnswer as DirectAnswerData } from '@yext/search-headless-react';
-import { renderHighlightedValue } from './utils/renderHighlightedValue';
-import classNames from 'classnames';
-import { ReactNode } from 'react';
-import { useComposedCssClasses } from '../hooks/useComposedCssClasses';
+import {
+  useSearchState,
+  DirectAnswerType,
+  DirectAnswer as DirectAnswerData,
+} from '@yext/search-headless-react';
 import {
   ThumbsFeedbackCssClasses,
   ThumbsFeedback,
   builtInCssClasses as thumbsFeedbackCssClasses
 } from './ThumbsFeedback';
-
+import classNames from 'classnames';
+import { useComposedCssClasses } from '../hooks/useComposedCssClasses';
 import { useCardAnalyticsCallback } from '../hooks/useCardAnalyticsCallback';
 import { useCardFeedbackCallback } from '../hooks/useCardFeedbackCallback';
+import { FieldValueDirectAnswer } from './FieldValueDirectAnswer';
 
 /**
  * Props for {@link DirectAnswer}.
@@ -71,56 +73,20 @@ export function DirectAnswer(props: DirectAnswerProps): JSX.Element | null {
   }
 
   const cssClasses = getCssClassesForAnswerType(composedCssClasses, directAnswerResult.type);
-  const title = directAnswerResult.type === DirectAnswerType.FeaturedSnippet
-    ? directAnswerResult.value as string // TODO: update with other direct answer changes
-    : `${directAnswerResult.entityName} / ${directAnswerResult.fieldName}`;
-  const description: ReactNode = directAnswerResult.type === DirectAnswerType.FeaturedSnippet
-    ? renderHighlightedValue(directAnswerResult.snippet, { highlighted: cssClasses.highlighted })
-    : directAnswerResult.value as string; // TODO: update with other direct answer changes
-  const link = directAnswerResult.relatedResult.link;
-
-  function getLinkText(directAnswerResult: DirectAnswerData) {
-    const isSnippet = directAnswerResult.type === DirectAnswerType.FeaturedSnippet;
-    const name = directAnswerResult.relatedResult.name;
-    const snippetLinkMessage = 'Read more about ';
-
-    return (<>
-      {isSnippet && name && <div className='pt-4 text-neutral'>
-        {snippetLinkMessage}
-        <a
-          className='text-primary'
-          href={link}
-          onClick={handleClickViewDetails}
-        >
-          {name}
-        </a>
-      </div>}
-      {!isSnippet && link && <div className='pt-4 text-neutral'>
-        <a
-          href={link}
-          className='text-primary'
-          onClick={handleClickViewDetails}
-        >
-          View Details
-        </a>
-      </div>}
-    </>);
-  }
-
-  const containerCssClasses = classNames(cssClasses.directAnswerContainer, {
-    [cssClasses.directAnswerLoading ?? '']: isLoading
+  const containerCssClasses = classNames(composedCssClasses.directAnswerContainer, {
+    [composedCssClasses.directAnswerLoading ?? '']: isLoading
   });
 
   return (
     <div className={containerCssClasses}>
-      <div className={cssClasses.answerContainer}>
-        {title &&
-          <div className={cssClasses.header}>{title}</div>}
-        <div className={cssClasses.content}>
-          <div className={cssClasses.body}>{description}</div>
-          {link && getLinkText(directAnswerResult)}
-        </div>
-      </div>
+      {directAnswerResult.type === DirectAnswerType.FieldValue
+        ? <FieldValueDirectAnswer
+          result={directAnswerResult}
+          cssClasses={cssClasses}
+          handleClickViewDetails={handleClickViewDetails}
+        />
+        : <div>Snippet!</div> //TODO: create FeaturedSnippetDirectAnswer component
+      }
       <ThumbsFeedback
         onClick={handleClickFeedbackButton}
         customCssClasses={composedCssClasses}
--- a/src/components/DirectAnswer.tsx
+++ b/src/components/DirectAnswer.tsx
@@ -8,8 +8,7 @@ import {
   ThumbsFeedback,
   builtInCssClasses as thumbsFeedbackCssClasses
 } from './ThumbsFeedback';
-import classNames from 'classnames';
-import { useComposedCssClasses } from '../hooks/useComposedCssClasses';
+import { twMerge, useComposedCssClasses } from '../hooks/useComposedCssClasses';
 import { useCardAnalyticsCallback } from '../hooks/useCardAnalyticsCallback';
 import { useCardFeedbackCallback } from '../hooks/useCardFeedbackCallback';
 import { FieldValueDirectAnswer } from './FieldValueDirectAnswer';
@@ -74,9 +73,10 @@ export function DirectAnswer(props: DirectAnswerProps): JSX.Element | null {
   }
 
   const cssClasses = getCssClassesForAnswerType(composedCssClasses, directAnswerResult.type);
-  const containerCssClasses = classNames(cssClasses.directAnswerContainer, {
-    [cssClasses.directAnswerLoading ?? '']: isLoading
-  });
+  const containerCssClasses = twMerge(
+    cssClasses.directAnswerContainer,
+    isLoading && cssClasses.directAnswerLoading
+  );
 
   const link = directAnswerResult.relatedResult.link;
   function getLinkText(directAnswerResult: DirectAnswerData) {
--- a/src/components/DirectAnswer.tsx
+++ b/src/components/DirectAnswer.tsx
@@ -13,6 +13,7 @@ import { useComposedCssClasses } from '../hooks/useComposedCssClasses';
 import { useCardAnalyticsCallback } from '../hooks/useCardAnalyticsCallback';
 import { useCardFeedbackCallback } from '../hooks/useCardFeedbackCallback';
 import { FieldValueDirectAnswer } from './FieldValueDirectAnswer';
+import { renderHighlightedValue } from './utils/renderHighlightedValue';
 
 /**
  * Props for {@link DirectAnswer}.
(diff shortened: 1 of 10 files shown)


src/components/FieldValueDirectAnswer.tsx
```diff
@@ -0,0 +1,122 @@
...
+/**
+ * Renders a field value direct answer provided by the Search API.
+ */
+export function FieldValueDirectAnswer({
+  result,
+  viewDetailsClickHandler,
+  cssClasses = {}
+}: FieldValueDirectAnswerProps): JSX.Element {
+  const title = `${result.entityName} / ${result.fieldName}`;
+  const link = result.relatedResult.link;
```
nmanu1, line 39: out of curiosity, do you know why we only check this link to determine if the DA has a link? I was looking in the HH theme code and it seems like we use this link if it's present, and if not, we look for a `landingPageUrl` in the `rawData` and use that if we find it

yen-tt: I think we might have based it off component.js in allfields-standard of the theme, which uses relatedResult.link (aka data.website). I see that `landingPageUrl` is only use documentsearch-standard of the theme for DA. I can update to fall back on landingPageUrl!

src/components/FieldValueDirectAnswer.tsx
```diff
@@ -0,0 +1,122 @@
...
+ */
+interface FieldValueDirectAnswerCssClasses {
+  header?: string,
+  body?: string,
+  content?: string,
+  answerContainer?: string
+}
+
+/**
+ * Renders a field value direct answer provided by the Search API.
```
oshi97, line 31: Could we have an `@internal` here if we don't plan for this component to be public

oshi97 (commented):
+1!

src/components/FieldValueDirectAnswer.tsx
yen-tt: sure!

src/components/FieldValueDirectAnswer.tsx
```diff
@@ -0,0 +1,122 @@
...
+        ? getListJsxElement(result.value, url => getUrlJsxElement(url))
+        : getUrlJsxElement(result.value);
+    case BuiltInFieldType.Phone:
+      return getUrlJsxElement(`tel:${result.value}`, result.value);
+    case BuiltInFieldType.Email:
+      return getListJsxElement(result.value, e => getUrlJsxElement(`mailto:${e}`, e));
+    case BuiltInFieldType.Address:
+      return getAddressJsxElement(result.value);
+    case BuiltInFieldType.RichText:
+      return <div>rich text</div>; //TODO: use react-markdown
```
oshi97, line 86: I guess we could still display the rich text raw string in the meantime? Do we have an item to use react-markdown?

nmanu1: just made SLAP-2340 for that!

yen-tt: looks like nidhi just made one so I can update the comment. sure I can display the raw string

oshi97 (dismissed):

tests/components/FieldValueDirectAnswer.test.tsx
```diff
@@ -0,0 +1,168 @@
+import { cleanup, render, screen } from '@testing-library/react';
+import { FieldValueDirectAnswer } from '../../src/components/FieldValueDirectAnswer';
+import { Address, BuiltInFieldType, DirectAnswerType, FieldValueDirectAnswer as FieldValueDirectAnswerType, Source } from '@yext/search-headless-react';
+import { userEvent } from '@storybook/testing-library';
+
+const fieldValueDAResult: FieldValueDirectAnswerType = {
```
nmanu1, line 6: can this be moved to fixtures? I think we already have a field value example there?

src/components/FieldValueDirectAnswer.tsx
```diff
@@ -0,0 +1,124 @@
...
+ *
+ * @internal
+ */
+export function FieldValueDirectAnswer({
+  result,
+  viewDetailsClickHandler,
+  cssClasses = {}
+}: FieldValueDirectAnswerProps): JSX.Element {
+  const title = `${result.entityName} / ${result.fieldName}`;
+  const link = result.relatedResult.link ?? result.relatedResult.rawData.landingPageUrl as string;
```
nmanu1, line 41: this `link` logic will be the same regardless of DA type. what do you think about leaving it in `DirectAnswer` and passing it as a prop?

src/components/DirectAnswer.tsx
```diff
@@ -71,14 +74,11 @@ export function DirectAnswer(props: DirectAnswerProps): JSX.Element | null {
...
 
   const cssClasses = getCssClassesForAnswerType(composedCssClasses, directAnswerResult.type);
-  const title = directAnswerResult.type === DirectAnswerType.FeaturedSnippet
-    ? directAnswerResult.value as string // TODO: update with other direct answer changes
-    : `${directAnswerResult.entityName} / ${directAnswerResult.fieldName}`;
-  const description: ReactNode = directAnswerResult.type === DirectAnswerType.FeaturedSnippet
-    ? renderHighlightedValue(directAnswerResult.snippet, { highlighted: cssClasses.highlighted })
-    : directAnswerResult.value as string; // TODO: update with other direct answer changes
-  const link = directAnswerResult.relatedResult.link;
+  const containerCssClasses = classNames(cssClasses.directAnswerContainer, {
```
nmanu1, line 77: I think we want to use `tw-merge` instead of `classNames` now, right?

src/components/FieldValueDirectAnswer.tsx
```diff
@@ -0,0 +1,124 @@
...
+  /** Handle onClick event for "View Details" link. */
+  viewDetailsClickHandler?: () => void,
+  /** CSS classes for customizing the component styling. */
+  cssClasses?: FieldValueDirectAnswerCssClasses
+}
+
+/**
+ *  The CSS class interface for FieldValueDirectAnswer.
+ */
+interface FieldValueDirectAnswerCssClasses {
```
nmanu1, line 23: should `DirectAnswerCssClasses` extend this?

src/components/FieldValueDirectAnswer.tsx
yen-tt: I'm incline to keep link in the individual BlahDirectAnswer component, since it's based on what they would display

src/components/FieldValueDirectAnswer.tsx
yen-tt: I think we can keep them separate. From alex's previous pr for simplifying the css classes. DirectAnswer css classes that is expose to user would specify `answer` and `description`, which is apply different for FeaturedSnippet and FieldValue's `body` and `header`.  (transform through the function `getCssClassesForAnswerType`)

tests/components/FieldValueDirectAnswer.test.tsx
```diff
@@ -0,0 +1,155 @@
...
+    const address: Address = {
+      line1: '61 Ninth Avenue',
+      line2: '14th Floor',
+      city: 'New York',
+      region: 'NY',
+      postalCode: '10011',
+      countryCode: 'US',
+      extraDescription
+    };
+    const result = {
```
nmanu1, line 145: maybe the result data for the different field types can also be moved to fixtures? I think it could be reused in the stories then?

src/components/DirectAnswer.tsx
yen-tt: sure, we should probably update other areas that uses classNames too. But that's gonna be a separate item

nmanu1: sounds good!

tests/components/FieldValueDirectAnswer.test.tsx
yen-tt: mm most of them are a bit different, maybe example emails and number. I think it's fine to keep them separate?

nmanu1: ok, just checking!

src/components/FieldValueDirectAnswer.tsx
```diff
@@ -0,0 +1,124 @@
...
+    case BuiltInFieldType.ComplexURL:
+      const url = result.value.url;
+      const displayUrl = result.value.preferDisplayUrl ? result.value.displayUrl : url;
+      return getUrlJsxElement(url, displayUrl);
+    case BuiltInFieldType.URL:
+      return Array.isArray(result.value)
+        ? getListJsxElement(result.value, url => getUrlJsxElement(url))
+        : getUrlJsxElement(result.value);
+    case BuiltInFieldType.Phone:
+      return getUrlJsxElement(`tel:${result.value}`, result.value);
```
nmanu1, line 80: can this be named slightly differently, since it looks like it's also used for things other than urls?

src/components/FieldValueDirectAnswer.tsx
```diff
@@ -0,0 +1,124 @@
...
+function getAddressJsxElement(address: Address): JSX.Element {
+  // user specified display Address in KM
+  if (address.extraDescription) {
+    return <div>{address.extraDescription}</div>;
+  }
+  return <div>
+    {address.line1 && <p>{address.line1}</p>}
+    {address.line2 && <p>{address.line2}</p>}
+    {address.line3 && <p>{address.line3}</p>}
+    <p>{address.city}, {address.region} {address.postalCode}</p>
```
nmanu1, line 121: are these fields all required?

tests/components/DirectAnswer.stories.tsx
```diff
@@ -12,14 +18,69 @@ const meta: ComponentMeta<typeof DirectAnswer> = {
...
+): JSX.Element {
   return (
     <SearchHeadlessContext.Provider value={generateMockedHeadless({
-      directAnswer: fieldValueDAState
+      directAnswer: {
+        result: {
+          ...baseDirectAnswerResult,
+          ...(fieldType !== undefined && { fieldType }),
+          ...(value !== undefined && { value }),
+        } as FieldValueDirectAnswerType
```
nmanu1, line 35: is there a way this could be refactored so casting isn't needed?

tests/components/DirectAnswer.test.tsx
```diff
@@ -21,28 +21,8 @@ it('renders null when there is no direct answer in state', () => {
   expect(container).toBeEmptyDOMElement();
 });
 
-describe('FieldValue direct answer', () => {
+describe('Field value direct answer analytics', () => {
   beforeEach(() => mockState(fieldValueDAState));
-
-  it('title text', () => {
-    render(<DirectAnswer />);
-    const expectedTitle = '[entityName] / [fieldName]';
```
nmanu1, line 29: do we need a test like this in the `FieldValueDirectAnswer` test suite?

yen-tt: didn't think it's needed, these tests are basically testing UI changes, which the storybook stories should take care of already! As for the link test, that is added to FieldValueDirectAnswer test suite already in the pr

src/components/FieldValueDirectAnswer.tsx
nmanu1: I thought only `countryCode` was required, so maybe we want to check that `city`, `region`, and `postalCode` exist before displaying them? otherwise it might look weird if, for example, there was no `city`

juliannzhou: since `city`, `region`, and `postalCode` are only required after a `countryCode` is set, I was thinking maybe we'd check for those after checking for `countryCode`?

src/components/FieldValueDirectAnswer.tsx
yen-tt: sure, would getAnchorTagJsxElement or getHrefJsxElement work?

tests/components/DirectAnswer.stories.tsx
yen-tt: I couldn't figure out a nice way to do it. It seems to be the TS error (without using the type cast `as`) may stem from the `RecursivePartial` type util we have which makes it think the result where should be of the address (error: `Type '"unknown"' is not assignable to type 'EnumOrLiteral<BuiltInFieldType.Address>'`). I can dig deeper if you like, or leave it for now and have another item to investigate?

nmanu1: got it. I think it's fine to leave as is then. I was just curious

src/components/FieldValueDirectAnswer.tsx
nmanu1: sure!

tests/components/DirectAnswer.stories.tsx
```diff
@@ -12,14 +18,69 @@ const meta: ComponentMeta<typeof DirectAnswer> = {
 };
 export default meta;
 
-export const FieldValue = (args: DirectAnswerProps) => {
+const baseDirectAnswerResult = fieldValueDAState.result as FieldValueDirectAnswerType;
```
nmanu1, line 21: can this assertion be removed since we have one in the function below?

tests/__fixtures__/data/directanswers.ts
```diff
@@ -22,10 +22,15 @@ export const fieldValueDAState: RecursivePartial<DirectAnswerState> = {
     type: DirectAnswerType.FieldValue,
     entityName: '[entityName]',
     fieldName: '[fieldName]',
+    fieldType: BuiltInFieldType.SingleLineText,
```
nmanu1, line 25: now that we've added more fields, can we remove the `RecursivePartial` type?

src/components/FieldValueDirectAnswer.tsx
yen-tt: updated to optionally display city, region, and postalCode

nmanu1 (approved):

PR #299, opened 2022-08-22
handle "unknown" field type in field value DA

yen-tt:
Update DirectAnswer component to accept an optional prop field `UnknownFieldTypeDisplay` to handle rendering results of "unknown" field type in field value direct answer. If not provided, FieldValueDirectAnswer will use the default component, which would attempt to render result.value as a string with toString() method. It would console log a warning for user if it encounter results of "unknown" field type where value is not of type string or number. 

J=SLAP-2337
TEST=auto&manual

Added jest tests for handling unknown field type in field value through custom and default UnknownFieldTypeDisplay component.
Also manually tested with a UnknownFieldTypeDisplay component in Universal page and with a query "what is yen banana price" for a fieldApiName "c_bananaPrice"


src/components/FieldValueDirectAnswer.tsx
```diff
@@ -58,7 +65,19 @@ export function FieldValueDirectAnswer({
...
 }
 
-function getResultContent(result: FieldValueDirectAnswerType): JSX.Element {
+function DefaultUnknownFieldTypeDisplay({ result }: { result: UnknownFieldValueDirectAnswer }): JSX.Element {
+  if (typeof result.value !== 'string' && typeof result.value !== 'number') {
+    console.warn(`Unknown field type for direct answer with "${result.fieldApiName}" fieldApiName. Rendering result's value as a string.`
+      + '\nConsider using prop "UnknownFieldTypeDisplay" in DirectAnswer component to properly render result of unknown field type.');
+  }
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  return getTextJsxElement((result.value as any).toString());
```
oshi97, line 74: Would we ever wanna use JSON.stringify here?

yen-tt: sure, that would provide a bit more info to user, even if it looks a bit ugly, which would prompt developer to provide UnknownFieldTypeDisplay.

oshi97 (approved):

PR #302, opened 2022-09-07
Further updates to account for changes in `SearchHeadlessProvider`.

tmeyer2115:
The README was updated to reflect the fact that `SearchHeadlessProvider` now takes in a Headless instance. The README also shows someone how to handle acquisition of a session tracking ID. I added session tracking logic to the test-site as well as a further reference.

TEST=auto

Ensured that all unit and WCAG tests passed locally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,7 @@ Once the library and its peer dependencies are installed, the components can be 
 ```tsx
 import { SearchHeadlessProvider } from '@yext/search-headless-react';
 import { SearchBar, UniversalResults } from '@yext/search-ui-react';
+import { v4 as uuidv4 } from 'uuid';
 
 const config = {
   apiKey: '<apiKey>',
@@ -50,7 +51,7 @@ const searcher = provideHeadless(config);
 searcher.setSessionTrackingEnabled(true);
 let sessionId = window.sessionStorage.getItem('sessionId');
 if (!sessionId) {
-  sessionId = 'newUUID';
+  sessionId = uuidv4();
   window.sessionStorage.setItem('sessionId', sessionId);
 }
 searcher.setSessionId(sessionId);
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -45,9 +46,19 @@ const config = {
...
+if (!sessionId) {
+  sessionId = uuidv4();
+  window.sessionStorage.setItem('sessionId', sessionId);
+}
+searcher.setSessionId(sessionId);
+
 function App() {
   return (
-    <SearchHeadlessProvider {...config}>
+    <SearchHeadlessProvider searcher={searcher}>
```
oshi97, line 61: Kind of a tangent but are we planning to get the SearchProvider (combined provider SearchHeadlessProvider and AnalyticsProvider) that Davish talked about in v2? It came to mind since we'd have to update the readme again when adding it

README.md
```diff
@@ -45,9 +46,19 @@ const config = {
   experienceVersion: 'PRODUCTION',
 }
 
+const searcher = provideHeadless(config);
+
+searcher.setSessionTrackingEnabled(true);
+let sessionId = window.sessionStorage.getItem('sessionId');
```
oshi97, line 52: So we're not going to provide an acquireSessionId method for now, and ask people to do this manually?

README.md
tmeyer2115: I don't really think we have time right now. It needn't be in a V2. We could still keep both the `SearchHeadlessProvider` and `AnalyticsProvder`. We'd be adding a convenience wrapper over them.

README.md
tmeyer2115: Yep!

oshi97 (approved):

README.md
yen-tt, line 52: I guess we are just providing an example of client side rendering so no need to guard window or anything?

yen-tt (commented):
oh we should probably update search-headless-react readme for the changes in SearchHeadlessProvider as well!

yen-tt (approved):

PR #304, opened 2022-09-07
update readme "Getting Started" section

yen-tt:
update readme wording now that `SearchHeadlessProvider` requires a `SearchHeadless` instance instead of directly passing in the credentials. Also include small changes such that the code sample works out of the box.

TEST=manual

copy and pasted the code sample to test-site App.tsx using the actual credentials, See that it works as expected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ Once the library and its peer dependencies are installed, our React Components s
 
 ```tsx
 import { provideHeadless, SearchHeadlessProvider } from '@yext/search-headless-react';
-import { SearchBar, UniversalResults } from '@yext/search-ui-react';
+import { SearchBar, UniversalResults, VerticalConfigMap } from '@yext/search-ui-react';
 import { v4 as uuidv4 } from 'uuid';
 
 const config = {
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -32,10 +32,10 @@ npx install-peerdeps @yext/search-ui-react
 ```
 The command will work with Yarn so long as NPM 6+ is installed on the machine.
 
-Once the library and its peer dependencies are installed, the components can be rendered using React by placing them inside the `SearchHeadlessProvider` with the appropriate credentials:
+Once the library and its peer dependencies are installed, the components can be rendered using React by placing them inside the `SearchHeadlessProvider`. `SearchHeadlessProvider` requires a `SearchHeadless` instance, which is created using `provideHeadless(...)` with the appropriate credentials:
```
tmeyer2115, line 35: Small nit: I'd say something like "Once the library and its peer dependencies are installed, our React Components should be nested inside the `SearchHeadlessProvider`". I think that's a bit more accurate and concise. React Components implies Components handled by React. The current wording makes it sound a bit like the Components won't render unless there's a `SearchHeadlessProvider`. I'm not sure this is true for all our Components. Some might render, they just would bomb out on user interaction.

README.md
```diff
@@ -60,7 +60,11 @@ function App() {
   return (
     <SearchHeadlessProvider searcher={searcher}>
       <SearchBar />
-      <UniversalResults />
+      <UniversalResults verticalConfigMap={{
```
oshi97, line 63: Can we have verticalConfigMap outside of the function? Since we've been following the react-perf guidelines even though they're debatable

oshi97 (approved):

PR #307, opened 2022-09-21
Only `e.preventDefault` when Dropdown is active

nmanu1:
Update the `Dropdown` component to only call `e.preventDefault` for the up/down arrow keys if the dropdown is active.

J=SLAP-2351
TEST=manual

Spin up the test-site, load results, and see that the up and down arrow keys scroll the page when the dropdown is closed. When the dropdown is active, up/down arrows work as expected to navigate through dropdown options. Tested both the `SearchBar` and `FilterSearch` dropdowns.

### DIFF
--- a/src/components/Dropdown/Dropdown.tsx
+++ b/src/components/Dropdown/Dropdown.tsx
@@ -1,4 +1,4 @@
-import { createElement, isValidElement, PropsWithChildren, ReactNode, useEffect, useMemo, useRef, useState } from 'react';
+import { createElement, isValidElement, KeyboardEvent, PropsWithChildren, ReactNode, useEffect, useMemo, useRef, useState } from 'react';
 import { DropdownContext, DropdownContextType } from './DropdownContext';
 import { InputContext, InputContextType } from './InputContext';
 import useRootClose from '@restart/ui/useRootClose';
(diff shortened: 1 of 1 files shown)


src/components/Dropdown/Dropdown.tsx
```diff
@@ -102,14 +102,14 @@ export function Dropdown(props: PropsWithChildren<DropdownProps>): JSX.Element {
   }, { disabled: !isActive });
 
   useGlobalListener('keydown', e => {
```
oshi97, line 104: Could we use a listener scoped to just the events in the Dropdown, instead of useGlobalListener?

nmanu1: sure, updated to limit it to the dropdown container instead of the whole document like useGlobalListener does

src/components/Dropdown/Dropdown.tsx
```diff
@@ -101,49 +100,57 @@ export function Dropdown(props: PropsWithChildren<DropdownProps>): JSX.Element {
...
-    if (e.key === 'ArrowDown' || e.key === 'ArrowUp') {
-      e.preventDefault();
-    }
-
-    if (!isActive) {
-      return;
-    }
+  useEffect(() => {
+    const dropdownContainer = containerRef.current;
+    dropdownContainer?.addEventListener('keydown', handleKeyDown);
```
oshi97, line 105: Is it possible to use `onKeyDown` on the container div instead of `addEventListener`?

nmanu1: yup, updated!

yen-tt (approved):

PR #308, opened 2022-09-29
Address test-site dependency vulnerabilities

nmanu1:
Upgrade `ejs`, `async`, and `terser` in the test-site to address vulnerabilities in older versions of those packages.

J=none
TEST=none

### DIFF
--- a/test-site/package-lock.json
+++ b/test-site/package-lock.json
@@ -1739,16 +1739,51 @@
         "has-flag": "^4.0.0"
       }
     },
+    "../node_modules/@jridgewell/gen-mapping": {
+      "version": "0.3.2",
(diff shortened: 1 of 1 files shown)


test-site/package-lock.json
```diff
@@ -1739,16 +1739,51 @@
         "has-flag": "^4.0.0"
```
yen-tt, line 1739: does the top level package.json needs to be updated as well? since some of these seems to be linked to the top level node_modules

nmanu1: only `terser` (and its dependencies) is in the top-level package.json too and it is already on a vulnerability-free version. we might have updated it there and then forgotten to also update it in the test-site

yen-tt (approved):

PR #275, opened 2022-08-03
Add Visual Coverage Test + Combined Code Coverage for Coveralls

yen-tt:
Use storybook test-runner and `@storybook/addon-coverage` to get visual coverage report. The setup requires spinning up storybook before generating the coverage numbers. Now, we have three test scripts:
- test => combined test coverage
- test:unit => jest test coverage only
- test:visual => story book coverage only (note: this only collect coverage on src/components)

Updated coverage github action to also create/update comment which display code coverage percentage for unit, visual, and combined tests

the test-runner only works with jest v27 right now. As nidhi noted in her [pr](https://github.com/yext/search-ui-react/pull/274), storybook will be adding support for v28 shortly (https://github.com/storybookjs/test-runner/pull/154).

[playwright](https://www.npmjs.com/package/playwright) library is needed in the github action for test-storybook.
Screen Shot 2022-08-03 at 4 51 03 PM

NOTE: The report generated from `nyc` merge command may not be super accurate after some comparison between to the unit vs visual vs merge report (ex: [github issue](https://github.com/istanbuljs/nyc/issues/1302)).

J=SLAP-2269 & SLAP-2270
TEST=manual&auto

- See that running `npm run test:visual` generates lcov report in coverage/visual folder. And the report is printed out in terminal.
- See that a comment is made to the PR about the three coverage percentages.
- See that coverall percentage increase without changes to tests (increased due to combined test coverage)

PR #309, opened 2022-09-29
Upgrade Jest to v29

nmanu1:
Upgrade Jest from v27 to v29. We had [previously](https://github.com/yext/search-ui-react/pull/169) upgraded Jest to v28, but had to [downgrade](https://github.com/yext/search-ui-react/pull/275) to v27 because at the time, `@storybook/test-runner` only supported up to v27. In v0.7 of `@storybook/test-runner`, they changed Jest to be an internal dependency instead of a peer dependency, allowing us to use whatever version of Jest we want.

There was only one breaking change in v29 that affected our code, besides those that already needed to be made when upgrading from v27 to v28 (see previous PR [description](https://github.com/yext/search-ui-react/pull/169#issue-1262431931)). This was the `jsdom` upgrade from v19 to v20 in `jest-environment-jsdom`, which requires the `typescript` version to be 4.5 or higher.

With the new support added in v29 and `uuid` v9, we are now able to remove the manual resolving needed in `tests/__setup__/resolver.ts`. For more context on why this was initially needed with Jest v28, see #169.

J=SLAP-2376
TEST=manual

See that existing tests still run properly.


package.json
```diff
@@ -82,13 +82,14 @@
     "eslint-plugin-react-hooks": "^4.3.0",
     "eslint-plugin-react-perf": "^3.3.1",
     "generate-license-file": "^1.3.0",
-    "jest": "^27.5.1",
+    "jest": "^29.1.1",
+    "jest-environment-jsdom": "^29.1.1",
```
yen-tt, line 86: to confirm, we still need to install this separately for v29?

nmanu1: yes, for v28 and higher (including v29) `jest-environment-jsdom` must be installed separately

yen-tt (approved):

PR #311, opened 2022-10-07
mapbox component jest tests

yen-tt:
added jest tests for Mapbox component, specifically for the two prop fields: `getCoordinate` and `onDrag`. The remaining props are best tested through storybook.

note that `mapbox-gl` is mocked as it seems to use some web browser functionality on initialization that is not supported in jest test environment (jsdom). Without mocking, jest would present an error `Error: Failed to initialize WebGL.`.
- tried third party libraries `jest-webgl-canvas-mock` and `jest-canvas-mock` still result in more window properties access related errors (`[TypeError: e.window.Worker is not a constructor]`, which could be resolve by [manually mocking Worker](https://github.com/facebook/jest/issues/3449), `[TypeError: this.target.addEventListener is not a function]`...).
- Looked into how pageJS test mapbox provider in their map component: the map wrapper component invoke a [load function](https://github.com/yext/pages/blob/main/packages/pages/src/components/map/map.tsx#L112) to construct [a script tag](https://github.com/yext/components-tsx-maps/blob/main/src/Providers/Mapbox.js#L137) from `yext/components-tsx-maps` and the assertions in the unit tests seem to check for the rendering of the wrapper component only and not waiting for the script to load. No content/interaction of the map was tested from my understanding.
- Looked into how mapbox-gl-js does their own testing: they had a couple util files to set up the environment for their tests (mock requests, [mock HtmlCanvas/WebLG in window](https://github.com/mapbox/mapbox-gl-js/blob/main/src/util/window.js) object, etc.).

I decided this complexity is probably unnecessary to add to the repo but we can discuss if other may think differently. We can have jest strictly test new isolated functionalities added from the component and leave all the UI rendering and mapbox interaction to storybook tests.

SLAP-2222
TEST=auto

new jest tests passed

### DIFF
--- a/src/components/MapboxMap.tsx
+++ b/src/components/MapboxMap.tsx
@@ -74,7 +74,7 @@ export interface MapboxMapProps<T> {
  * `import 'mapbox-gl/dist/mapbox-gl.css';`
  *
  * Or, user may add a stylesheet link in their html page:
- * `<link href="https://api.mapbox.com/mapbox-gl-js/VERSION/mapbox-gl.css" rel="stylesheet">`
+ * `<link href="https://api.mapbox.com/mapbox-gl-js/v2.9.2/mapbox-gl.css" rel="stylesheet" />`
  *
  * @param props - {@link MapboxMapProps}
  * @returns A React element containing a Mapbox Map
--- a/tests/components/MapboxMap.test.tsx
+++ b/tests/components/MapboxMap.test.tsx
@@ -1,0 +1,112 @@
+import { render } from '@testing-library/react';
+import { mockAnswersState } from '../__utils__/mocks';
+import { CoordinateGetter, MapboxMap, Coordinate } from '../../src/components/MapboxMap';
+import { Source, State } from '@yext/search-headless-react';
+import { Map, Marker } from 'mapbox-gl';
+
+jest.mock('@yext/search-headless-react');
+jest.mock('mapbox-gl');
+
+interface Location {
+  customCoordinate: Coordinate
+}
+
+const mockedStateDefaultCoordinate: Partial<State> = {
+  vertical: {
+    verticalKey: 'vertical',
+    results: [{
+      rawData: {
+        yextDisplayCoordinate: {
+          latitude: 1,
+          longitude: 1
+        }
+      },
+      source: Source.KnowledgeManager
+    }]
+  }
+};
+
+const mockedStateCustomCoordinate: Partial<State> = {
+  vertical: {
+    verticalKey: 'vertical',
+    results: [{
+      rawData: {
+        customCoordinate: {
+          latitude: 2,
+          longitude: 2
+        }
+      },
+      source: Source.KnowledgeManager
+    }]
+  }
+};
+
+const mockedStateWrongCoordinateType: Partial<State> = {
+  vertical: {
+    verticalKey: 'vertical',
+    results: [{
+      rawData: {
+        yextDisplayCoordinate: [1, 1]
+      },
+      source: Source.KnowledgeManager
+    }]
+  }
+};
+
+describe('default "getCoordinate"', () => {
+  it('uses result\'s "yextDisplayCoordinate" for marker location', () => {
+    mockAnswersState(mockedStateDefaultCoordinate);
+    const setLngLat = jest.spyOn(Marker.prototype, 'setLngLat').mockReturnValue(Marker.prototype);
+    render(<MapboxMap mapboxAccessToken='TEST_KEY' />);
+    expect(setLngLat).toBeCalledWith({ lat: 1, lng: 1 });
+  });
+
+  it('displays an error when "yextDisplayCoordinate" field is not present in result', () => {
+    mockAnswersState(mockedStateCustomCoordinate);
+    const setLngLat = jest.spyOn(Marker.prototype, 'setLngLat').mockReturnValue(Marker.prototype);
+    const errorSpy = jest.spyOn(console, 'error').mockImplementation();
+    render(<MapboxMap mapboxAccessToken='TEST_KEY' />);
+    expect(errorSpy).toBeCalledTimes(1);
+    const expectedMessage = 'Unable to use the default "yextDisplayCoordinate" field as the result\'s coordinate';
+    expect(errorSpy).toBeCalledWith(expect.stringContaining(expectedMessage));
+    expect(setLngLat).not.toBeCalled();
+  });
+
+  it('displays an error when "yextDisplayCoordinate" field is not of type "Coordinate"', () => {
+    mockAnswersState(mockedStateWrongCoordinateType);
+    const setLngLat = jest.spyOn(Marker.prototype, 'setLngLat').mockReturnValue(Marker.prototype);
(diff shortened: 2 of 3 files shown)


src/components/MapboxMap.tsx
```diff
@@ -74,7 +74,7 @@ export interface MapboxMapProps<T> {
  * `import 'mapbox-gl/dist/mapbox-gl.css';`
  *
  * Or, user may add a stylesheet link in their html page:
- * `<link href="https://api.mapbox.com/mapbox-gl-js/VERSION/mapbox-gl.css" rel="stylesheet">`
+ * `<link href="https://api.mapbox.com/mapbox-gl-js/v2.9.2/mapbox-gl.css" rel="stylesheet" />`
```
nmanu1, line 77: is there a reason this is changed to a specific version?

tests/components/MapboxMap.test.tsx
```diff
@@ -0,0 +1,112 @@
...
+      },
+      source: Source.KnowledgeManager
+    }]
+  }
+};
+
+describe('default "getCoordinate"', () => {
+  it('uses result\'s "yextDisplayCoordinate" for marker location', () => {
+    mockAnswersState(mockedStateDefaultCoordinate);
+    const setLngLat = jest.spyOn(Marker.prototype, 'setLngLat').mockReturnValue(Marker.prototype);
```
nmanu1, line 59: could this spy/mock happen in a `beforeEach` to reduce repetition?

tests/components/MapboxMap.test.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+it('executes custom "getCoordinate" and use the derived coordinate for marker location', () => {
+  mockAnswersState(mockedStateCustomCoordinate);
+  const setLngLat = jest.spyOn(Marker.prototype, 'setLngLat').mockReturnValue(Marker.prototype);
+  const customGetCoordinate: CoordinateGetter<Location> = jest.fn().mockImplementation(
+    result => result.rawData.customCoordinate);
+
+  render(<MapboxMap mapboxAccessToken='TEST_KEY' getCoordinate={customGetCoordinate} />);
+  expect(customGetCoordinate).toBeCalledTimes(1);
+  expect(setLngLat).toBeCalledWith({ lat: 2, lng: 2 });
```
nmanu1, line 95: do we want to check that no console error gets logged here?

tests/components/MapboxMap.test.tsx
```diff
@@ -0,0 +1,112 @@
...
+});
+
+it('registers "onDrag" callback to Mapbox\'s event listener for "drag to pan" interaction', () => {
+  jest.useFakeTimers();
```
nmanu1, line 99: for my own understanding, what does this do? does this set up timers so we can call `jest.advanceTimersByTime` later?

tests/components/MapboxMap.test.tsx
```diff
@@ -0,0 +1,112 @@
...
+  jest.spyOn(Marker.prototype, 'setLngLat').mockReturnValue(Marker.prototype);
+  const mapOnEventListener = jest.spyOn(Map.prototype, 'on')
+    .mockImplementation((e, cb) => {
+      e === 'drag' && cb({});
+      return Map.prototype;
+    });
+  const onDragFn = jest.fn().mockImplementation(() => console.log('HERE'));
```
nmanu1, line 106: can we either check for this console log or remove it from the mocked implementation?

tests/components/MapboxMap.test.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+  render(<MapboxMap mapboxAccessToken='TEST_KEY' onDrag={onDragFn} />);
+  expect(mapOnEventListener).toBeCalledWith('drag', expect.anything());
```
nmanu1, line 109: does a map drag automatically occur when it's rendered for the first time?

src/components/MapboxMap.tsx
yen-tt: it's the version that we currently have in package.json so I think we should just specify it here to match it

tests/components/MapboxMap.test.tsx
yen-tt: I need access to the function being spied on in certain tests; I slightly prefer calling it in the individual tests then having a global `let setLngLat` and set it in beforeEach for the tests -- but I can change to it if you think that's fine

tests/components/MapboxMap.test.tsx
yen-tt: I didn't think it's needed if there is already an assertion for derived coordinate with `setLngLat`, since the console errors are related not being able to get the coordinate field.

tests/components/MapboxMap.test.tsx
yen-tt: yup, `jest.useFakeTimers()` is replacing the original implementation of `setTimeout()` and other timer functions. Then `jest.advanceTimersByTime` "advance" the time and executes all pending "macro-tasks" that have been queued via setTimeout() or setInterval() during that time frame. This help with testing `onDrag` since we wrapped it with our `useDebouncedFuntion` (for more info: [jest docs](https://jestjs.io/docs/timer-mocks#advance-timers-by-time))

tests/components/MapboxMap.test.tsx
yen-tt: no

src/components/MapboxMap.tsx
nmanu1: I'm imagining it might be a little weird then if we bump to a new minor/patch but forget to update here. our docs would essentially be telling people to use an older version. but, if that seems unlikely, then I'm fine with leaving it in

tests/components/MapboxMap.test.tsx
nmanu1: maybe I'm missing something then. what causes the event listener to be called then?

tests/components/MapboxMap.test.tsx
nmanu1: oh I see that not all the tests are wrapped in this `describe`. I'm fine with leaving as is then

src/components/MapboxMap.tsx
yen-tt: I think leaving it as VERSION seems a bit vague and we do ant people to adhere to the same version we use in package.json to avoid potential version conflict, so we just need to be mindful about updating docs when we bump mapbox version.

tests/components/MapboxMap.test.tsx
yen-tt: since mapbox-gl is mocked, I can't simulate the actual drag event. So, I mocked `Map.on` event listener register to immediately invoked the callback function if the event type is 'drag' instead of waiting for a drag event to occur. Assuming that mapbox-gl's Map.on works as expected, we are testing in our code whether we register the event listener properly when initialize the map and if callback with the debounce time is executed as expected

tests/components/MapboxMap.test.tsx
yen-tt: sure thing

nmanu1 (approved):

PR #312, opened 2022-10-10
docs(MapboxMap): add storybook stories

tatimblin:
single marker story
multiple marker story
custom marker story - follow docs.mapbox.com guide to add a simple popup pin

J=SLAP-2223
TEST=manual
 
serve and view stories


.storybook/preview.js
```diff
@@ -1,4 +1,5 @@
 import './index.css';
+import 'mapbox-gl/dist/mapbox-gl.css';
```
tatimblin, line 2: I believe this is an extra step required for all third-party css to render in storybook. Tailwind also requires it. Wanted to bring attention to it though

.storybook/tailwind.config.cjs
```diff
@@ -4,5 +4,10 @@ module.exports = {
   ...tailwindConfig,
   content: [
     './src/**/*.{ts,tsx}'
-  ]
+  ],
+  safelist: [
```
tatimblin, line 8: This adds tailwind classes to storybook to be used outside of a story in the ComponentMeta setup region. This was needed for `h-screen` to make the map full width. I figured why not allow all classes since this only affects the storybook environment, but I can limit to just what I need as well.

test-site/src/components/MapPin.tsx
```diff
@@ -1,9 +1,48 @@
...
+
+export const MapPin: PinComponent<Location> = (props: MapPinProps) => {
+  const popupRef = useRef(new Popup({ offset: 15 }));
+  const initialRender = useRef(true);
+  const [active, setActive] = useState(false);
+
+  useEffect(() => {
+    if (!initialRender.current && active) {
+      const popup = document.createElement('div');
+      ReactDOM.render(<MapPopup entity={props.result} />, popup);
```
oshi97, line 28: I feel like we might not need this MapPopup component if it's literally just a div? Maybe we could just call setText instead and then skip out on the ReactDOM.render?

test-site/src/components/MapPin.tsx
```diff
@@ -1,9 +1,48 @@
...
+      popupRef.current
+        .setLngLat(coordinate)
+        .setDOMContent(popup)
+        .addTo(props.mapbox);
+    }
+    initialRender.current = false;
+  }, [active]);
```
oshi97, line 35: I think this effect is missing dependencies? It looks like there's a number of eslint complaints, TIL our linting workflow doesn't run when you merge into a feature branch. We probably will update this in the future but in the meantime if you could manually run eslint?

test-site/src/components/MapPin.tsx
```diff
@@ -1,9 +1,48 @@
...
+  const yextCoordinate: Location["yextDisplayCoordinate"] = props.result.rawData.yextDisplayCoordinate;
+  if (!yextCoordinate) return <></>;
+  const coordinate = transformToMapboxCoord(yextCoordinate);
 
-export const MapPin: PinComponent<Location> = () => {
   return (
-    <div className='relative z-10'>
+    <button
+      className='relative z-10'
+      onClick={() => setActive(!active)}
```
oshi97, line 44: I might be misreading but I think here `active` doesn't actually determine the active state? I checked out this branch out of curiosity and it looks like that after the initial render of the popup, if you click out of it to close it, you need to click twice on the button to reopen it? I assume this is because 
1. when you click outside of the Popup mapbox automatically calls Popup.remove() and
2. the first click on the button changes active to false, which doesn't do anything, and only on the second click does it switch back to true which then re-adds the popup

tests/components/MapboxMap.stories.tsx
```diff
@@ -0,0 +1,88 @@
+import { useCallback } from 'react';
+import { ComponentMeta } from '@storybook/react';
+import { SearchHeadlessContext, Source } from '@yext/search-headless-react';
+
+import { generateMockedHeadless } from '../__fixtures__/search-headless';
+import { MapboxMap, MapboxMapProps } from '../../src/components/MapboxMap';
+import { MapPin } from '../../test-site/src/components/MapPin';
+
+const meta: ComponentMeta<typeof MapboxMap<Location>> = {
```
oshi97, line 9: I think this is using some built in Location interface which doesn't seem to be the one we want?

tests/components/MapboxMap.stories.tsx
```diff
@@ -0,0 +1,88 @@
...
+        id: 'id1'
+      }
+    ],
+    resultsCount: 1,
+    limit: 1
+  }
+};
+
+// Deep copy state to make an alternative with multiple locations
+let mockedHeadlessStateMultiple = JSON.parse(JSON.stringify(mockedHeadlessState));
```
oshi97, line 48: maybe lodash clonedeep instead?

oshi97: There might be some value trying to use some preexisting fixtures but imo it's also fine having bespoke data

sample.env
```diff
@@ -0,0 +1 @@
+REACT_APP_MAPBOX_API_KEY=[[Your Mapbox API Key]]
```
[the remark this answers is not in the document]
oshi97: Oooh is this to show the "shape" of the env file? Kinda curious how most repos do this

.github/workflows/percy-snapshots.yml
```diff
@@ -13,3 +13,4 @@ jobs:
       percy_script: npx percy storybook ./storybook-static
     secrets:
       PERCY_TOKEN: ${{ secrets.PERCY_TOKEN }}
+      REACT_APP_MAPBOX_API_KEY: ${{ secrets.REACT_APP_MAPBOX_API_KEY }}
```
nmanu1, line 16: I don't think the reusable workflow for the Percy snapshots expects a secret with this name, so you might have to update that [workflow](https://github.com/yext/slapshot-reusable-workflows/blob/main/.github/workflows/percy_snapshots.yml) too

test-site/src/components/MapPin.tsx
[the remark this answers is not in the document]
tatimblin: updated

.github/workflows/percy-snapshots.yml
tatimblin: @nmanu1 you're right the build failed `Invalid secret, REACT_APP_MAPBOX_API_KEY is not defined in the referenced workflow`. Do I just need to merge that into v1 (like the other prs) or is there ui stuff to do in GitHub as well?

yen-tt: no additional setup is needed -- just need the pr with the changes to merge to v1 branch (and main branch) in this [repo](https://github.com/yext/slapshot-reusable-workflows)

test-site/src/pages/LocationsPage.tsx
```diff
@@ -10,16 +10,14 @@ import {
...
 import { LngLat, LngLatBounds } from 'mapbox-gl';
 import { useCallback, useLayoutEffect } from 'react';
 import { MapPin } from '../components/MapPin';
 
 export interface Location {
-  yextDisplayCoordinate: {
-    latitude: number,
-    longitude: number
-  } | undefined
+  yextDisplayCoordinate: Coordinate | undefined
```
yen-tt, line 20: can this be `yextDisplayCoordinate?: Coordinate` instead?

test-site/src/components/MapPin.tsx
```diff
@@ -1,9 +1,48 @@
...
+}
+
+const transformToMapboxCoord = (coordinate: Coordinate): LngLatLike => ({
+  lng: coordinate.longitude,
+  lat: coordinate.latitude,
+});
+
+export const MapPin: PinComponent<Location> = (props: MapPinProps) => {
+  const popupRef = useRef(new Popup({ offset: 15 }));
+  const initialRender = useRef(true);
```
yen-tt, line 22: what is `initialRender` for? The useEffect below is guarded by the field `active` so I think we could remove this?

test-site/src/components/MapPin.tsx
```diff
@@ -1,9 +1,48 @@
...
+  const [active, setActive] = useState(false);
+
+  useEffect(() => {
+    if (!initialRender.current && active) {
+      const popup = document.createElement('div');
+      ReactDOM.render(<MapPopup entity={props.result} />, popup);
+      popupRef.current
+        .setLngLat(coordinate)
+        .setDOMContent(popup)
```
yen-tt, line 31: I think it might be simpler if we just do something like `setText(props.result.name)` so we don't need  the component `MapPopup` to render text

tests/components/MapboxMap.stories.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+export const MultiplePins = (args: MapboxMapProps<Location>) => {
+  return (
+    <SearchHeadlessContext.Provider value={generateMockedHeadless(mockedHeadlessStateMultiple)}>
+      <MapboxMap {...args} />
+    </SearchHeadlessContext.Provider>
+  );
+};
+
+export const CustomPin = (args: MapboxMapProps<Location>) => {
```
yen-tt, line 81: I think you forgot an import for this `Location` interface -- it's currently using some default typescript location interface

test-site/src/components/MapPin.tsx
tatimblin: renamed to something clearer. I can only get the mapboxgl.Map to update in useEffect [according to their docs](https://docs.mapbox.com/help/tutorials/use-mapbox-gl-js-with-react/), and my experience trying alternatives. If you have any other ideas I'd be happy to try, this feels a little weird.

.github/workflows/percy-snapshots.yml
tatimblin: talked offline

tests/components/MapboxMap.stories.tsx
```diff
@@ -0,0 +1,88 @@
...
+        type: null,
+      }
+    }
+  },
+  parameters: { layout: 'fullscreen' },
+  decorators: [(Story) => (<div className="h-screen"><Story /></div>)]
+};
+export default meta;
+
+const mockedHeadlessState = {
```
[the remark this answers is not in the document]
nmanu1: Sorry, I meant if you could look at the mocked state objects in `Mapbox.test.tsx` and maybe merge those objects with the ones in this file and then make a new file in `__fixtures__/data` (like `__fixtures__/data/mapbox.ts`) and put them in there

test-site/src/components/MapPin.tsx
tatimblin: Reverted back to `active` and added an `on close` event to the popup following docs here. Like this much better https://docs.mapbox.com/mapbox-gl-js/api/markers/#popup.event:close

test-site/src/components/MapPin.tsx
tatimblin: This prevents the code running onMount

test-site/src/components/MapPin.tsx
tatimblin: Fixed this in another. I kept the `setDOMContent` because I think it is more likely that someone would want to add HTML over just text

test-site/src/components/MapPin.tsx
yen-tt: The code in the useEffect below is guarded by `active` field, which is set to false as the initial state value. without `initialRender`, I believe the code portion for adding a popup still wouldn't run on the first render / onMount until active is set to true through `onClick`.

tatimblin: Oh I see. I removed `active` in a followup commit so I thought you meant both weren't needed. This is fixed

test-site/src/components/MapPin.tsx
```diff
@@ -1,18 +1,55 @@
...
+  const { coordinate, mapbox, result } = props;
+  const [active, setActive] = useState(false);
+  const popupRef = useRef(new Popup({ offset: 15 })
+    .on('close', () => setActive(false))
+  );
+
+  useEffect(() => {
+    if (active) {
+      const popup = document.createElement('div');
+      popup.innerHTML = `<div>${result.name}</div>`;
```
oshi97, line 29: I double checked in browser console and here popup will be a div with a div inside it `<div><div>result.name</div></div>`, when I think we just want `<div> result.name </div>`. Maybe just `popup.innerText = result.name`?

test-site/src/components/MapPin.tsx
```diff
@@ -1,18 +1,55 @@
...
+}
+
+const transformToMapboxCoord = (coordinate: Coordinate): LngLatLike => ({
+  lng: coordinate.longitude,
+  lat: coordinate.latitude,
+});
+
+export const MapPin: PinComponent<Location> = (props: MapPinProps<Location>) => {
+  const { coordinate, mapbox, result } = props;
+  const [active, setActive] = useState(false);
```
oshi97, line 21: instead of maintaining our own active state does [`Popup.isOpen`](https://docs.mapbox.com/mapbox-gl-js/api/markers/#popup#isopen) work?

tests/components/MapboxMap.stories.tsx
```diff
@@ -0,0 +1,90 @@
+import { ComponentMeta, Story } from '@storybook/react';
+import _ from 'lodash';
```
oshi97, line 2: nit but I think I prefer importing directly from `lodash/clonedeep`? it doesn't really matter here but I think(?) the bundle size is a little smaller that way

.github/workflows/percy-snapshots.yml
```diff
@@ -7,9 +7,10 @@ on:
 
 jobs:
```
oshi97, line 8: would you mind adding `features/*` to the branches list in https://github.com/yext/search-ui-react/blob/544236fa3042725affa356dc5f2cad9406d19f35/.github/workflows/run-linting.yml

test-site/src/components/MapPin.tsx
tatimblin: We could use that for the conditional in useEffect, but having the active state hook is necessary for triggering the useEffect. There doesn't seem to be any Popup event for triggering one to open, only a callback for when one is

.storybook/preview.js
```diff
@@ -35,7 +36,8 @@ export const parameters = {
         'SpellCheck',
         'ResultsCount',
         'LocationBias',
-        'Dropdown'
+        'Dropdown',
+        'MapboxMap',
```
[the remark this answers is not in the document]
tatimblin: asking https://yext.slack.com/archives/C032CKFARGS/p1665597903799669

test-site/src/components/MapPin.tsx
oshi97: Instead of a useEffect we could just add the popup in the click handler right?

oshi97: I think that might be slightly simpler...?

tatimblin: No that hasn't been able to work for me or @nmanu1

test-site/src/components/MapPin.tsx
```diff
@@ -1,18 +1,55 @@
...
+  const popupRef = useRef(new Popup({ offset: 15 })
+    .on('close', () => setActive(false))
+  );
+
+  useEffect(() => {
```
oshi97, line 26: should this be a useLayoutEffect since it's doing something visual to the page?

test-site/src/components/MapPin.tsx
```diff
@@ -1,18 +1,55 @@
...
+      const popup = document.createElement('div');
+      popup.innerHTML = `<div>${result.name}</div>`;
+      popupRef.current
+        .setLngLat(transformToMapboxCoord(coordinate))
+        .setDOMContent(popup)
+        .addTo(mapbox);
+    }
+  }, [active, coordinate, mapbox, result]);
+
+  function handleClick() {
```
oshi97, line 35: can we use useCallback here? I think we're (at least I'm) not 100% sold on using it for all functions are passed as props but we've been very consistent with it in the component lib

test-site/src/components/MapPin.tsx
tatimblin: useLayoutEffect does not work, and [useEffect is preferred](https://reactjs.org/docs/hooks-reference.html#:~:text=prefer%20the%20standard%20useeffect%20when%20possible%20to%20avoid%20blocking%20visual%20updates)

src/components/MapboxMap.tsx
```diff
@@ -9,7 +9,12 @@ import ReactDOM from 'react-dom';
  *
  * @public
  */
-export type PinComponent<T> = (props: { result?: Result<T>, index?: number }) => JSX.Element;
+export type PinComponent<T> = (props: {
+  coordinate: Coordinate,
```
tmeyer2115, line 13: Adding a `coordinate` attribute here feels a bit strange to me. It reads as somewhat redundant since `result` is also provided (from which `coordinate` is derived). I definitely understand adding a reference to the mapbox object for various pin effects, etc. 

Could we undo the addition of `coordinate` in this PR? I think that requires some further thought.

tatimblin: Yeah I can remove it

tests/__fixtures__/data/mapbox.ts
```diff
@@ -0,0 +1,76 @@
+import { Source } from '@yext/search-headless-react';
+
+const locationVerticalSingle = {
```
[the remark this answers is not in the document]
tatimblin: I can do this in a separate PR

test-site/src/components/MapPin.tsx
```diff
@@ -17,20 +16,21 @@ const transformToMapboxCoord = (coordinate: Coordinate): LngLatLike => ({
...
+  const { mapbox, result } = props;
+  const yextCoordinate = result.rawData.yextDisplayCoordinate;
   const [active, setActive] = useState(false);
   const popupRef = useRef(new Popup({ offset: 15 })
     .on('close', () => setActive(false))
   );
 
   useEffect(() => {
-    if (active) {
+    if (active && yextCoordinate) {
```
tmeyer2115, line 27: I'm assuming we need to keep this in because of TypeScript, but is the `yextCoordinate` check actually necessary here? Doesn't the `Mapbox` Component take care of that essentially?

.storybook/main.js
```diff
@@ -37,6 +37,6 @@ module.exports = {
   },
   env: (config) => ({
     ...config,
-    REACT_APP_MAPBOX_API_KEY: process.env.REACT_APP_MAPBOX_API_KEY,
+    REACT_APP_MAPBOX_API_KEY: process.env.MAPBOX_API_KEY || process.env.REACT_APP_MAPBOX_API_KEY,
```
tmeyer2115, line 40: Will there ever be a `REACT_APP_MAPBOX_API_KEY` env var since we renamed the secret?

tatimblin: We renamed the secret in github ci, but keeping `REACT_APP_MAPBOX_API_KEY` here means we can reuse the env variable used in the create-react-app generated test-site otherwise our .env would be like this

```
MAPBOX_API_KEY=pk.1234
REACT_APP_MAPBOX_API_KEY=pk.1234
```

which isn't terrible, but seems like a better experience to set it once. fyi it is a create-react-app limitation to not be able to read env variables without their prefix

test-site/src/components/MapPin.tsx
oshi97: Out of curiosity is it possible to make `yextCoordinate` required in the `Location` interface?

tatimblin:
Still not able to get the map to render in the percy snapshot even though the testing-library checks pass which check that the map was loaded. Think I'll need some help on this

oshi97 (approved):

yen-tt:
is there suppose to be multiple pins shown here: https://percy.io/cd93112c/answers-react-components/builds/22287752/changed/1241974795?browser=edge&browser_ids=18%2C22%2C23%2C25&subcategories=approved&viewLayout=side-by-side&viewMode=new&width=1280&widths=375%2C1280?

PR #314, opened 2022-10-12
Fix TS errors in `tests` directory

nmanu1:
Fix all TS errors in the `tests` folder.

J=SLAP-2360
TEST=manual

Go through all files in `tests` and see that there are no more errors.

### DIFF
--- a/tests/__fixtures__/core/search-service.ts
+++ b/tests/__fixtures__/core/search-service.ts
@@ -1,5 +1,5 @@
 /* eslint-disable @typescript-eslint/no-unused-vars */
-import { SearchService, UniversalSearchRequest, UniversalSearchResponse, VerticalSearchRequest, VerticalSearchResponse } from '@yext/search-core';
+import { SearchService, Source, UniversalSearchRequest, UniversalSearchResponse, VerticalSearchRequest, VerticalSearchResponse } from '@yext/search-core';
 
 /* A sample Universal Search response. */
 const mockedUniversalResponse: UniversalSearchResponse = {
@@ -11,7 +11,14 @@ const mockedUniversalResponse: UniversalSearchResponse = {
 const mockedVerticalResponse: VerticalSearchResponse = {
   uuid: '123',
   queryId: 'query-id',
-  verticalResults: null
+  verticalResults: {
+    appliedQueryFilters: [],
+    queryDurationMillis: 500,
+    verticalKey: 'vertical',
+    results: [],
+    resultsCount: 0,
+    source: 'KNOWLEDGE_MANAGER' as Source
+  }
 };
 
 /**
--- a/tests/__utils__/mocks.ts
+++ b/tests/__utils__/mocks.ts
@@ -18,9 +18,9 @@ export function spyOnActions(): jest.Mocked<SearchActions> {
 
 export type RecursivePartial<T> = {
   [P in keyof T]?:
-  T[P] extends (infer U)[] ? RecursivePartial<U>[] :
+  T[P] extends ((infer U)[] | undefined) ? RecursivePartial<U>[] :
     // eslint-disable-next-line @typescript-eslint/ban-types
-    T[P] extends object ? RecursivePartial<T[P]> :
+    T[P] extends (object | undefined) ? RecursivePartial<T[P]> :
       T[P];
 };
 
--- a/tests/__utils__/mocks.ts
+++ b/tests/__utils__/mocks.ts
@@ -18,10 +18,10 @@ export function spyOnActions(): jest.Mocked<SearchActions> {
 
 export type RecursivePartial<T> = {
   [P in keyof T]?:
-  T[P] extends ((infer U)[] | undefined) ? RecursivePartial<U>[] :
-    // eslint-disable-next-line @typescript-eslint/ban-types
-    T[P] extends (object | undefined) ? RecursivePartial<T[P]> :
-      T[P];
+  T[P] extends undefined ? undefined :
+    T[P] extends ((infer U)[] | undefined) ? RecursivePartial<U>[] :
+      T[P] extends (object | undefined) ? RecursivePartial<T[P]> :
+        T[P];
 };
 
 export function mockAnswersState(
--- a/tests/components/AppliedFilters.test.tsx
+++ b/tests/components/AppliedFilters.test.tsx
@@ -1,6 +1,6 @@
 import { render, screen } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
-import { Matcher, Source, State, FiltersState, SelectableStaticFilter } from '@yext/search-headless-react';
+import { Matcher, Source, State, FiltersState, SelectableStaticFilter, AppliedQueryFilterType } from '@yext/search-headless-react';
 import { AppliedFilters } from '../../src/components/AppliedFilters';
 import { getSelectableFieldValueFilters } from '../../src/utils/filterutils';
 import { createHierarchicalFacet } from '../__utils__/hierarchicalfacets';
@@ -65,7 +65,8 @@ const mockedState: Partial<State> = {
         fieldId: 'produt',
         matcher: Matcher.Equals,
         value: 'Yext Listings'
-      }
+      },
+      type: AppliedQueryFilterType.FieldValue
     }]
   },
   searchStatus: {
@@ -102,14 +103,14 @@ describe('AppliedFilters', () => {
 
   it('Facets are rendered', () => {
     render(<AppliedFilters />);
-    const facetOptionDisplayName = mockedState.filters.facets[0].options[0].displayName;
-    expect(screen.getByText(facetOptionDisplayName)).toBeDefined();
+    const facetOptionDisplayName = mockedState.filters?.facets?.[0].options[0].displayName;
+    expect(facetOptionDisplayName && screen.getByText(facetOptionDisplayName)).toBeDefined();
   });
 
   it('Applied query filters are rendered', () => {
     render(<AppliedFilters />);
-    const appliedFilterDisplayName = mockedState.vertical.appliedQueryFilters[0].displayValue;
-    expect(screen.getByText(appliedFilterDisplayName)).toBeDefined();
+    const appliedFilterDisplayName = mockedState.vertical?.appliedQueryFilters?.[0].displayValue;
+    expect(appliedFilterDisplayName && screen.getByText(appliedFilterDisplayName)).toBeDefined();
   });
 
   it('Filters with the fieldId of "builtin.entityType" are hidden by default', () => {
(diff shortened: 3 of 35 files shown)


tests/components/AppliedFilters.test.tsx
```diff
@@ -259,7 +260,7 @@ describe('AppliedFilters with hierarchical facets', () => {
     expect(filterPills).toHaveLength(3);
 
     const fruitButton = screen.queryByLabelText('Remove "food" filter');
-    userEvent.click(fruitButton);
+    fruitButton && userEvent.click(fruitButton);
```
yen-tt, line 263: for these cases where we are certain the element must be defined, I believe we can just use `getBy...` functions instead of `queryBy...` so we don't need to check for undefined (e.g. use `getByLabelText` here so we don't need the `fruitButton &&` part)

tests/components/LocationBias.test.tsx
```diff
@@ -68,8 +63,8 @@ beforeEach(() => {
 
 it('renders the proper text (location DisplayName, method, and update btn)', () => {
   render(<LocationBias />);
-  const expectedLocationName = mockedStateVaDevice.location.locationBias.displayName;
-  const locationNameElement = screen.getByText(expectedLocationName);
+  const expectedLocationName = mockedStateVaDevice.location?.locationBias?.displayName;
+  const locationNameElement = expectedLocationName && screen.getByText(expectedLocationName);
```
yen-tt, line 67: This is needed because of the `Partial<State>` type for `mockedStateVaDevice` right? Since these fields are technically defined in the const variable, I think it would be best if we don't need to do checks like this in multiple places.

What do you think about updating the type for `mockedStateVaDevice` to something like `Partial<State & { location: LocationBias }>` and we can keep these two lines as before, without the optional chaining and conditional guards? If so, we can update the types for other const variables for mocking states as well

tests/__utils__/mocks.ts
```diff
@@ -18,9 +18,9 @@ export function spyOnActions(): jest.Mocked<SearchActions> {
 
 export type RecursivePartial<T> = {
   [P in keyof T]?:
-  T[P] extends (infer U)[] ? RecursivePartial<U>[] :
+  T[P] extends ((infer U)[] | undefined) ? RecursivePartial<U>[] :
     // eslint-disable-next-line @typescript-eslint/ban-types
-    T[P] extends object ? RecursivePartial<T[P]> :
+    T[P] extends (object | undefined) ? RecursivePartial<T[P]> :
```
yen-tt, line 23: what is the reason we want it to extend a union type with `undefined`? if the field `T[P]` is undefined, I imagine we wouldn't want to perform `RecursivePartial<T[P]>` again here to derive the type but rather use the type `T[P]` or `undefined` as its own type.

tests/components/RangeInput.stories.tsx
```diff
@@ -1,78 +1,53 @@
...
 };
 
 export const DisabledForceDisplayTooltip = Disabled.bind({});
 DisabledForceDisplayTooltip.play = ({ canvasElement }) => {
   const canvas = within(canvasElement);
   const tooltip = canvas.getByText('Unselect an option to enter in a range.').parentElement;
-  tooltip.style.visibility = 'visible';
+  if (tooltip) {
+    tooltip.style.visibility = 'visible';
+  }
```
yen-tt, line 50: can you check if we actually need the `parentElement` to be visible or just the element return from `canvas.getByText('Unselect an option to enter in a range.')`? If it's the latter, we can remove the if condition.

tests/components/UniversalResults.stories.tsx
```diff
@@ -7,9 +7,20 @@ import { RecursivePartial } from '../__utils__/mocks';
 import { verticalResults } from '../__fixtures__/data/universalresults';
 import { DefaultRawDataType } from '../../src/models/DefaultRawDataType';
 
+const verticalConfigMap = {
```
yen-tt, line 10: nit: can we add type to this?

tests/components/UniversalResults.test.tsx
```diff
@@ -33,16 +33,16 @@ describe('UniversalResults', () => {
...
     expect(screen.getByText(verticals[0].verticalKey)).toBeDefined();
     expect(screen.getByText(verticals[1].verticalKey)).toBeDefined();
     expect(screen.queryByText(verticals[2].verticalKey)).toBeNull();
 
     expect(screen.queryByText(verticals[0].appliedQueryFilters[0].displayValue)).toBeNull();
 
     function checkResultData(resultData: Record<string, unknown>) {
-      expect(screen.getByText(resultData.name.toString())).toBeDefined();
-      expect(screen.getByText(resultData.description.toString())).toBeDefined();
+      expect(typeof resultData.name === 'string' && screen.getByText(resultData.name)).toBeDefined();
```
yen-tt, line 44: I'm a little wary of this format of `expect(blah && blah).toBeDefined();` since `expect(false).toBeDefined()` would still pass even though we intend to assess the actual return value of the second part of the condition. Is there something else we can do? like changing the type or the way we assert things here

tests/components/LocationBias.test.tsx
nmanu1: sure, updated the type so the optional chaining and guards could be removed

tests/components/UniversalResults.test.tsx
nmanu1: changed it to assert `toBeTruthy` instead of `toBeDefined` so the false case won't pass

tests/__utils__/mocks.ts
nmanu1: the problem with the current type is that it doesn't correctly identify optional object or array types. these types don't just extend `(infer U)[]` or `object`, they also extend `undefined`, which is why I needed to extend the union. I added another check first for types that only extend `undefined` so it doesn't become `RecursivePartial<undefined>`

tests/__fixtures__/core/search-service.ts
```diff
@@ -11,7 +11,14 @@ const mockedUniversalResponse: UniversalSearchResponse = {
...
   uuid: '123',
   queryId: 'query-id',
-  verticalResults: null
+  verticalResults: {
+    appliedQueryFilters: [],
+    queryDurationMillis: 500,
+    verticalKey: 'vertical',
+    results: [],
+    resultsCount: 0,
+    source: 'KNOWLEDGE_MANAGER' as Source
```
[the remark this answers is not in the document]
tmeyer2115: Gotcha. Actually, I'll need to think on if that's something we want in Core. People likely aren't crafting their own `Result` sets.

tests/__utils__/mocks.ts
```diff
@@ -18,10 +18,10 @@ export function spyOnActions(): jest.Mocked<SearchActions> {
 
 export type RecursivePartial<T> = {
   [P in keyof T]?:
-  T[P] extends (infer U)[] ? RecursivePartial<U>[] :
-    // eslint-disable-next-line @typescript-eslint/ban-types
-    T[P] extends object ? RecursivePartial<T[P]> :
-      T[P];
+  T[P] extends undefined ? undefined :
```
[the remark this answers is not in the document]
nmanu1: I looked around a little and it seems like our original version of RecursivePartial was taken from a StackOverflow [post](https://stackoverflow.com/questions/41980195/recursive-partialt-in-typescript). A few other places suggest this as well, because there's no TS utility that currently does this. I tried looking for a 3rd party library, but the couple I was able to find (e.g. `DeepPartial` from https://github.com/joonhocho/tsdef) don't quite work for our purposes because they make primitive types partials too.

A comment in the original StackOverflow post I linked above actually suggests almost the same change I made because they encountered the same problem with optional array and object types not working correctly

tests/__fixtures__/core/search-service.ts
yen-tt, line 20: can we do `Source.KnowledgeManager` instead?

tests/components/SpellCheck.test.tsx
```diff
@@ -38,12 +38,14 @@ describe('SpellCheck', () => {
   it('Suggestion is formatted properly', () => {
     render(<SpellCheck />);
     expect(screen.getByText('Did you mean')).toBeDefined();
-    expect(screen.getByText(mockedState.spellCheck.correctedQuery)).toBeDefined();
+    const correctedQuery = mockedState.spellCheck?.correctedQuery;
+    expect(correctedQuery && screen.getByText(correctedQuery)).toBeTruthy();
```
yen-tt, line 42: Similar thing to LocationBias test, can we try updating the mockedState type to specify that spellCheck is defined and remove optional chaining? There may be more places in other test files where we can update the types too.

tests/__fixtures__/core/search-service.ts
nmanu1: that fixes the TS error but when running the Jest tests, all of the test suites that use vertical search fail with the error that it's referencing the `KnowledgeManager` property of `Source`, which is undefined. this is probably because we're mocking Core, so it can't find `Source`. I could re-define the `Source` enum in this file to fix both errors, but that seemed a little weird. I don't have a strong preference though if that seems better than the assertion

tests/components/SpellCheck.test.tsx
nmanu1: I initially didn't update this file (or the others) to be like `LocationBias` because even in the `SpellCheckState`, the `correctedQuery` is optional (and same with `displayName` for facets and `displayValue` for applied query filters). So, we'd be constructing an entirely new type to force the TS compiler to realize that certain fields are defined. It seemed a little weird to be defining a bunch of new types just for the tests.

Another option could be to use a const assertion on the mocked state objects instead of giving them a `Partial<State>` type. For example, 
```ts
const mockedState = {
  spellCheck: {
    correctedQuery: 'Correction',
    enabled: false
  },
  ...
} as const;
```
I think I prefer this over creating new types in each test file. Thoughts?

yen-tt: mm but that would remove the State typing for `mockedState` right? no TS suggestion or warn about accidentally add a new field to the variable

nmanu1: yeah, it would only give an error if you did something incorrect and then tried passing it somewhere that expects a type of `Partial<State>`

yen-tt: I do like the TS suggestions but const approach would avoid the need for optional chaining and guards. I'm okay with either way then, keep Partial<State> type or using const.

tests/__fixtures__/core/search-service.ts
yen-tt: I see, that's fine then

yen-tt (approved):

PR #315, opened 2022-10-14
Fix React SSR hydration ID mismatch error

yen-tt:
This PR is to fix the errors `Warning: Prop "id" did not match. Server: "some-id". Client: "another-id"`, reported in https://yext.slack.com/archives/C032CKFARGS/p1664916209295449

React expects that the **rendered content is identical between the server and the client**. However, our usage of `uuid` means different id is generated from the HTML content generated through SSR vs the first render in client side during hydration process. This is a known issues in React SSR (issues in react repo[[1](https://github.com/facebook/react/issues/4000)][[2](https://github.com/facebook/react/issues/5867)][[3](https://github.com/facebook/react/issues/1137)]) and React released a new hook `useId` in React 18 to address it.

I found three potential solutions for React <18 to **generate unique but predictable id**, each with an external library we could use:
1) Use context to keep a counter in state, so counter is not shared across renders. This approach is implemented in [react-aria](https://www.npmjs.com/package/react-aria). When using SSR with React Aria, applications must be wrapped in an `SSRProvider` component to reset the count in server side to be zero for each fresh render. Multiple copies of React Aria is not supported so it would be a peer-dep
```
//in component lib (dropdown)
import { useId } from 'react-aria';
...
const screenReaderUUID: string = useId();

//in pageJS template using SSR
import { SSRProvider } from 'react-aria';
...
<SSRProvider>
    <TestComponent />
</SSRProvider>
```
2) use a global counter to increment on initial renders and expose a reset function for SSR side. This approach is implemented in [react-id-generator](https://www.npmjs.com/package/react-id-generator). There's a function `resetId` required to invoke from SSR side (such as in PageJS `serverRenderRoute` or somewhere in user's template page) to avoid mismatch during refresh, since browser generator will always restart from "1" but server's count will keep incrementing.

3) Don't use IDs at all on the server; patch after first render. This approach is implemented in [reach/auto-id.](https://www.npmjs.com/package/@reach/auto-id) ID returned as undefined/empty on the first render so server and client have the same markup during hydration process. After render, patch the components with an incremented ID (second render) in useLayoutEffect. (more info [here](https://github.com/reach/reach-ui/blob/dev/packages/auto-id/src/reach-auto-id.ts)) No changes needed outside of using `import { useId } from '@reach/auto-id';` in our component lib.

I decided to go with the last approach using `react/auto-id` because:
- requires no additional work / changes from user side (only in component lib)
- contain fallback implementation to React 18 `useId` if it's available
- avoid potential concurrent rendering problem (such as suspense boundaries)
- the main issue with this approach is that it causes a double render on any components with `useId`. However, since we (1) only update the ID attribute on DOM on second render, and (2) our components that uses `useId` already requires multiple renders to get setup, the performance is not greatly affected. I used Profiler in our test-site to measure performance, the number of renders and time spent for `SearchBar`, `StaticFilters`, and `FilterSearch` did not change as the dispatch from `useId` is grouped with other changes required in our components on second render. This is only needed for React <18

J=SLAP-2403
TEST=manual&auto

`npm pack` a build of the component lib with the new changes and tested SearchBar, FilterSearch, and StaticFilers:
- in the starter PageJS project (https://github.com/yext/pages-starter-react-locations) with React 17
- in another [yext-sites-starter](https://github.com/tmeyer2115/yext-sites-starter) repo with React 18 override

See that the warning about Prop `id` mismatched no longer appear in console

added new jest test for Dropdown and StaticFilter, see that test failed the previous usage of uuid, and passed with reach/auto-id.

### DIFF
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -30,6 +30,10 @@ module.exports = {
   framework: '@storybook/react',
   staticDirs: ['./public'],
   webpackFinal: async (config) => {
+    //use commonjs entry point for "@reach" packages
+    config.resolve.alias['@reach/auto-id'] = require.resolve('@reach/auto-id');
+    config.resolve.alias['@reach/utils'] = require.resolve('@reach/utils');
+
     config.resolve.alias['./SearchCore'] = require.resolve('../tests/__fixtures__/core/SearchCore.ts');
     config.resolve.alias['../utils/location-operations'] = require.resolve('../tests/__fixtures__/utils/location-operations.ts');
     return config;
--- a/src/components/Dropdown/Dropdown.tsx
+++ b/src/components/Dropdown/Dropdown.tsx
@@ -3,11 +3,11 @@ import { DropdownContext, DropdownContextType } from './DropdownContext';
 import { InputContext, InputContextType } from './InputContext';
 import useRootClose from '@restart/ui/useRootClose';
 import { FocusContext, FocusContextType } from './FocusContext';
-import { v4 as uuid } from 'uuid';
 import { ScreenReader } from '../ScreenReader';
 import { recursivelyMapChildren } from '../utils/recursivelyMapChildren';
 import { DropdownItem, DropdownItemProps, DropdownItemWithIndex } from './DropdownItem';
 import useLayoutEffect from 'use-isomorphic-layout-effect';
+import { useId } from '@reach/auto-id';
 
 interface DropdownItemData {
   value: string,
@@ -52,7 +52,7 @@ export function Dropdown(props: PropsWithChildren<DropdownProps>): JSX.Element {
   } = props;
 
   const containerRef = useRef<HTMLDivElement>(null);
-  const screenReaderUUID: string = useMemo(() => uuid(), []);
+  const screenReaderUUID = useId();
   const [screenReaderKey, setScreenReaderKey] = useState<number>(0);
   const [hasTyped, setHasTyped] = useState<boolean>(false);
   const [childrenWithDropdownItemsTransformed, items] = useMemo(() => {
@@ -242,7 +242,7 @@ function useDropdownContextInstance(
   value: string,
   index: number,
   focusedItemData: Record<string, unknown> | undefined,
-  screenReaderUUID: string,
+  screenReaderUUID: string | undefined,
   setHasTyped: (hasTyped: boolean) => void,
   onToggle?: (
     isActive: boolean,
--- a/tests/ssr/utils.tsx
+++ b/tests/ssr/utils.tsx
@@ -1,11 +1,11 @@
 import { render } from '@testing-library/react';
 import { renderToString } from 'react-dom/server';
-import React from 'react';
+import { FunctionComponent } from 'react';
 
-const USE_LAYOUT_EFFECT_ERROR = 'useLayoutEffect does nothing on the server';
+const USE_LAYOUT_EFFECT_ERROR = /useLayoutEffect does nothing on the server/;
 const originalConsoleError = console.error.bind(console.error);
 
-export function testSSR(App: React.FunctionComponent) {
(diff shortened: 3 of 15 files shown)


yen-tt:
looking into the storybook issue

.storybook/main.js
```diff
@@ -30,6 +30,10 @@ module.exports = {
   framework: '@storybook/react',
   staticDirs: ['./public'],
   webpackFinal: async (config) => {
+    //use commonjs entry point for "@reach" packages
+    config.resolve.alias['@reach/auto-id'] = require.resolve('@reach/auto-id');
```
oshi97, line 34: out of curiosity why is this necessary?

src/components/Dropdown/Dropdown.tsx
```diff
@@ -242,7 +242,7 @@ function useDropdownContextInstance(
   value: string,
   index: number,
   focusedItemData: Record<string, unknown> | undefined,
-  screenReaderUUID: string,
+  screenReaderUUID: string | undefined,
```
oshi97, line 245: why is undefined allowed here?

yen-tt: because `useId()` can return undefined on first render

oshi97: Oh I see it's for the server

.storybook/main.js
yen-tt: our storybook uses commonjs and doesn't seem to handle mjs files. 
Screen Shot 2022-10-17 at 12 59 15 PM

`@reach/auto-id` does export a commonjs version. But its package.json is set to `"module": "./dist/reach-auto-id.mjs",` and `"main": "./dist/reach-auto-id.cjs.js",`. I believe Webpack, used by Storybook, by default will try to use "[module" over "main"](https://github.com/webpack/webpack/issues/5756). Alternatively, we could also do `config.resolve.mainFields = ['main', 'module'];` but I thought the alias would be a bit clearer.

tests/ssr/utils.tsx
```diff
@@ -0,0 +1,37 @@
...
+    });
+
+  // server render components to static markup
+  container.innerHTML = renderOnServer();
+
+  // hydrate a container whose HTML contents were rendered by ReactDOMServer
+  render(<App />, { container, hydrate: true });
+  expect(consoleErrorSpy).not.toBeCalledWith(
+    expect.stringContaining('Warning: Prop `%s` did not match. Server: %s Client: %s%s'),
+    'id',
```
tatimblin, line 32: don't need to use string concatenation for a hardcoded value

yen-tt: removed the `expect.stringContaining`, the rest of the arguments is needed in order of the params received by console error.

oshi97: Here does the %s basically mean "any string"?

oshi97: I think here it's react itself that's calling console.error
If react changes the implementation of their console.error call this test will stop functioning properly right? is there some other way to test this?

yen-tt: yea, all prop mismatched error will use this message format. I suppose we don't need to explicitly check for `id` mismatched only, although I don't see other ways of testing this beside checking console error. I attempted to check for the html content of server vs client but it didn't work -- the actual of the container after hydration uses the same as server even though the error is present.
I decided to update the test to assert the number of error logged outside of the useLayoutEffect warning.

oshi97: Sure that makes sense, I can't think of a better way to test this. If they switch to console.warn for example maybe we start getting false positives from the test, but we'd also notice console.warn's in the jest output (hopefully). If they remove the console messages then 🤷 but seems unlikely

oshi97 (approved):

oshi97 (approved):

PR #319, opened 2022-10-27
use createRoot when possible for rendering pin components

oshi97:
This PR uses a dynamic imports with a .catch fallback for when react-dom/client does not exist.

MapboxMap code had to be tweaked due to differences between ReactDOM.render and ReactDOM.createRoot().render(). Namely, createRoot() does not modify the container element unless it already exists on the physical page. This means that we can't call document.creatElement('div'), use createRoot, and THEN attach it to the mapbox map. Instead, we have to attach the div to the map first.

This also seems it make it more difficult to unit test, my guess would be because createRoot does not do anything on the server. For now I use some jest.mocks to ensure the right methods are being called in the right environments.

TEST=manual,auto

storybook can start up (react 17)
test site can display custom pin in both react 17 and 18

### DIFF
--- a/src/components/utils/renderToContainer.tsx
+++ b/src/components/utils/renderToContainer.tsx
@@ -4,9 +4,7 @@ type Renderer = (reactElement: ReactElement, container: Element) => void;
 
 // eslint-disable-next-line @typescript-eslint/ban-ts-comment
 // @ts-ignore we support both react-dom 17 and 18, react-dom/client does not exist in react-dom 17
-const reactDomClientPromise = import('react-dom/client');
-
-const rendererPromiseWithFallback: Promise<Renderer> = reactDomClientPromise.then(reactDomClient => {
+const rendererPromiseWithFallback: Promise<Renderer> = import('react-dom/client').then(reactDomClient => {
   const { createRoot } = reactDomClient;
   const render: Renderer = (reactElement, container) => {
     const root = createRoot(container);
--- a/src/components/utils/renderToContainer.tsx
+++ b/src/components/utils/renderToContainer.tsx
@@ -5,10 +5,13 @@ type Renderer = (reactElement: ReactElement, container: Element) => void;
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 let reactDomClientPromise: Promise<any> = Promise.reject();
 try {
+  // This must be put into a separate variable, otherwise webpack will
+  // try to statically resolve the dynamic import and fail to do so
+  const importString = 'react-dom/client';
   // eslint-disable-next-line @typescript-eslint/ban-ts-comment
   // @ts-ignore we support both react-dom 17 and 18, but need this
   // ts-ignore since react-dom/client does not exist in react-dom 17
-  reactDomClientPromise = import('react-dom/client');
+  reactDomClientPromise = import(importString);
 } catch (e) {}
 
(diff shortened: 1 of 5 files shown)


tsconfig.json
```diff
@@ -1,5 +1,6 @@
 {
   "compilerOptions": {
+    "module": "esnext",
```
oshi97, line 3: this is needed so vscode doesn't complain about us doing dynamic imports without setting `module` to something like "esnext" or "commonjs", since vscode only knows about this tsconfig and not the cjs and esm ones which already have `module` set

src/components/utils/renderToContainer.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * Renders the given reactElement into the container.
+ * Will use createRoot and root.render over ReactDOM.render when possible.
+ */
+const renderToContainer = async (
+  reactElement: ReactElement,
+  container: HTMLElement
+): Promise<void> => {
+  const render = await rendererPromiseWithFallback;
```
tatimblin, line 36: Is this `await` required because of the dynamic import?

src/components/utils/renderToContainer.tsx
```diff
@@ -0,0 +1,40 @@
+import { ReactElement } from 'react';
+
```
tatimblin, line 2: Can you add a heading comment that this can be removed when we stop supporting react 17

tests/components/utils/renderToContainer.test.tsx
```diff
@@ -0,0 +1,23 @@
...
+it('can render a react component into native dom elements', async () => {
+  const createRootRender = jest.fn();
+  const createRootSpy = jest.spyOn(require('react-dom/client'), 'createRoot')
+    .mockImplementation(() => ({ render: createRootRender }));
+  const reactDomRenderSpy = jest.spyOn(require('react-dom'), 'render');
+
+  const container = document.createElement('div');
+
+  await renderToContainer(<span>test me</span>, container);
+  const isReactDom18 = packageJson.devDependencies['react-dom'].match(/18\.\d*\.\d*/);
```
tatimblin, line 19: do you have to change the react version on your machine to test each case?

oshi97: our ci takes care of that since it runs in all react 16 17 and 18
https://github.com/yext/search-ui-react/blob/main/.github/workflows/run-tests.yml
but if you wanted to test it locally then yeah

src/components/utils/renderToContainer.tsx
oshi97: updated!

tests/components/utils/renderToContainer.test.tsx
tatimblin: ci covering that sounds good to me

tatimblin (approved):
+1 (percy failed but no snapshot looks broken, maybe rerun?)

tests/tsconfig.json
```diff
@@ -1,5 +1,8 @@
 {
   "extends": "../tsconfig.json",
+  "compilerOptions": {
+    "resolveJsonModule": true
```
nmanu1, line 4: why is this needed?

tatimblin: To import the package.json I think

nmanu1: oh right, I see that in the test file now

PR #318, opened 2022-10-18
Create hot-fix v1.0.1.

tmeyer2115:
This hot-fix resolves #294. We had `lodash` as a devDependency, even though it's used in `src`. I also updated the import of `isEqual` in `filterutils.tsx` to follow best practices.

J=SLAP-2404
TEST=none

PR #321, opened 2022-11-01
Expose styling of facet headers in FilterGroup

tatimblin:
add env to gitignore

J=SLAP-2428
TEST=auto,manual

served site with custom label class added. toggled collapsible


src/components/Filters/CollapsibleLabel.tsx
```diff
@@ -22,15 +34,15 @@ export interface CollapsibleLabelProps {
...
+export function CollapsibleLabel({ label, customCssClasses }: CollapsibleLabelProps): JSX.Element {
   const { isExpanded, getToggleProps } = useFilterGroupContext();
   const iconClassName = classNames('w-3 text-gray-400', {
     'transform rotate-180': !isExpanded
   });
 
   return (
     <button className='w-full flex justify-between items-center mb-4' {...getToggleProps()}>
-      <div className='text-neutral-dark text-sm font-medium text-left'>
+      <div className={twMerge('text-neutral-dark text-sm font-medium text-left', customCssClasses?.label)}>
```
nmanu1, line 45: can we put the default styling in a `builtInCssClasses` variable like we do for other components? and we can use `useComposedCssClasses` instead of `twMerge` directly

nmanu1 (commented):
can the base branch of the PR be changed to develop?

src/components/FilterGroup.tsx
```diff
@@ -75,11 +78,15 @@ export function FilterGroup({
...
   }, [customCssClasses]);
 
+  const collapsibleLabelCssClasses: CollapsibleLabelCssClasses = {
+    label: cssClasses.label
+  };
+
   function renderTitle() {
     return collapsible
-      ? <CollapsibleLabel label={title} />
+      ? <CollapsibleLabel label={title} customCssClasses={collapsibleLabelCssClasses} />
```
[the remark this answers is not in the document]
tatimblin: we could, but then CollapsibleLabel would be able to take in cssClasses for things that are not relevant to it. Only label will be used in `CollapsibleLabel`

```
export interface FilterGroupCssClasses {
  label?: string,
  searchInput?: string,
  optionsContainer?: string,
  option?: string,
  optionInput?: string,
  optionLabel?: string
}
```

src/components/FilterGroup.tsx
```diff
@@ -67,19 +74,23 @@ export function FilterGroup({
 }: PropsWithChildren<FilterGroupProps>) {
   const cssClasses = useMemo(() => {
     const { option, optionLabel, optionInput, ...remainingClasses } = customCssClasses;
-    return {
+    return useComposedCssClasses(builtInCssClasses, {
```
yen-tt, line 77: linter issue: React Hook "useComposedCssClasses" cannot be called inside a callback

src/components/FilterGroup.tsx
```diff
@@ -75,11 +83,15 @@ export function FilterGroup({
     };
   }, [customCssClasses]);
 
+  const collapsibleLabelCssClasses: CollapsibleLabelCssClasses = {
```
oshi97, line 86: we'd want to memoify this

src/components/Filters/CollapsibleLabel.tsx
oshi97: +2 on this!

src/components/FilterGroup.tsx
```diff
@@ -65,6 +72,7 @@ export function FilterGroup({
   showMoreLimit = filterOptions.length,
   children
 }: PropsWithChildren<FilterGroupProps>) {
+  customCssClasses = useComposedCssClasses(builtInCssClasses, customCssClasses);
```
nmanu1, line 75: generally, for all of our components, we want to call this as low down in the chain of components as possible (so in `CollapsibleLabel` rather than here). this way, the merging of CSS classes only happens once

tatimblin: It is already being merged in `CollapsibleLabel`. It was duplicated [here](https://github.com/yext/search-ui-react/blob/39a8996f330c873edf46fbba96208127c07c4b6a/src/components/FilterGroup.tsx#L94) for when `CollapsibleLabel` is disabled, so right now the merging is needed in `FilterGroup` as well. I think a solution to this would be that `CollapsibleLabel` should be able to disable it's collapsing behavior so it can be used to render the label regardless of if `collapsible` is set, but that is out of scope

src/components/FilterGroup.tsx
```diff
@@ -51,6 +54,10 @@ export interface FilterGroupProps {
   showMoreLimit?: number
 }
 
+const builtInCssClasses: Readonly<FilterGroupCssClasses> = {
+  titleLabel: 'text-neutral-dark text-sm font-medium text-left mb-4',
```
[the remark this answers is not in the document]
tatimblin: It does. It was being duplicated here for if `collapsible` is `false`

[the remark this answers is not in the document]
nmanu1: if there's a `builtInCssClasses` there too, could we reference that in this file instead of creating another `builtInCssClasses` specially for it here?

src/components/FilterGroup.tsx
nmanu1: to avoid the multiple merges, could we update it so `useComposedCssClasses` is used in `CollapsibleLabel`, not here, but `twMerge` is used directly for the title `div` in this file (when `CollapsibleLabel` is not being used)?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@yext/search-ui-react",
-  "version": "1.0.0",
+  "version": "1.0.1",
```
oshi97, line 3: We more or less use git flow and wouldn't want to bump the version here but in the hotfix or release branch

tatimblin: @nmanu1 could this have been picked up by a release you did? I think a Github action made this change

nmanu1: I think because this branch was based off of main instead of develop, this PR is now picking up the changes that were in main but haven't been merged into develop yet. this change comes from the hotfix made in #318

oshi97: Tristan you can probably just merge your dev/filtergroup-css branch with develop (you shouldn't need to do a rebase just a regular merge normally works)

src/components/FilterGroup.tsx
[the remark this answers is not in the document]
tatimblin: They are off by 1 so I need a tsMerge to add it back to just the div not CollapsibleLabel

src/components/FilterGroup.tsx
tatimblin: I updated to do this. Perhaps it should be imported directly instead of through index.ts to not expose it so far?

src/components/FilterGroup.tsx
```diff
@@ -75,11 +83,17 @@ export function FilterGroup({
...
+    };
+  }, [cssClasses]);
+
   function renderTitle() {
     return collapsible
-      ? <CollapsibleLabel label={title} />
+      ? <CollapsibleLabel label={title} customCssClasses={collapsibleLabelCssClasses} />
       : (title &&
-        <div className='text-neutral-dark text-sm font-medium text-left mb-4'>
+        <div className={twMerge('mb-4', builtInCssClasses.titleLabel, cssClasses.titleLabel)}>
```
nmanu1, line 96: why is there an extra `mb-4` class here that isn't in the CollapsibleLabel styling? can it be removed so that they are the same?

src/components/FilterGroup.tsx
```diff
@@ -51,6 +55,10 @@ export interface FilterGroupProps {
   showMoreLimit?: number
 }
 
+const builtInCssClasses: Readonly<FilterGroupCssClasses> = {
+  titleLabel: builtInCollapsibleLabelCssClasses.titleLabel,
```
nmanu1, line 59: why do we need to make a `builtInCssClasses` variable for just this styling? can we just refer to `builtInCollapsibleLabelCssClasses.titleLabel` in the one place it's used below (for the `div`)?

src/components/Filters/CollapsibleLabel.tsx
```diff
@@ -1,6 +1,17 @@
...
 import { useFilterGroupContext } from './FilterGroupContext';
+import { useComposedCssClasses } from '../../hooks/useComposedCssClasses';
+
+/**
+ * The CSS class interface for {@link CollapsibleLabelProps}.
+ *
+ * @public
+ */
+export interface CollapsibleLabelCssClasses {
+  button?: string,
```
nmanu1, line 12: I don't imagine many people would want to customize the styling for the collapse button. It might be better to leave it out of the interface and leave it hard-coded as it was before

src/components/Filters/CollapsibleLabel.tsx
```diff
@@ -1,6 +1,17 @@
...
+  titleLabel?: string
```
nmanu1, line 13: it feels a little weird for this to be named `titleLabel` in terms of the CSS classes for a collapsible label. for this component, it's simply it's label. it doesn't have the notion of it being a "title" label. can this just be called `label`?

src/components/FilterGroup.tsx
```diff
@@ -75,11 +83,17 @@ export function FilterGroup({
     };
   }, [customCssClasses]);
 
+  const collapsibleLabelCssClasses = useMemo((): CollapsibleLabelCssClasses => {
```
nmanu1, line 86: nit: can this be simplified a tiny bit to the following?
```ts
const collapsibleLabelCssClasses: CollapsibleLabelCssClasses = useMemo(() => ({
  titleLabel: cssClasses.titleLabel
}), [cssClasses]);
```

src/components/FilterGroup.tsx
tatimblin: it was off by one class because `mb-4` is on a wrapping class on `CollapsibleLabel`, that the non-collapsible version doesn't have

oshi97:
could you resolve the merge conflict?

src/components/FilterGroup.tsx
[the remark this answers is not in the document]
tatimblin: closing because discussion moved [here](https://github.com/yext/search-ui-react/pull/321#discussion_r1012319966)

src/components/FilterGroup.tsx
```diff
@@ -75,11 +79,17 @@ export function FilterGroup({
...
+    };
+  }, [cssClasses]);
+
   function renderTitle() {
     return collapsible
-      ? <CollapsibleLabel label={title} />
+      ? <CollapsibleLabel label={title} customCssClasses={collapsibleLabelCssClasses} />
       : (title &&
-        <div className='text-neutral-dark text-sm font-medium text-left mb-4'>
+        <div className={twMerge('mb-4', builtInCollapsibleLabelCssClasses.label, cssClasses.titleLabel)}>
```
yen-tt, line 92: you may want this `twMerge` in a useMemo as well then pass that memoized value to className

src/components/FilterGroup.tsx
yen-tt: are you referring to how `builtInCssClasses` in `CollapsibleLabel ` should be expose? We currently do restrict what is exported out of Filters/index.ts from the top level components folder (in code: https://github.com/yext/search-ui-react/blob/develop/src/components/index.ts#L90-L94) so I think it should be fine leaving `builtInCssClasses` exported through Filters/index.ts and import it like how it's now

src/components/FilterGroup.tsx
oshi97: could use `useComposedCssClasses` instead of directly calling twMerge

src/components/FilterGroup.tsx
tatimblin: That sounds good

yen-tt (approved):

src/components/FilterGroup.tsx
```diff
@@ -75,11 +79,21 @@ export function FilterGroup({
     };
   }, [customCssClasses]);
 
+  const replicateCollapsibleLabelCssClasses = useMemo(() => {
```
oshi97, line 82: We shouldn't need two useMemo usages here. I feel like we could call useComposedCssClasses here?

src/components/FilterGroup.tsx
```diff
@@ -75,11 +79,21 @@ export function FilterGroup({
...
+    return twMerge('mb-4', builtInCollapsibleLabelCssClasses.label, cssClasses.titleLabel);
```
oshi97, line 83: Is it necessary to call twMerge here? I took a peak and CollapsibleLabel is using the classNames package instead of useComposedCssClasses (which uses twMerge). Is it possible to update CollapsibleLabel with useComposedCssClasses and then not have to import builtInCollapsibleLabelCssClasses here?

tatimblin: `builtInCollapsibleLabelCssClasses` is needed because that is an export from CollapsibleLabel. This helps replicate the title labels style wether collapsible or not

src/components/FilterGroup.tsx
tatimblin: Ok I am now doing the merge directly on the className

src/components/FilterGroup.tsx
```diff
@@ -75,11 +79,17 @@ export function FilterGroup({
...
+    };
+  }, [cssClasses]);
+
   function renderTitle() {
     return collapsible
-      ? <CollapsibleLabel label={title} />
+      ? <CollapsibleLabel label={title} customCssClasses={collapsibleLabelCssClasses} />
       : (title &&
-        <div className='text-neutral-dark text-sm font-medium text-left mb-4'>
+        <div className={twMerge('mb-4', builtInCollapsibleLabelCssClasses.label, collapsibleLabelCssClasses.label)}>
```
oshi97, line 92: I was thinking we could put this in the useMemo like `label: twMerge('mb-4', builtInCollapsibleLabelCssClasses.label, collapsibleLabelCssClasses.label)` and then have this be `className={collapsiblelabelCssClasses.label}` but it's ok

oshi97 (approved):

src/components/FilterGroup.tsx
tatimblin: `builtInCollapsibleLabelCssClasses` comes from `CollapsibleLabel` so I don't want to merge it back into `collapsibleLabelCssClasses` because then it'll merge in twice

PR #322, opened 2022-11-01
Export CSS bundle without Tailwind preflight

nmanu1:
Add an export for a CSS bundle without the Tailwind preflight CSS resets. While testing, I noticed that some of our components rely on Tailwind's default border color, which is set in the CSS resets. For users who use the new bundle and specify the Tailwind resets as custom resets, the default border color isn't found because they aren't using Tailwind. I updated the components with borders that were missing a border-color so the styling is consistent regardless of what CSS resets (if any) a user provides. I also updated the default border color in the Storybook and test-site Tailwind configs to match what a no-resets user would see if we forgot to include a border-color for a border in the future.

J=SLAP-2418
TEST=manual

Remove Tailwind from the test-site and instead import the new bundle without Tailwind preflight. Add the Tailwind preflight CSS resets to `test-site/src/index.css` and see that the components' styling matches how they look with Tailwind (or with the CSS bundle with preflight). Scope the Tailwind CSS resets with a class name and see that only parts of the app with that class name have the CSS resets applied, while the rest of the app is unaffected.


package.json
```diff
@@ -21,7 +21,8 @@
       "import": "./lib/esm/index.js",
       "require": "./lib/commonjs/index.js"
     },
-    "./bundle.css": "./lib/bundle.css"
+    "./bundle.css": "./lib/bundle.css",
+    "./bundle-no-preflight.css": "./lib/bundle-no-preflight.css"
```
tatimblin, line 25: maybe `bundle-no-tailwind` or `bundle-no-resets` I imagine if you don't use tailwind you probably don't know what preflight is

tatimblin (approved):

package.json
oshi97, line 25: Is it going to be hard to maintain the promise that the non-preflight css bundle looks as expected?

src/components/AppliedFilters.tsx
```diff
@@ -24,8 +24,8 @@ export const builtInCssClasses: Readonly<AppliedFiltersCssClasses> = {
   // Use negative margin to remove space above the filters on mobile
   appliedFiltersContainer: 'flex flex-wrap -mt-3 md:mt-0 mb-2',
   appliedFiltersLoading: 'opacity-50',
-  nlpFilter: 'border rounded-3xl px-3 py-1.5 text-sm font-medium text-neutral-dark mr-2 mb-2',
-  removableFilter: 'flex items-center border rounded-3xl px-3 py-1.5 text-sm font-medium text-neutral-dark mr-2 mb-2',
+  nlpFilter: 'border border-gray-200 rounded-3xl px-3 py-1.5 text-sm font-medium text-neutral-dark mr-2 mb-2',
```
oshi97, line 27: do we need to take care of margin differences with preflight/no preflight?

package.json
nmanu1: The non-preflight bundle definitely looks off, so users will need to supply resets on their end to have the components look as expected. It will be on them to update those resets as needed to keep the styling consistent

src/components/AppliedFilters.tsx
nmanu1: the margins can be easily reset from the user's side if they supply custom resets similar to Tailwind preflight. the reason the border needed to be updated here is because our components were inconsistent in defining a color for the border. Tailwind has a default color for borders that we were sometimes relying on, and sometimes not. so, this update just makes that consistent so that if a user sets a custom reset for the border color, it doesn't change the border color for some of our components but not others. for margins, there is no "default", so we're always specifying the exact value in our components

package.json
nmanu1: updated to `bundle-no-resets`

oshi97: In terms of maintaining the library though, I'm a little concerned that we'll add  a new component and forget to specify border in this specific way. I wouldn't want to get random techops or github issues in the future and play whack-a-mole with missing border issues. Do you have any ideas for preventing these kinds of issues?

nmanu1: ah I see what you're saying. I updated the Tailwind configs for Storybook and the test-site so we'll get an indication if we missed specifying a border-color. with the default border color changed to black, the styling that we see when testing will match what users will see if they use the no-resets bundle with the Tailwind preflight resets. so if we don't want a black border, we'll know to add a border-color when looking at the component in either Storybook or the test-site

oshi97: ah ok cool!

oshi97 (approved):

package.json
oshi97: lot more elegant than say taking a separate set of snapshots

PR #323, opened 2022-11-02
Add onSelect prop to FilterSearch

nmanu1:
Add an `onSelect` prop to `FilterSearch` and deprecate `searchOnSelect`. If an `onSelect` function is passed in, the default selecting (and searching) logic is bypassed and we call the `onSelect`. If `searchOnSelect=true` is passed with an `onSelect`, we log a console warning that the former will be ignored. To allow for the UCSD use case mentioned in the spec, `currentFilter` was updated to allow any `StaticFilter`, not just a field value filter.

J=SLAP-2431
TEST=auto, manual

See that the added Jest tests pass. In the test-site, pass in an `onSelect` function that matches the current functionality with `searchOnSelect` as either true or false. Also, mimic a UCSD-like use case with compound filters and see that the component behaves correctly when selecting and un-selecting a filter. If both an `onSelect` and `searchOnSelect=true` are passed, see that a console warning is logged.


src/components/FilterSearch.tsx
```diff
@@ -51,6 +51,15 @@ export interface FilterSearchProps {
   placeholder?: string,
   /** Whether to trigger a search when an option is selected. Defaults to false. */
   searchOnSelect?: boolean,
+  /** A function which is called when a filter is selected. To use this, searchOnSelect must be false. */
```
tatimblin, line 54: can `searchOnSelect` be undefined which is falsey or does it have to be `false`? If falsey is ok, can you change to "_To use this, searchOnSelect must not be true_"

src/components/FilterSearch.tsx
```diff
@@ -119,25 +125,35 @@ export function FilterSearch({
...
-        searchActions.setFilterOption({ filter: newFilter, displayName: newDisplayName, selected: true
-        });
-        setCurrentFilter(newFilter);
-        setFilterQuery(newDisplayName);
-        executeFilterSearch(newDisplayName);
-        if (searchOnSelect) {
-          searchActions.setOffset(0);
-          searchActions.resetFacets();
-          executeSearch(searchActions);
+        if (onSelect && !searchOnSelect) {
```
tatimblin, line 128: This is getting a little messy, can we move to it's own function `handleSelect` and use gates like

```
if (!newFilter && !newDisplayName) {
  return;
}

if (!select) {
  return;
}

if (onSelect && !searchOnSelect) {
  onSelect(...)
} else {
  ...
}
```

src/utils/filterutils.tsx
```diff
@@ -43,6 +43,26 @@ export function isDuplicateFieldValueFilter(
   return true;
 }
 
+/**
+ * Returns true if the two given static filters are the same.
```
tatimblin, line 47: What are cases that this happens? Is it for checking if two variables reference the same filter or for catching when the same field is being filtered on by two separate instances of the StaticFilter?

nmanu1: This is for the former (two variables referencing the same filter)

src/components/FilterSearch.tsx
```diff
@@ -51,6 +51,15 @@ export interface FilterSearchProps {
   placeholder?: string,
   /** Whether to trigger a search when an option is selected. Defaults to false. */
   searchOnSelect?: boolean,
+  /** A function which is called when a filter is selected. To use this, searchOnSelect must be false. */
+  onSelect?: (
+    currentFilter: StaticFilter | undefined,
+    setCurrentFilter: (filter: StaticFilter) => void,
```
oshi97, line 57: are all of these params necessary?

src/components/FilterSearch.tsx
oshi97, line 128: It feels weird that onSelect is only called when searchOnSelect is false. You are still *selecting* something even if searchOnSelect is true right?

oshi97: I also think some refactoring here would be good since we're adding additional complexity in this PR!

nmanu1: I think the idea was that if `searchOnSelect` is true, then the thing you want happening on select is the default "search" behavior. but, I talked with Davish about it again because the spec doesn't mention that (only the item description does), and I'm going to update to have both be allowed and independent as per his clarification

tmeyer2115: The item description was based on an earlier conversation with Davish. We decided that `onSelect` would only be invoked if `searchOnSelect` was false. In an ideal world, I think there wouldn't be both `onSelect` and `searchOnSelect`. Instead, `onSelect` would be optional and the default behavior would be to fire off a search. Having both, and allowing them to be independent, feels odd to me. It's two props that allow you to dictate the behavior on select. 

I think the only way I'd feel good about having both and allowing them to be independent is if we marked `searchOnSelect` as deprecated. Internally, we'd refactor the Component so the default `onSelect` would be firing off a search. @oshi97 and @nmanu1 what do you think of that? We'd need to run it by Product too.

nmanu1: I originally was also thinking of `onSelect` and `searchOnSelect` as both dictating how the selecting behavior should work. But, after thinking about it more and talking about it with Davish based on Oliver's comment, it made sense to me to have them be independent. Because `onSelect` dictates whether to use the default behavior or custom behavior when a filter is selected. But, `searchOnSelect` determines whether a searching action occurs or not after the filter selecting logic (whether it's default or custom) occurs. The user may want both a search to occur after a filter is selected as well as custom logic that happens when a filter is selected. In this case, a better name for `searchOnSelect` may be `searchAfterSelect`, although that may be confused with a boolean controlling the ordering of the searching and selecting actions. Does that make sense?

oshi97: My first intuition is that `onSelect` isn't related to whether or not FilterSearch runs a searchOnSelect. I think this was due to naming. If I saw the prop `onSelect` on FilterSearch I wouldn't *assume* that setting it would override the default behavior. 

If we want `onSelect` to override the default handler then maybe `handleSelect` would be a better name. In my brain, in general you can always have any amount of "onSomeEvent" listeners, but "handleSomeEvent" feels more like "this is the single designated handler". I'm probably splitting hairs.

Thinking about it a little more, though, having `handleSelect` override the default behavior (no `searchOnSelect`) lets us have less configuration. 
I think this is the ideal state.
Not because this is the most semantically clear, because that seems very subjective, but because that leads to the least configuration options (the least "entropy"). 

In order to prepare for that state, we'd want the props right now to behave in a way where we could remove `searchOnSelect` and no code would break. Right now, if we had `onSelect` NOT override the default behavior, in a even v2 we would have to tell people to change their `onSelect` handlers to also run a search at the start (or the end, this ordering business is also another reason we should stay away from `searchOnSelect`). Even though in a v2 we **can** make breaking changes, we shouldn't if we don't need to.

So anyways long story short after reading Tom's comment I'm kind of thinking we should have it override the default behavior? So we can mark `searchOnSelect` as deprecated, and if you have `handleSelect` set to something, that effectively sets `searchOnSelect` to false with higher priority.

nmanu1: I thought about it some more and looked at the `SearchBar` component for comparison, I'm starting to agree with this approach. It makes sense that we would deprecate `searchOnSelect` because `onSelect` would allow the user to add that functionality if they wanted (although it would be harder than using `searchOnSelect`). I agree with Oliver that a better name for it though might `handleSelect` as opposed to `onSelect` since we're replacing the default behavior. But, right now in the `SearchBar`, we have `onSearch` and `onSelectVerticalLink` props that also overrides the default behavior instead of adding on to it. In that sense, `onSelect` is consistent.

Right now, the default behavior of `FilterSearch` is to have `searchOnSelect` as false. If we wanted to maintain that default and deprecate and eventually remove `searchOnSelect` (maybe in a v2), then eventually we would remove the code needed to conduct a search after selecting a filter. This code right now handles reseting the facets and offset before executing a search. My concern is that it may be unintuitive (and require some knowledge of business logic) for users to know that they would need to do those other things as part of executing a search to have it behave how they expect. If we're fine with changing the default behavior (i.e. most people want `searchOnSelect` to be true right now), then I think it makes sense to have default selecting and searching behavior (no `searchOnSelect` prop), which could be completely bypassed by custom logic in an `onSelect` prop.

To summarize:
- If product is fine with the change in the default behavior to have it include a search being executed, then I'll update the default value for `searchOnSelect` to true and mark it as deprecated.
- I'll keep the `onSelect` prop with the current naming (instead of `handleSelect`) unless we want to consider updating the naming in the `SearchBar` as well.
- I'll update `onSelect` to have higher priority than `searchOnSelect`. In other words, if `onSelect` is specified, it would completely bypass the selecting and searching logic, regardless of the value of `searchOnSelect`.
- If the user passes in an `onSelect` function and `searchOnSelect: true`, then we can log a console warning that the latter will be ignored.

Thoughts?

src/components/FilterSearch.tsx
nmanu1: yes, in order to be able to exactly replicate the current filter search behavior, a user would need access to all of these params

src/components/FilterSearch.tsx
```diff
@@ -51,6 +51,15 @@ export interface FilterSearchProps {
   placeholder?: string,
   /** Whether to trigger a search when an option is selected. Defaults to false. */
   searchOnSelect?: boolean,
+  /** A function which is called when a filter is selected. To use this, searchOnSelect must be false. */
+  onSelect?: (
+    currentFilter: StaticFilter | undefined,
```
yen-tt, line 56: So currentFilter was updated to allow any StaticFilter because the selected field value filter could transformed into a compound, conjunction filter for the UCSD case, is that correct?
I think it's confusing that the currently filter is of type `StaticFilter` and the newFilter is `FieldValueStaticFilter`. I would expect the current and new filter type to match.

src/components/FilterSearch.tsx
tmeyer2115: I think those action items sound good! I do prefer the naming of `handleSelect` versus `onSelect`. But, it's not much of a preference. I'd be fine with either route in terms of naming. @oshi97 what do you think?

oshi97: yeah that sgtm!

nmanu1: Heard back from Product and they're on board with deprecating the `searchOnSelect` prop now that we’re adding an `onSelect` prop. And, they said we can change the default value of `searchOnSelect` to true. I updated the PR accordingly

nmanu1: Did some refactoring to simplify things. Turns out, the `handleToggleDropdown` code was no longer needed because we're not doing special behavior when a user clicks out of filter search

src/components/FilterSearch.tsx
nmanu1: now that `searchOnSelect` is being marked as deprecated and it's value doesn't affect `onSelect`, I removed this line from the docs

src/components/FilterSearch.tsx
```diff
@@ -49,8 +49,20 @@ export interface FilterSearchProps {
...
-  /** Whether to trigger a search when an option is selected. Defaults to false. */
+  /**
+   * Whether to trigger a search when an option is selected. Defaults to false.
+   *
+   * @deprecated Use the `onSelect` prop instead.
+   */
   searchOnSelect?: boolean,
+  /** A function which is called when a filter is selected. */
+  onSelect?: (
+    currentFilter: StaticFilter | undefined,
```
oshi97, line 60: perhaps the params to onSelect can be an object instead? that way somebody can pick out just the params they need for what they want, without having to do have unused params like `onSelect(_unused1, _unused2, newFilter)`

src/components/FilterSearch.tsx
```diff
@@ -99,56 +108,55 @@ export function FilterSearch({
...
-        setFilterQuery(value);
-        executeFilterSearch(value);
+    if (!newFilter || !newDisplayName) {
+      return;
+    }
+
+    if (onSelect) {
+      if (searchOnSelect) {
+        console.warn('Both searchOnSelect and onSelect props were passed to the component.'
+        + ' Using onSelect instead of searchOnSelect as the latter is deprecated.');
```
oshi97, line 136: Is "using" the write tense?

src/components/FilterSearch.tsx
```diff
@@ -99,56 +108,55 @@ export function FilterSearch({
...
-        }
-      } else {
-        setFilterQuery(value);
-        executeFilterSearch(value);
+    if (!newFilter || !newDisplayName) {
+      return;
+    }
+
+    if (onSelect) {
+      if (searchOnSelect) {
```
oshi97, line 134: we can combine these if statements right?

src/components/FilterSearch.tsx
nmanu1: I was trying to indicate that both props were encountered, so the component is **using** `onSelect` instead of `searchOnSelect` in the present moment. so I think it is correct?

oshi97: Ah sorry I misread

src/components/FilterSearch.tsx
nmanu1: no, because we still want to run `onSelect` whether `searchOnSelect` is true or not. only the warning is logged if `searchOnSelect` is true

src/components/FilterSearch.tsx
nmanu1: I thought about it, but I can't really think of a time when people wouldn't be using all of the params to perform some kind of filter search selecting behavior. I think it would be rare if that happened and even then I think they would use most of the params. Because of that, I didn't think it was worth making an object for the params because most people would then have to add code to either destructure the object or refer to the params as `params.blah`. what do you think?

oshi97: By "add code to destructure the object" do you mean add curly braces around the function like `onSelect({ currentFilter })` instead of `onSelect(currentFilter)`? I feel like that's not an issue and from my experience most people prefer destructuring things. It's also more flexible long term. I think the only downside is that it's more annoying for people to rename the variables if they don't like our names (they can just do `onSelect({ currentFilter: myFilter })`, though)

I was imagining the most common use for onSelect being somebody wanting to customize "what happens when a new filter is selected". They might not even care about FilterSearch maintaining it's own "currentFilter" state. This was something Alex and I talked about when product first started asking for FilterSearch changes - Alex thought somebody might NOT want FilterSearch to deselect the previous filters that were selected.
Anyways for that specific use case somebody would ONLY care about "newFilter", so they would have to do `onSelect(_currentFilter, _setCurrentFilter, newFilter)`. Which again is really not a big deal but I definitely prefer destructuring. We could also reorder the params ourselves in a way that we think would be used more commonly, and I think I would probably put newFilter at the very start, but I don't want to presume I can think of every use case and what the ideal ordering would be. Destructuring also makes it easier to add in additional params later if we need to.

nmanu1: Got it. I forgot for a moment that they could just add the curly braces instead of needing `onSelect(params) { const { currentFilter } = params; ... }`. And destructuring would definitely be better in terms of flexibility and future scalability. I'll update to make an object!

oshi97 (approved):
nice!

PR #326, opened 2022-11-07
Address Snyk vulnerabilities

nmanu1:
Update the versions of `@xmldom/xmldom`, `babel-loader`, and `loader-utils` to address Snyk vulnerabilities.

J=none
TEST=none

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -54,7 +54,7 @@
         "@yext/search-headless-react": "^2.0.0",
         "axe-playwright": "^1.1.11",
         "babel-jest": "^27.0.6",
-        "babel-loader": "^8.2.3",
+        "babel-loader": "^8.3.0",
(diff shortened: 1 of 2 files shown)


tatimblin (commented):
Looks like from the chart on this page we meet all the requirements to go to version 9 https://www.npmjs.com/package/babel-loader

nmanu1:
> Looks like from the chart on this page we meet all the requirements to go to version 9 https://www.npmjs.com/package/babel-loader

I did some digging and the reason we had `babel-loader` as a dev dependency was because it was a peer dependency for one of the Storybook addons. In v6.5.0, the peer dependency was removed so we no longer need to include `babel-loader` in our package.json. I've removed it accordingly

tatimblin (approved):
nice find

PR #328, opened 2022-11-08
v1.0.2

nmanu1:
### Bug Fixes
- Address security vulnerabilities in `ejs`, `async`, and `terser` (#308)
- Address security vulnerabilities in `@xmldom/xmldom` and `loader-utils` (#320, #326)


oshi97 (approved):

yen-tt (commented):
as part of this fix, do you think we should update "@yext/analytics" dep to not be beta anymore?

tmeyer2115 (approved):

PR #329, opened 2022-11-08
Update `@yext/analytics` version

nmanu1:
Update the version of `@yext/analytics` from a beta version to v0.2.2.


yen-tt (commented):
it shouldn't have any breaking changes by bumping patch version, but since it was a beta and the major is still 0, did you double check the commits in yext/analytics if there were any potential changes that could affect the things we use in here?

nmanu1:
> it shouldn't have any breaking changes by bumping patch version, but since it was a beta and the major is still 0, did you double check the commits in yext/analytics if there were any potential changes that could affect the things we use in here?

there aren't any breaking changes that affect how the components in this repo report analytics events. but, after looking further, there were breaking interface changes in [this PR](https://github.com/yext/analytics/pull/21). so if someone used the `AnalyticsProvider` and `useAnalytics`, the service they got back would have a `report` method that took in optional `additionalRequestAttributes` of a re-named type after the version change. if people were referring to or extending the old interface, they would get TS errors.

also, when the pages analytics support was added to the analytics library, some of the interfaces were renamed (e.g. `AnalyticsService` -> `SearchAnalyticsService`). while, we did keep aliases to the original names, so technically no changes are needed in this repo, it might be good if we renamed to the Search-specific interface names

PR #331, opened 2022-11-09
Perceived performance improvement with Mapbox Static Images API

yen-tt:
Using the [Mapbox Static Images API](https://docs.mapbox.com/api/maps/#static-images) to improve the perceived performance for your users -- i.e immediately show a static image while Mapbox GL JS works in the background. From lighthouse analysis, this is generally similar to the performance before this change, occasionally a couple points increase from the norm.

Note: the API only support width and height up to 1280 pixels, and will response with 422 error code otherwise. We could attempt to stretch the image to cover the space (e.g. such as using background-cover css style) but it wouldn't match with the display of the actual map on load (without initial search) so this PR currently will not use static image API unless the dimension of the map is within the restriction.

J=SLAP-2434
TEST=manual

load location page and see that a static image is fetched over network and display until the canvas for the interactive map is loaded, then that become visible to the user instead.

### DIFF
--- a/src/components/MapboxMap.tsx
+++ b/src/components/MapboxMap.tsx
@@ -1,8 +1,9 @@
-import { useRef, useEffect } from 'react';
+import { useRef, useEffect, useMemo, useState } from 'react';
 import mapboxgl, { Map, Marker, MapboxOptions, LngLatBounds, MarkerOptions, LngLat } from 'mapbox-gl';
 import { Result, useSearchState } from '@yext/search-headless-react';
 import { useDebouncedFunction } from '../hooks/useDebouncedFunction';
 import ReactDOM from 'react-dom';
+import useLayoutEffect from 'use-isomorphic-layout-effect';
 
 /**
  * A functional component that can be used to render a custom marker on the map.
@@ -67,6 +68,11 @@ export interface MapboxMapProps<T> {
 }
 
 /**
+ * Defined mapbox options as defaults. These options are also needed for Mapbox Static Images API.
+ */
+type DefaultMapboxOptions = Required<Pick<MapboxOptions, 'center' | 'style' | 'zoom'>>;
+
+/**
  * A component that renders a map with markers to show result locations using Mapbox GL.
  *
  * @remarks
@@ -96,6 +102,7 @@ export function MapboxMap<T>({
     mapboxgl.accessToken = mapboxAccessToken;
   }, [mapboxAccessToken]);
 
+  const parentContainer = useRef<HTMLDivElement>(null);
   const mapContainer = useRef<HTMLDivElement>(null);
   const map = useRef<Map | null>(null);
   const markers = useRef<Marker[]>([]);
@@ -103,16 +110,22 @@ export function MapboxMap<T>({
   const locationResults = useSearchState(state => state.vertical.results) as Result<T>[];
   const onDragDebounced = useDebouncedFunction(onDrag, 100);
 
+  const options: Omit<MapboxOptions & DefaultMapboxOptions, 'container'> = useMemo(() => {
+    return {
+      style: 'mapbox://styles/mapbox/streets-v11?optimize=true',
+      center: [-74.005371, 40.741611],
+      zoom: 9,
+      ...mapboxOptions
+    };
+  }, [mapboxOptions]);
+  const mapboxStaticImageDiv = useMapboxStaticImage(mapboxAccessToken, parentContainer, options);
+
   useEffect(() => {
     if (mapContainer.current && !map.current) {
-      const options: MapboxOptions = {
+      map.current = new Map({
         container: mapContainer.current,
-        style: 'mapbox://styles/mapbox/streets-v11',
-        center: [-74.005371, 40.741611],
-        zoom: 9,
-        ...mapboxOptions
-      };
-      map.current = new Map(options);
+        ...options
+      });
       const mapbox = map.current;
       mapbox.resize();
       if (onDragDebounced) {
@@ -120,8 +133,11 @@ export function MapboxMap<T>({
           onDragDebounced(mapbox.getCenter(), mapbox.getBounds());
         });
       }
+      mapbox.on('load', () => {
+        mapbox.getContainer().style.visibility = 'visible';
+      });
     }
-  }, [mapboxOptions, onDragDebounced]);
+  }, [options, onDragDebounced]);
 
   useEffect(() => {
     markers.current.forEach(marker => marker.remove());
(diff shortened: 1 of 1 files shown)


tmeyer2115 (commented):
Would it be possible to add a Storybook story as well to test the loading state?

src/components/MapboxMap.tsx
```diff
@@ -66,6 +67,11 @@ export interface MapboxMapProps<T> {
   onDrag?: OnDragHandler
 }
 
+/**
+ * Defined mapbox options as defaults. These options are also needed for Mapbox Static Images API.
```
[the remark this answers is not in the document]
yen-tt: this interface is removed after some refactoring

src/components/MapboxMap.tsx
```diff
@@ -66,6 +67,11 @@ export interface MapboxMapProps<T> {
...
+ */
+type DefaultMapboxOptions = Required<Pick<MapboxOptions, 'center' | 'style' | 'zoom'>>;
```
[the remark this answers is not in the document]
yen-tt: The original intention was for the three fields here ('center' | 'style' | 'zoom') to always be defined as part of our default options so we can construct a valid mapbox static image url. This would help bypass the need for manual checking of potential undefined values since they would always be require for the default -- typescript would take care of it.

I don't feel too strongly about keeping it if you think this seems confusing. This interface is removed after some refactoring, with a new interface for the configuration for Mapbox Static Image API url, and add in manual checking to ensure they are defined.

src/components/MapboxMap.tsx
```diff
@@ -161,10 +177,53 @@ export function MapboxMap<T>({
...
+    }
+    const { width, height } = divDimension;
+    // Mapbox Static Image API only support width and height between 1-1280 pixels.
+    if (width > 1280 || height > 1280) {
+      return undefined;
+    }
+    const stylesheet = style.split('mapbox://styles/')[1].split('?')[0];
+    const centerAndZoom = `${center[0]},${center[1]},${zoom}`;
+    const dimension = `${width}x${height}`;
+    const staticMapboxUrl = `https://api.mapbox.com/styles/v1/${stylesheet}/static/${centerAndZoom}/${dimension}?access_token=${mapboxAccessToken}`;
```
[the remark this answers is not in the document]
yen-tt: sure!

src/components/MapboxMap.tsx
```diff
@@ -96,32 +97,42 @@ export function MapboxMap<T>({
...
+      zoom: 9,
+      ...mapboxOptions
+    };
+  }, [mapboxOptions]);
+  const mapboxStaticImageDiv = useMapboxStaticImage(mapboxAccessToken, parentContainer, options);
+
   useEffect(() => {
     if (mapContainer.current && !map.current) {
-      const options: MapboxOptions = {
+      map.current = new Map({
```
oshi97, line 120: super nit but maybe reordering stuff like `const mapbox = new Map(); map.current = mapbox`?

src/components/MapboxMap.tsx
```diff
@@ -161,10 +172,79 @@ export function MapboxMap<T>({
...
   return (
-    <div ref={mapContainer} className='h-full w-full' />
+    <div className='grid h-full w-full' ref={parentContainer}>
+      <div ref={mapContainer} className="col-span-full row-span-full invisible"/>
+      {mapboxStaticImageDiv}
+    </div>
   );
 }
 
+function useMapboxStaticImage(
```
oshi97, line 182: Instead of a hook that returns JSX would it make sense for this to just be a component?
It might make it easier to add a story for it as well then?

src/components/MapboxMap.tsx
```diff
@@ -161,10 +172,79 @@ export function MapboxMap<T>({
...
+  const { mapboxAccessToken, style, center, zoom, width, height } = urlConfig;
+  // Mapbox Static Image API only support width and height between 1-1280 pixels.
+  if (width > 1280 || height > 1280) {
+    return undefined;
+  }
+  const stylesheet = style.split('mapbox://styles/')[1].split('?')[0];
+  const centerAndZoom = `${center[0]},${center[1]},${zoom}`;
+  const dimension = `${width}x${height}`;
+  return `https://api.mapbox.com/styles/v1/${stylesheet}/static/${centerAndZoom}/${dimension}?access_token=${mapboxAccessToken}`;
+}
```
oshi97, line 246: I hate asking for more legal stuff but do we need to poke product about the static image api pricing? https://docs.mapbox.com/api/maps/static-images/#static-images-api-pricing
the first 50k are free which is nice at least https://www.mapbox.com/pricing/#static-images-api

src/components/MapboxMap.tsx
yen-tt: sure!

src/components/MapboxMap.tsx
yen-tt: per discussion during all hands, tom will run this by product and legal before the feature is merge into develop

src/components/MapboxStaticImage.tsx
```diff
@@ -0,0 +1,81 @@
...
+  useLayoutEffect(() => {
+    if (mapContainer.current) {
+      setDivDimension({
+        width: mapContainer.current.clientWidth,
+        height: mapContainer.current.clientHeight
+      });
+    }
+  }, []);
+
+  const staticMapboxStyle: React.HTMLAttributes<HTMLDivElement>['style'] = useMemo(() => {
```
oshi97, line 35: I think `React.HTMLAttributes<HTMLDivElement>['style']` can just be `CssProperties`?

oshi97 (approved):

tests/components/MapboxMap.stories.tsx
```diff
@@ -20,7 +21,13 @@ const meta: ComponentMeta<typeof MapboxMap> = {
       control: false,
     },
   },
-  parameters: { layout: 'fullscreen', percy: { enableJavascript: true } },
+  parameters: {
+    layout: 'fullscreen',
+    percy: {
+      enableJavascript: true,
+      waitForSelector: '.mapboxgl-map[style=\'visibility: visible;\']'
```
oshi97, line 28: nice!

oshi97 (approved):

tests/components/MapboxMap.stories.tsx
```diff
@@ -55,3 +60,18 @@ CustomPin.play = async ({ canvasElement }) => {
...
 };
+
+export const StaticImageBeforeLoad: Story<MapboxStaticImageProps> = (args) => (
+  <SearchHeadlessContext.Provider value={generateMockedHeadless(locationVerticalSingle)}>
+    <div className='grid h-full w-full'>
+      <MapboxStaticImage {...args} />
+    </div>
+  </SearchHeadlessContext.Provider>
+);
+StaticImageBeforeLoad.args = {
```
nmanu1, line 71: out of curiosity, since this component has different props/args than `MapboxMap`, does Storybook know to update the props in the controls/docs section for this story to match `MapboxStaticImageProps`? or does it still show the info from `MapboxMapProps`?

tests/components/MapboxMap.stories.tsx
```diff
@@ -20,7 +21,13 @@ const meta: ComponentMeta<typeof MapboxMap> = {
       control: false,
```
nmanu1, line 21: nit: can we also remove the controls for `getCoordinate` and `onDrag`? we don't want people to be able to pass in functions or components through the Storybook site

tests/components/MapboxMap.stories.tsx
yen-tt: it shows props from `MapboxMapProps`

nmanu1: in that case, what do you think about either separating out this story into another file so the controls match the correct props, or switching off the controls in this particular story for any props that don't overlap between the two interfaces? I'm just imagining it may be confusing if someone tries using the controls for `MapboxMap` props and it has no effect on the `MapboxStaticImage`

yen-tt: I do like it as part of the MapboxMap stories since it's only use as an internal component to MapboxMap (we don't export MapboxStaticImage. Showing the controls for MapboxMap make sense to me because that is what user will be expose, and if they provide their own mapbox key and style configuration in the control section, that will be use to construct MapboxStaticImage. If you think that's still confusing, I can disable/hide controls for that particular story.

nmanu1: oh ok, so the only controls that are configurable are ones that also affect `MapboxStaticImage`? in that case, I'm fine leaving it as is, especially since we're not exporting `MapboxStaticImage`

tests/components/MapboxMap.stories.tsx
yen-tt: decided to use the storybook play function to force await for the dynamic map instead

yen-tt: made an item for it: https://yexttest.atlassian.net/browse/SLAP-2462

nmanu1 (commented):
looks like the wcag tests complete but some of the stories are failing. the error is different from the one we were seeing before

.github/workflows/wcag_test.yml
```diff
@@ -6,8 +6,9 @@ on:
 
 jobs:
```
nmanu1, line 7: like you mentioned in your comment, could we update this to run on feature branches too?

yen-tt: yup! was gonna update it when I change back the workflow from dev branch to v1 in a bit

nmanu1 (commented):
looks like there's a wcag violation in one of the map stories that's causing the check to fail

src/components/MapboxStaticImage.tsx
```diff
@@ -0,0 +1,81 @@
+import { CSSProperties, useMemo, useRef, useState } from 'react';
+import { MapboxOptions, LngLatLike } from 'mapbox-gl';
+import useLayoutEffect from 'use-isomorphic-layout-effect';
+
+/**
+ * Props for {@link MapboxStaticImage}
```
nmanu1, line 6: nit: missing period at the end here and a few other places in this file

yen-tt:
> looks like there's a wcag violation in one of the map stories that's causing the check to fail

that is fixed in this PR: https://github.com/yext/search-ui-react/pull/337 for wcag related issues coming from the existing mapbox stories. will merge the change to this branch after.

nmanu1 (approved):

oshi97:
putting wip tag to exclude this from our PR dashboard (even though the PR is done)

PR #332, opened 2022-11-14
Mapbox component

yen-tt:
merge mapbox component feature to develop branch

PR #333, opened 2022-11-14
Update default FilterSearch behavior

nmanu1:
Update the default behavior of `FilterSearch`:
- On select, clobber other static filters in State that have a `fieldId` that matches with one of the search fields for `FilterSearch`. If there are multiple filters that were clobbered, log a console warning.
- If the static filters state is updated from outside of the component, log a console warning if there are multiple "matching" filters in State and there is no custom `onSelect` prop passed in.
  - And if there is no filter currently associated with the component, update the input text to the display name of the first "matching" filter that is selected in State.
    - If there are no matching filters in State, clear out the input and filter search response.

Note: We only look for field value filters in State for the "matching" filters. We don't prescribe how compound filters should be handled when comparing one to a search field. In the UCSD use case, a developer would pass in a custom `onSelect` function that would set compound static filters in State. In this case, the `currentFilter` may not be part of the `matchingFilters` array, which is why the concept of a `currentFilter` is still needed.

J=SLAP-2432
TEST=auto, manual

See that the added Jest tests pass. Spin up the test-site and test that the above situations match the expected behavior with different combinations of filters in State and passing an `onSelect` prop or not.

### DIFF
--- a/src/components/FilterSearch.tsx
+++ b/src/components/FilterSearch.tsx
@@ -111,7 +111,9 @@ export function FilterSearch({
   const staticFilters = useSearchState(state => state.filters.static);
   const matchingFilters: SelectableStaticFilter[] = useMemo(() => {
     return staticFilters?.filter(({ filter, selected }) =>
-      selected && filter.kind === 'fieldValue' && searchFields.some(s => s.fieldApiName === filter.fieldId)
+      selected
+      && filter.kind === 'fieldValue'
+      && searchFields.some(s => s.fieldApiName === filter.fieldId)
     ) ?? [];
   }, [staticFilters, searchFields]);
 
@@ -129,8 +131,10 @@ export function FilterSearch({
 
   useEffect(() => {
     if (matchingFilters.length > 1 && !onSelect) {
-      console.warn('More than one selected static filter found that matches the filter search fields.'
-        + ' Please update the state to remove the extra filters.');
+      console.warn('More than one selected static filter found that matches the filter search fields: ['
+        + searchFields.map(s => s.fieldApiName).join(', ')
+        + ']. Please update the state to remove the extra filters.'
+        + ' Picking one filter to display in the input.');
     }
 
     if (currentFilter && staticFilters?.find(f =>
@@ -153,7 +157,8 @@ export function FilterSearch({
     staticFilters,
     executeFilterSearch,
     onSelect,
-    matchingFilters
+    matchingFilters,
+    searchFields
   ]);
 
   const sections = useMemo(() => {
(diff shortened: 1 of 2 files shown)


src/components/FilterSearch.tsx
```diff
@@ -109,6 +109,11 @@ export function FilterSearch({
   const [currentFilter, setCurrentFilter] = useState<StaticFilter>();
   const [filterQuery, setFilterQuery] = useState<string>();
   const staticFilters = useSearchState(state => state.filters.static);
+  const matchingFilters: SelectableStaticFilter[] = useMemo(() => {
+    return staticFilters?.filter(({ filter, selected }) =>
+      selected && filter.kind === 'fieldValue' && searchFields.some(s => s.fieldApiName === filter.fieldId)
```
tatimblin, line 114: nit can you put on their own line
```suggestion
      selected
      && filter.kind === 'fieldValue'
      && searchFields.some(s => s.fieldApiName === filter.fieldId)
```

src/components/FilterSearch.tsx
```diff
@@ -123,14 +128,33 @@ export function FilterSearch({
   );
 
   useEffect(() => {
+    if (matchingFilters.length > 1 && !onSelect) {
+      console.warn('More than one selected static filter found that matches the filter search fields.'
+        + ' Please update the state to remove the extra filters.');
```
tatimblin, line 133: what does the concatenation do vs a single string?

nmanu1: This is just to split up the line so it doesn't become too long and exceed the max line length (linting error)

src/components/FilterSearch.tsx
```diff
@@ -123,14 +128,33 @@ export function FilterSearch({
   );
 
   useEffect(() => {
+    if (matchingFilters.length > 1 && !onSelect) {
```
yen-tt, line 131: I feel like it may be beneficial to warn user regardless if onSelect is provided here?

src/components/FilterSearch.tsx
```diff
@@ -123,14 +128,33 @@ export function FilterSearch({
...
+      console.warn('More than one selected static filter found that matches the filter search fields.'
```
yen-tt, line 132: nit: can we add to the message that we will use the first selected static filter found?

src/components/FilterSearch.tsx
nmanu1: this was a deliberate decision because the thinking is that if you provide your own `onSelect` prop, you might be fine with having multiple filters selected in state at once and you wouldn't want this warning to to be logged. it's only assumed to be a problem for the default component behavior. people shouldn't have to fork the component just to remove this warning if they don't want clobbering behavior

src/components/FilterSearch.tsx
nmanu1: what do you think about adding "Picking one filter to display in the input." instead? because technically, it will only pick the first filter it finds if the `currentFilter` is not selected in state anymore. or does that seem too vague?

yen-tt (approved):

nmanu1:
The failing WCAG and coverage checks are from PR #332. As mentioned there, an [item](https://yexttest.atlassian.net/browse/SLAP-2458) was made to address it.

PR #336, opened 2022-11-16
Add Geolocation component

yen-tt:
This PR adds Geolocation component to the repo, with jest tests and storybook stories.

Note: waiting on product to get SVG design for the default geolocation icon. Edit: Will update in another PR

J=SLAP-2396,SLAP-2397,SLAP-2398
TEST=manual&auto

- tested manually through test site in Products page
- jest tests passed and storybook stories display component as expected

### DIFF
--- a/src/hooks/useGeolocationHandler.ts
+++ b/src/hooks/useGeolocationHandler.ts
@@ -1,4 +1,3 @@
-
 import { Matcher, SelectableStaticFilter, useSearchActions, useSearchState } from '@yext/search-headless-react';
 import { executeSearch } from '../utils/search-operations';
 import { getUserLocation } from '../utils/location-operations';


src/components/Geolocation.tsx
```diff
@@ -0,0 +1,124 @@
+import { Matcher, SelectableStaticFilter, useSearchActions, useSearchState } from '@yext/search-headless-react';
+import { executeSearch } from '../utils/search-operations';
+import { getUserLocation } from '../utils/location-operations';
+import { useComposedCssClasses } from '../hooks/useComposedCssClasses';
+import { useCallback, useState } from 'react';
+import LoadingIndicator from '../icons/LoadingIndicator';
+import { YextIcon } from '../icons/YextIcon';
+
+/**
+ * The CSS class interface for the Geolocation component.
```
yen-tt, line 10: I cannot use `{@link Geolocation}` here due to an issue with API Extractor adding `_2` suffix to duplicate interface/class names. In search-ui-react, [Geolocation from Typescript](https://microsoft.github.io/PowerBI-JavaScript/interfaces/_node_modules_typedoc_node_modules_typescript_lib_lib_dom_d_.geolocation.html)'s "lib.dom.d.ts" file conflicts with Geolocation component.
See: https://github.com/microsoft/rushstack/issues/2534 and https://github.com/microsoft/rushstack/issues/2976
Open PR: https://github.com/microsoft/rushstack/pull/2608

Screen Shot 2022-11-16 at 12 15 08 PM

What to know if the team have any thoughts on this. A work around is to add a script to manually replace naming in the files generated from API extractor, which is hacky.

src/components/Geolocation.tsx
```diff
@@ -0,0 +1,124 @@
...
+      console.warn(e);
+    } finally {
+      setIsFetchingLocation(false);
+    }
+  }, [geolocationOptions, onClick, radius, searchActions, staticFilters]);
+
+  return (
+    <div className={cssClasses.geolocationContainer}>
+      <button className={cssClasses.button} onClick={handleGeolocationClick}>
+        {label}
```
oshi97, line 117: Should this button still be clickable after we've gotten the user's location?

src/components/Geolocation.tsx
```diff
@@ -0,0 +1,124 @@
...
+  geolocationOptions?: PositionOptions,
+  /**
+   * The radius, in miles, around the user's location to find results. Defaults to 50.
+   * If location accuracy is low, a larger radius may be used automatically.
+   */
+  radius?: number,
+  /** The label for the button. Defaults to 'Use my location'. */
+  label?: string,
+  /** Custom icon component to display along with the button. */
+  GeolocationIcon?: React.FC,
```
oshi97, line 45: not 100% sure but I think React.FunctionalComponent is preferred over React.FC

src/components/Geolocation.tsx
```diff
@@ -0,0 +1,124 @@
...
+        }
+      };
+      const nonLocationFilters = staticFilters?.filter(filter => {
+        return !(filter.filter.kind === 'fieldValue'
+          && filter.filter.fieldId === LOCATION_FIELD_ID);
+      }) ?? [];
+      searchActions.setStaticFilters([...nonLocationFilters, locationFilter]);
+      executeSearch(searchActions);
+    } catch (e) {
+      console.warn(e);
```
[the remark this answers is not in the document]
yen-tt: Product was fine with a console warn and didn't have any UX behavior in mind

src/components/Geolocation.tsx
yen-tt: I think it's okay for user to still click it and update their location after if it potentially change somehow

src/components/Geolocation.tsx
yen-tt: sure! I can update

src/components/Geolocation.tsx
tatimblin: They could move since the last search

src/components/Geolocation.tsx
```diff
@@ -0,0 +1,124 @@
...
+  onClick,
+  customCssClasses,
+}: GeolocationProps): JSX.Element | null {
+  const searchActions = useSearchActions();
+  const staticFilters = useSearchState(s => s.filters.static || []);
+  const [isFetchingLocation, setIsFetchingLocation] = useState<boolean>(false);
+  const cssClasses = useComposedCssClasses(builtInCssClasses, customCssClasses);
+
+  const handleGeolocationClick = useCallback(async () => {
+    onClick?.();
```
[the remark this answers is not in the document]
yen-tt: product was thinking this prop would be use in case user want to fire an analytics event for example. I agree with your comment, I can bring this concern up to them!

src/components/Geolocation.tsx
oshi97: Ah ok either way is fine then, I think I read some variant of this blog post a while back https://spin.atomicobject.com/2022/01/04/think-twice-react-fc/ but there's a lot of hot takes on the internet

oshi97: Reading it again I think the React.FC they're talking about (from the create-react-app template) is different than this one

src/components/Geolocation.tsx
oshi97: I was thinking about mirroring what LocationBias currently does but that makes sense

.storybook/preview.js
```diff
@@ -36,6 +36,7 @@ export const parameters = {
         'AlternativeVerticals',
         'SpellCheck',
         'ResultsCount',
+        'Geolocation',
         'LocationBias',
```
oshi97, line 40: will we mark LocationBias as deprecated in this pr?

yen-tt: oh I didn't pick up that item, it's gonna be in another pr

src/components/Geolocation.tsx
yen-tt: LocationBias still lets you click on the button after we've gotten the user's location

tests/components/Geolocation.test.tsx
```diff
@@ -0,0 +1,230 @@
...
+  const actions = spyOnActions();
+  render(<Geolocation />);
+  clickUpdateLocation();
+
+  await waitFor(() => {
+    expect(actions.executeVerticalQuery).toBeCalled();
+  });
+});
+
+it('handles erorr when collecting user\'s location', async () => {
```
nmanu1, line 197: nit: erorr

tests/components/Geolocation.stories.tsx
```diff
@@ -0,0 +1,45 @@
...
+import { VerticalSearcherState } from '../__fixtures__/headless-state';
+import { userEvent, within } from '@storybook/testing-library';
+import { Geolocation, GeolocationProps } from '../../src/components/Geolocation';
+
+const meta: ComponentMeta<typeof Geolocation> = {
+  title: 'Geolocation',
+  component: Geolocation,
+  argTypes: {
+    geolocationOptions: {
+      control: false
```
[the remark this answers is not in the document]
yen-tt: we mock location-operations utils for our stories, which include mocking the result of getUserLocation, so this option `geolocationOptions` isn't going to change anything. This is the same for our LocationBias component.

src/components/Geolocation.tsx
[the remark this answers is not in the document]
yen-tt: updated based on conversation on slack

.storybook/preview.js
oshi97: oh I see gotcha didn't realize we had another item

src/components/Geolocation.tsx
```diff
@@ -0,0 +1,124 @@
...
+            lat: latitude,
+            lng: longitude,
+            radius: Math.max(accuracy, radius * METERS_PER_MILE)
+          },
+        }
+      };
+      const nonLocationFilters = staticFilters.filter(filter => {
+        return !(filter.filter.kind === 'fieldValue'
+          && filter.filter.fieldId === LOCATION_FIELD_ID);
+      }) ?? [];
```
nmanu1, line 104: nit: is this nullish coalescing operator needed with the empty array now that the optional chaining is removed from `staticFilters`?

src/components/Geolocation.tsx
oshi97: Ohh the behavior I was remember was that if the user clicks "block" when the location request prompt pops up the button will disappear. It looks like this only happens in the SDK's LocationBias component and not the react one, which is fine.

src/components/Geolocation.tsx
oshi97: I'm fine with it, the api extractor docs aren't made to be extremely beginner friendly

nmanu1: could we link to `Geolocation_2`? or would that not resolve correctly?

yen-tt: it would link correctly in the doc pages but we would see this in search-ui-react.api.md file:
Screen Shot 2022-11-16 at 3 28 29 PM

oshi97:
is there something up with the code coverage github action? it's not passing but haven't dug too deep

src/components/Geolocation.tsx
nmanu1: oh ok, I'm fine leaving it without a link

src/components/Geolocation.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+  const handleGeolocationClick = useCallback(async () => {
+    setIsFetchingLocation(true);
+    try {
+      const position = await getUserLocation(geolocationOptions);
+      if (handleClick) {
+        handleClick(position);
+        return;
+      }
+      const { latitude, longitude, accuracy } = position.coords;
```
tmeyer2115, line 86: Internally, in the Component, could we always have a `handleClick`? The setting of the static filters and firing of a search would be the default `handleClick`. I think if we extract this code block into a default `handleClick` things will read a bit cleaner.

tests/components/Geolocation.test.tsx
```diff
@@ -0,0 +1,230 @@
...
+  const actions = spyOnActions();
+  render(<Geolocation handleClick={mockedHandleClickFn} />);
+  clickUpdateLocation();
+  await waitFor(() => {
+    expect(mockedHandleClickFn).toHaveBeenCalledWith(newGeoPosition);
+  });
+  expect(actions.executeVerticalQuery).not.toBeCalled();
+});
+
+it('sets a location filter with user\'s coordinates in static filters state when clicked', async () => {
```
tmeyer2115, line 127: There are a lot of tests in this file. Could we organize them with some `describe` blocks? For instance, this test and many of the ones below assume there is no `handleClick`. But, the test names don't indicate that. We could provide that additional context with a `describe` block reading `using default click handling` or something.

yen-tt:
> is there something up with the code coverage github action? it's not passing but haven't investigated why

code coverage has been failing for the past couple PRs, I believe there's an issue with the visual-coverage script killing the process before it can run the rest of the commands for generating lcov coverage and moving files to the expected directory. That will be fix in another item/PR.

src/components/Geolocation.tsx
```diff
@@ -0,0 +1,124 @@
...
+  GeolocationIcon = YextIcon,
+  handleClick,
+  customCssClasses,
+}: GeolocationProps): JSX.Element | null {
+  const searchActions = useSearchActions();
+  const staticFilters = useSearchState(s => s.filters.static || []);
+  const [isFetchingLocation, setIsFetchingLocation] = useState<boolean>(false);
+  const cssClasses = useComposedCssClasses(builtInCssClasses, customCssClasses);
+
+  const handleGeolocationClick = useCallback(async () => {
```
oshi97, line 105: This change would be mostly aesthetic in nature, but I kind of like moving handleGeolocationClick into its own hook like `useGeolocationHandler`, since we could also move useSearchActions and useSearchState into that hook to make it clear it's ONLY being used in the handler.

```ts
function useGeolocationHandler() {
  const searchActions = useSearchActions();
  const staticFilters = useSearchState(s => s.filters.static || []);
  // blah
  return handlerFunction

}
```

yen-tt: sure, updated

src/hooks/useGeolocationHandler.ts
```diff
@@ -0,0 +1,82 @@
...
+  handleUserPosition?: (position: GeolocationPosition) => void
+}
+
+/**
+ * Returns a function that will collect user's geolocation and, by default, will set
+ * a built-in location filter and execute a search.
+ *
+ * @internal
+ *
+ * @param props - {@link GeolocationProps}
```
tmeyer2115, line 35: Update the prop type `@link`.

src/hooks/useGeolocationHandler.ts
```diff
@@ -0,0 +1,82 @@
...
+  setIsFetchingLocation: React.Dispatch<React.SetStateAction<boolean>>,
+  /** Configuration used when collecting the user's location. */
+  geolocationOptions?: PositionOptions,
+  /**
+   * The radius, in miles, around the user's location to find results. Defaults to 50.
+   * If location accuracy is low, a larger radius may be used automatically.
+   */
+  radius?: number,
+  /** Custom handler function to call after user's position is successfully determined. */
+  handleUserPosition?: (position: GeolocationPosition) => void
```
tmeyer2115, line 26: Small nit: I'm not wild about the names `handleUserPosition` or `setIsFetchingLocation`. I think the latter is a bit long and convoluted. The former is a bit vague: `handleUserPosition` doesn't sound like a callback that fires when the user's position is obtained.

src/hooks/useGeolocationHandler.ts
```diff
@@ -0,0 +1,82 @@
...
+
+const LOCATION_FIELD_ID = 'builtin.location';
+const METERS_PER_MILE = 1609.344;
+
+/**
+ * The props for {@link useGeolocationHandler} hook.
+ *
+ * @internal
+ */
+interface GeolocationhandlerProps {
```
tmeyer2115, line 15: Could we leave `Props` out of the name here? When I see `Props`, I expect a React Component, not a hook.

yen-tt: sure, I can update it to `GeolocationhandlerArgs` instead

src/hooks/useGeolocationHandler.ts
yen-tt: `setIsFetchingLocation` is following the common pattern -- `setIsBlah` -- we use for a **boolean setter** function of react `useState` (e.g. setIsFeedbackProvided or setIsOptionsDisabled) so I thought the naming here would make things consistent. Outside of the pattern, maybe `setIsUserLocationFound` or `updateGetLocationStatus` but I don't think it conveys that it can be set to a boolean true, regardless if the location was determined successfully or not, as long as the fetching process is done? Similarly with our callbacks `handleBlah` pattern. Let me know if you have other names you prefer!

src/hooks/useGeolocationHandler.ts
oshi97: Can the H in handler be capitalized?

yen-tt: yup!

src/hooks/useGeolocationHandler.ts
```diff
@@ -0,0 +1,82 @@
+
```
oshi97, line 1: extra line

src/hooks/useGeolocationHandler.ts
```diff
@@ -27,20 +24,21 @@ interface GeolocationhandlerProps {
...
- * @returns A function to collect and process user's geolocation
+ * @param props - {@link GeolocationHandlerArgs}
+ * @returns - A function to collect and process user's geolocation
+ *          - A boolean to indicate if user's geolocation is being fetch
  */
 export function useGeolocationHandler({
-  setIsFetchingLocation,
   geolocationOptions,
   radius = 50,
   handleUserPosition
```
tmeyer2115, line 39: How about `onPositionObtained`?

src/hooks/useGeolocationHandler.ts
tmeyer2115: I don't totally agree with the idea that we should never use the word "on" and always use "handle". I think if done correctly, we could use both and each could convey a different meaning. But, that's a higher-level team convo.

tmeyer2115 (approved):

PR #337, opened 2022-11-16
Fix wcag github action mapbox issues

yen-tt:
paired with the changes in this [PR](https://github.com/yext/slapshot-reusable-workflows/pull/22) in WCAG workflow, this PR updates the WCAG github action in the repo to pass in the mapbox key. Also updated the wcag test to exclude checking elements coming from mapbox canvas container as any potential violations coming from there is outside of our repo's control.

WCAG github action also run on pull request to feature branch now.

J=SLAP-2458
TEST=auto

see that WCAG github action now passes

### DIFF
--- a/.storybook/test-runner.ts
+++ b/.storybook/test-runner.ts
@@ -1,28 +1,36 @@
 import { injectAxe, checkA11y } from 'axe-playwright';
 import { Page } from 'playwright-core';
 import { runOnly } from './wcagConfig';
+import { TestContext, TestRunnerConfig } from '@storybook/test-runner';
 
 /**
  * See https://storybook.js.org/docs/react/writing-tests/test-runner#test-hook-api-experimental
  * to learn more about the test-runner hooks API.
  */
-const renderFunctions = {
+const renderFunctions: TestRunnerConfig = {
   async preRender(page: Page) {
     await injectAxe(page);
   },
-  async postRender(page: Page, context) {
-    await checkA11y(page, '#root', {
-      axeOptions: {
-        runOnly,
-        rules: {
-          'color-contrast': { enabled: context.name !== 'Loading' }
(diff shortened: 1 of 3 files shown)


nmanu1 (commented):
do you know why the coverage check is still failing even though the WCAG check passes? I thought they were related

.storybook/test-runner.ts
```diff
@@ -1,28 +1,36 @@
...
-      },
-      detailedReport: true,
-      detailedReportOptions: {
-        html: true,
+  async postRender(page: Page, context: TestContext) {
+    await checkA11y(
+      page,
+      {
+        include: ['#root'],
+        exclude: ['#root .mapboxgl-canvas-container'],
```
tmeyer2115, line 19: For my own edification, was this line the only change needed to get the WCAG tests passing? If we're ignoring Mapbox, the passing of the `MAPBOX_API_KEY` shouldn't really matter, right?

yen-tt:
> do you know why the coverage check is still failing even though the WCAG check passes? I thought they were related

I did a brief investigation yesterday and it appears to be a separate issue, where the visual coverage script seems to be killing the process after running `test-storybook --coverage` and didn't get to the rest of the commands for generating lcov coverage and moving the files to the expected directory. That will be another item / PR.

.storybook/test-runner.ts
yen-tt: Both the passing of mapbox api key and the change in test-runner is needed to get WCAG test passing. It couldn't properly run checks on mapbox stories because the mapbox component never gets render due to no api key, and that causes errors within `play` function and closed the browser/session. With `MAPBOX_API_KEY`, the mapbox stories was evaluated correctly, but there are a couple violations within mapbox canvas, which is what the change in test-runner is for.

tmeyer2115 (approved):

PR #343, opened 2022-12-05
FilterSearch updates for `builtin.location`

nmanu1:
Update `FilterSearch` to account for special backend logic when searching on the `builtin.location` field. In this case, the backend can return filters on `builtin.region` or `address.countryCode` as well, so these `fieldId`s are also considered "matching" for the component. Filters on these fields can be used to populate the input and should be unselected when a new filter is selected from `FilterSearch`.

J=SLAP-2495
TEST=auto, manual

See that the added Jest tests pass. Spin up the test-site and add static filters on `builtin.region`. See that they are used to populate the `FilterSearch` input, warnings are logged, and the filters are unselected by `FilterSearch` as expected.

### DIFF
--- a/src/components/FilterSearch.tsx
+++ b/src/components/FilterSearch.tsx
@@ -105,6 +105,14 @@ export function FilterSearch({
   const searchParamFields = searchFields.map((searchField) => {
     return { ...searchField, fetchEntities: false };
   });
+  const matchingFieldIds: Set<string> = useMemo(() => {
+    const fieldIds = new Set(searchFields.map(s => s.fieldApiName));
+    if (fieldIds.has('builtin.location')) {
+      ['builtin.region', 'address.countryCode'].forEach(s => fieldIds.add(s));
+    }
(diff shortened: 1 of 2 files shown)


oshi97 (approved):
I was wondering if we'd want to put this weird business logic into headless, since this is the kind of weird business logic that's easy to forget about, but at least right now I'm pretty sure we wouldn't want to do that.

nmanu1:
> I was wondering if we'd want to put this weird business logic into headless, since this is the kind of weird business logic that's easy to forget about, but at least right now I'm pretty sure we wouldn't want to do that.

Yeah, I didn't think we'd want to bake this into headless because it seems to mostly affect the UI for now and it sounds like Product is thinking of making a separate endpoint for this case anyways. I'm not a big fan of having this kind of weird logic in `search-ui-react` either, but I'm hoping we'll be able to remove it soon and having it higher in the stack will probably make it easier to isolate/remove

PR #344, opened 2022-12-05
Update Geolocation to remove all location fields

nmanu1:
Update `useGeolocationHandler` so that it replaces all location filters with the new filter. In other words, filters on the `builtin.location`, `builtin.region`, and `address.countryCode` fields are all removed when a new filter is selected via the `Geolocation` component. This matches the behavior of `FilterSearch` and ensures the `State` doesn't end up with multiple location filters applied at the same time, which can have unpredictable results from the backend.

J=SLAP-2495
TEST=auto

See that the Jest tests still pass even when the mocked state includes other kinds of location filters besides those on `builtin.location`.

### DIFF
--- a/src/hooks/useGeolocationHandler.ts
+++ b/src/hooks/useGeolocationHandler.ts
@@ -3,7 +3,8 @@ import { executeSearch } from '../utils/search-operations';
 import { getUserLocation } from '../utils/location-operations';
 import { useCallback, useState } from 'react';
 
-const LOCATION_FIELD_ID = 'builtin.location';
+const GEOLOCATION_FIELD_ID = 'builtin.location';
+const LOCATION_FIELD_IDS = [GEOLOCATION_FIELD_ID, 'builtin.region', 'address.countryCode'];
 const METERS_PER_MILE = 1609.344;
 
 /**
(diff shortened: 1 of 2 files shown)


src/hooks/useGeolocationHandler.ts
```diff
@@ -3,7 +3,8 @@ import { executeSearch } from '../utils/search-operations';
 import { getUserLocation } from '../utils/location-operations';
 import { useCallback, useState } from 'react';
 
-const LOCATION_FIELD_ID = 'builtin.location';
+const GEOLOCATION_FIELD_ID = 'builtin.location';
+const LOCATION_FIELD_IDS = [GEOLOCATION_FIELD_ID, 'builtin.region', 'address.countryCode'];
```
yen-tt, line 7: can these consts be use in filtersearch?

nmanu1: I didn't use them in the FilterSearch PR because that was off of the release branch, which doesn't have Geolocation. similarly, develop doesn't have those FilterSearch changes, so I can't directly update it in this PR either. given the current naming of the consts (i.e. `GEOLOCATION_FIELD_ID`), I think it might be better to wait until the FilterSearch changes are in develop and there's more context for geolocation. but, if you feel strongly about it, I can extract these consts into a common file and make another PR to refer to them from there in the release branch

yen-tt: that's alright, not a big deal

yen-tt (approved):

PR #346, opened 2022-12-07
add checks to fail combined-coverage tests after each part

tatimblin:
Currently tests are failing silently in Github actions. This PR adds checks to combined-coverage.sh script to exit if a test fails. The three tests currently failing on main will now be caught, I created [this item](https://yexttest.atlassian.net/browse/SLAP-2508) to handle them separately, as it will require a change to make react versions to test programmable in reusable workflows

J=SLAP-2459
TEST=manual

ran actions with an intentional failure ([log](https://github.com/yext/search-ui-react/actions/runs/3649682095/jobs/6164675871)). Also commented out `npm publish` from the `slap publish` command and tested that it makes it passed the check there

### DIFF
--- a/tests/scripts/combined-coverage.sh
+++ b/tests/scripts/combined-coverage.sh
@@ -1,9 +1,11 @@
 #!/bin/bash
 
 npm run test:unit
-echo $?
+if $? ; then
+  exit 1
 npm run test:visual
-echo $?
+if $? ; then
+  exit 1
 
 # merge
 mkdir -p coverage/merge
--- a/tests/scripts/combined-coverage.sh
+++ b/tests/scripts/combined-coverage.sh
@@ -1,13 +1,7 @@
 #!/bin/bash
 
-npm run test:unit
-if $? ; then
-  exit 1
-fi
-npm run test:visual
-if $? ; then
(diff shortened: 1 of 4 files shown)


tests/scripts/combined-coverage.sh
```diff
@@ -10,3 +10,5 @@ cp coverage/visual/coverage-storybook.json coverage/merge/coverage-storybook.jso
 
 nyc report --reporter=lcov --reporter=text -t coverage/merge --report-dir coverage/merge
 cp coverage/merge/lcov.info coverage/lcov.info
+
+exit 0
```
oshi97, line 14: Is this last exit 0 necessary?

tatimblin: The last exit 0 is specific to getting the `slap publish` command to work. Now that there is an exit it will move on to the next check, currently it hangs

yen-tt: why does it hang? without the last exit 0, I would still expect this script to end

tatimblin: I thought it was last week, but ran again and I didn't need it

PR #347, opened 2022-12-15
v1.1.0

tmeyer2115:
### Features
- Default behavior of `FilterSearch` was changed to better support Locators and Doctor Finders. Additionally, a new `onSelect` prop was added to the Component. The `searchOnSelect` prop is now deprecated.  (#323, #343, #333)
- A new CSS bundle without the Tailwind resets is exported. (#322)
- We've added a `MapboxMap` Component, powered by v2 of their JavaScript API. (#332)

### Changes
- Assorted updates to improve our GH Actions. 
- Styling of Facet Headers is now exposed in `FilterGroupCssClasses`. (#321)

### Bug Fixes
- Vulnerabilities were addressed for the repo and its test-site. 
- Fixed the Dropdown Component to invoke `preventDefault` only when it is active. (#307)
- Corrected a small error in the generation of SSR Hydration IDs. (#315)


nmanu1 (commented):
> Default behavior of `FilterSearch` was changed to better support Locators and Doctor Finders. Additionally, a new `onSelect` prop was added to the Component.

For this point, can we also mention that `searchOnSelect` is now deprecated?
Also, can we add #321 to the notes

nmanu1 (commented):
> Styling of Facet Headers is now exposed in `FilterGroup`

nit: can this be updated to refer to `FilterGroupCssClasses` instead of `FilterGroup`, since the latter is not publicly exported?

nmanu1 (approved):

PR #350, opened 2023-01-04
Pin react-collapsed to minor version

nmanu1:
This PR pins `react-collapsed` to the latest minor version (v3.6.0) that supports React 17. In v3.7.0 and v3.8.0, `react-collapsed` fixed a bug ([Slack thread](https://yext.slack.com/archives/C032CKFARGS/p1671130163844249), [GH issue](https://github.com/roginfarrer/react-collapsed/issues/103)) that occurs when using it with React 18 by introducing a breaking change where they dropped compatibility with React 16 and 17. v3.7.0+ would break a search experience using collapsible facets with React <18. Since they are not able to support both React 17 and 18 together, it seems preferable to pin to a version that can be used with both, but is a little buggy with React 18, rather than fully functional with React 18, but breaks other React experiences.

J=SLAP-2521
TEST=manual

See that collapsible facets works as expected on a test-site running on React 17 and that the site running on React 18 still has most functionality, except for the bug described above.

### DIFF
--- a/THIRD-PARTY-NOTICES
+++ b/THIRD-PARTY-NOTICES
@@ -3522,22 +3522,6 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 The following NPM package may be included in this product:
 
- - performance-now@2.1.0
-
-This package contains the following license and notice below:
-
-Copyright (c) 2013 Braveg1rl
-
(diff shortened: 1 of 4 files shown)


oshi97:
It's expected for the react 18 tests to fail right?

nmanu1:
> It's expected for the react 18 tests to fail right?

Yes, the react 18 tests have been failing for a while now (they've only been explicitly visible since we updated the script to actually fail properly). It's unrelated to the change in this PR, but any component/test that uses `renderToString` (SSR tests) or `ReactDOM.render` (`MapboxMap`) will fail in the react 18 tests

tmeyer2115 (approved):

PR #360, opened 2023-04-28
Create Facets component

cea2aj:
Create a facets component which renders any facets returned by the search. If any hierarchical facets are returned, they are rendered as standard facets.

J=BACK-2220
TEST=manual, auto

Manually tested this component on the test site and confirmed it rendered all types of facets. Also added a storybook page and automated tests and confirmed they passed.


tests/__utils__/facets.ts
```diff
@@ -0,0 +1,5 @@
+import { DisplayableFacetOption } from '@yext/search-headless-react';
+
+export function getOptionLabelText(option: DisplayableFacetOption) {
```
EmilyZhang777, line 3: can you add a an example of what this is trying to match?

test-site/src/pages/PeoplePage.tsx
```diff
@@ -55,11 +55,13 @@ export function PeoplePage() {
             title='Static Employee Department'
             filterOptions={employeeFilterConfigs}
           />
+          <FilterDivider />
           <StaticFilters
             fieldId='c_hierarchicalFacet'
             title='Static Hierarchical Facets'
             filterOptions={hierarchicalFilterConfigs}
           />
+          <FilterDivider />
```
EmilyZhang777, line 64: why is this only wrapping StaticFilters and br tag is used on line 73?

EmilyZhang777 (commented):
LGTM, just two minor things that mostly are just questions

benmcginnis (approved):
Lgtm as well, just another minor comment

test-site/src/pages/PeoplePage.tsx
cea2aj: These aren't wrapping StaticFilters and are instead standalone filter dividers. The facets dynamically include dividers where needed, but static filters do not

PR #370, opened 2023-05-10
Add Singular StandardFacet

EmilyZhang777:
This change adds a singular StandardFacet that can be used to override facets rendered in Facets.

J=BACK-2221
TEST=auto,manual

Ran `npm run test` and confirmed all tests pass.
Manually added a StandardFacet with Facets on the people's page under test-site/. Confirmed the facet gets overridden correctly.


src/components/FacetProps.tsx
```diff
@@ -0,0 +1,54 @@
...
+  /** A function to transform facet's options. */
+  transformOptions?: (options: DisplayableFacetOption[]) => DisplayableFacetOption[]
+}
+
+/**
+ * Props for a standard facet.
+ *
+ * @public
+ */
+export interface BaseStandardFacetProps {
```
benmcginnis, line 32: Should this extend BaseFacetProps?

src/components/Facets.tsx
```diff
@@ -29,17 +38,21 @@ export interface FacetsProps {
  * and {@link HierarchicalFacets} components can be used instead for more control over facet
  * configuration.
  *
+ * To override a single facet, use {@link StandardFacet} and set the field id on overriddenFieldIds.
```
benmcginnis, line 41: So I don't think that's actually how we want this to work. I think we just want to base it on whether or not a child is provided with a field id prop that matches one of the field ids from the response.

src/components/StandardFacet.tsx
```diff
@@ -0,0 +1,116 @@
...
+ * @public
+ */
+export function StandardFacet({ props, renderDivider = false }) {
+  const {
+    fieldId,
+    searchOnChange,
+    customCssClasses,
+  } = props;
+  return (
+    <FacetsProvider searchOnChange={searchOnChange} className={customCssClasses?.standardFacetContainer || ''}>
```
benmcginnis, line 45: Sorry this wasn't clear from the item description, but the intent was for this provider to live in the Facets component, not each of the singular Facet components

benmcginnis (changes requested):
I think we want to talk through the plan, I don't think I did a good job of explaining what I was thinking in the items.

src/components/Facets.tsx
EmilyZhang777: Done

src/components/Facets.tsx
```diff
@@ -29,17 +26,95 @@ export interface FacetsProps {
...
+                facetType = FacetType.STANDARD;
+              } else if (isNumericalFacet(facet)) {
+                facetType = FacetType.NUMERICAL;
+              }
+            }
+
+            let facetComponent: ReactElement;
+            switch (facetType) {
+              case FacetType.NUMERICAL:
+                facetComponent = <div/>;
```
benmcginnis, line 71: is it safe to use an empty fragment here? `<></>` it feels weird to insert something into the dom

benmcginnis: also do we need to handle the hierarchical case?

src/components/StandardFacetContent.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * A component that displays the content of a standard facet.
+ *
+ * @param props - props to render the component
+ * @returns A React component for the content of a standard facet
+ *
+ * @internal
+ */
+export function StandardFacetContent({
```
benmcginnis, line 13: is this exported as StandardFacet anywhere?  I see it in the Stories byt don't see it elsewhere (i'm probably missing something)

benmcginnis (approved):
LGTM, a couple of questions

src/components/Facets.tsx
EmilyZhang777: It's just a placeholder till we implement the NumericalFacet (which I'm doing right now)

EmilyZhang777: I imagine we'll eventually add the hierarchical case. But I want to keep the behavior of Facets to be the same as of now

src/components/StandardFacetContent.tsx
EmilyZhang777: I'm exporting a StandardFacet inside Facets, which basically just passes the props through and is different from this one. The StandardFacetContent needs to take in a facet so I made it internal and separate from the StandardFacet. Let me know if that's fine with you!

src/components/Facets.tsx
```diff
@@ -29,17 +26,95 @@ export interface FacetsProps {
...
+}
+
+/**
+ * A component that displays a single standard facet. Use this to override the default rendering.
+ *
+ * @param props - {@link StandardFacetProps}
+ * @returns ReactElement
+ * @public
+ */
+export function StandardFacet(props: StandardFacetProps) { return null; }
```
benmcginnis, line 104: Oh I see this is just returning null though?

EmilyZhang777: I just use this to pass along the props from users. The actual rendering requires the facet so I created the StandardFacetContent to handle that

benmcginnis: Yeah I think we should just expose the facet as a prop, right? That would allow the component to be tested easily with arbitrary data (for example)? As well as eventually providing users with an easily modifiable base.

EmilyZhang777: The facet is currently handled by the FacetsProvider. I don't think exposing it as a prop on StandardFacet is what we want. Users won't be able to access the facets easily, unless they use the StandardFacet inside a FacetsProvider, which is what we are trying to blackbox for them.

benmcginnis: Ok

src/components/FacetProps.tsx
```diff
@@ -0,0 +1,62 @@
+import { DisplayableFacetOption } from '@yext/search-core';
```
cea2aj, line 1: nit: is there a reason this is imported directly from core rather than from search-headless-react?

src/components/Facets.tsx
cea2aj, line 104: The general convention in this library is for each public component to be in its own file, along with the props and css classes in that file as well. Is there a reason behind this organization instead?

cea2aj (commented):
looks great! just two small comments

src/components/Facets.tsx
EmilyZhang777: I put it here since in the [spec](https://docs.google.com/document/d/1V1n_mqzZQATM4uQVcmj0_Ia8e-bYXDG58t7G9t0Lop8/edit#), they use `<Facets.StandardFacet/>`. Let me know what you think!

src/components/FacetProps.tsx
EmilyZhang777: nice catch! It was auto generated. Updated

PR #375, opened 2023-05-26
Add Facets Level Css Support for FilterGroup

EmilyZhang777:
This change allows users to provide css classes for FilterGroup on the Facets level. If same css classes are provided on the singular facet level, the latter will be used.

J=BACK-2306
TEST=auto,manual

Added a test for this.
Also manually confirmed the behavior.

### DIFF
--- a/src/components/Facets.tsx
+++ b/src/components/Facets.tsx
@@ -79,39 +79,15 @@ export function Facets(props: FacetsProps) {
               && (!onlyRenderChildren || fieldIdToCustomFacetProps.has(fieldId)))
             .map((fieldId, i) => {
               const facet: DisplayableFacet = fieldIdToFacet.get(fieldId);
-              let facetType: FacetType = FacetType.STANDARD;
-              let facetProps: FacetProps = {
-                fieldId: facet.fieldId,
-                label: facet.displayName,
-              };
-              if (fieldIdToCustomFacetProps.has(facet.fieldId)) {
-                const customFacetElement: ReactElement =
-                  fieldIdToCustomFacetProps.get(facet.fieldId);
-                facetProps = { ...facetProps, ...customFacetElement.props };
-                facetType = getFacetTypeFromReactElementType(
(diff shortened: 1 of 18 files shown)


src/components/Facets.tsx
```diff
@@ -95,6 +95,14 @@ export function Facets(props: FacetsProps) {
                 facetType = getFacetTypeFromFacet(facet, hierarchicalFieldIds);
               }
 
+              facetProps = {
+                customCssClasses: {
+                  ...customCssClasses,
+                  ...facetProps.customCssClasses,
+                },
+                ...facetProps
+              };
```
cea2aj, line 104: is this line `...facetProps.customCssClasses,` necessary? It seems like if the `customFacetElement` has customCssClasses, then those will already take precedence over the customCssClasses defined here

src/components/Facets.tsx
```diff
@@ -95,6 +95,14 @@ export function Facets(props: FacetsProps) {
                 facetType = getFacetTypeFromFacet(facet, hierarchicalFieldIds);
               }
 
+              facetProps = {
```
cea2aj, line 98: I would like to shout out the `useComposedCssClasses` hook in case you haven't seen it yet, which is what we typically use for merging css classes. I don't think it's necessary here because we aren't merging the styling, instead we are using the customCssClasses of a customFacetElement if they are defined and ignoring the classes set at the Facets level. If we wanted to merge the css classes from the Facets and the custom facets, then the `useComposedCssClasses` hook should be used

EmilyZhang777: This merges the two objects but I want to have the classes on a singular facet override the classes set on the facets to make sure the styles on the singular facet are used.

src/components/Facets.tsx
EmilyZhang777: Ack. see other comment

benmcginnis (approved):

PR #377, opened 2023-06-12
truncate search autcomplete results and prevent resizing of icons

saahithjanapati:
truncate search autcomplete results and prevent resizing of icons

J=BACK-2239

T=manual

ran test site and verified with product team

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@yext/search-ui-react",
-  "version": "1.2.0",
+  "version": "1.2.1",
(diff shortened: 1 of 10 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@yext/search-ui-react",
-  "version": "1.1.0",
```
EmilyZhang777, line 3: I'm not quite sure what happened but isn't this 1.3.0-beta.1 on the latest develop branch? https://github.com/yext/search-ui-react/blob/develop/package.json#L3

PR #379, opened 2023-06-20
add support to display richtextv2 and markdown that has been converted to HTML

saahithjanapati:
J=BACK-2241

TEST=manual

ran test-site of search-ui-react with entity that had formatted RichTextv2 data and verified that it was rendered properly

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@yext/search-ui-react",
-  "version": "1.4.0-beta.0",
+  "version": "1.3.0-beta.2",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@yext/search-ui-react",
-      "version": "1.4.0-beta.0",
+      "version": "1.3.0-beta.2",
       "license": "BSD-3-Clause",
       "dependencies": {
         "@microsoft/api-documenter": "^7.15.3",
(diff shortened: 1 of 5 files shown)


package-lock.json
```diff
@@ -1,12 +1,12 @@
 {
   "name": "@yext/search-ui-react",
-  "version": "1.2.0",
+  "version": "1.4.0-beta.0",
```
cea2aj, line 3: this can be 1.3.0-beta.0 since v1.3 hasn't been released yet

src/components/FeaturedSnippetDirectAnswer.tsx
```diff
@@ -29,17 +30,32 @@ interface FeaturedSnippetDirectAnswerCssClasses {
...
-  if (result.fieldType === 'rich_text') {
-    console.warn('Rendering markdown for rich text direct answer is currently not supported. Displaying the unrendered markdown string as is.');
+  if (unsupportedTextFormats.includes(result.fieldType)) {
+    console.warn('Rendering ' + result.fieldType + ' direct answer is currently not supported. ' +
+        'You can modify your search configuration to convert ' + result.fieldType + ' to HTML to be rendered '
+        + 'on the page.');
+  }
+  let snippet: JSX.Element;
+  const snippetValue = useMemo(() =>
+  { return { __html: result.snippet.value }; }, [result.snippet.value]);
```
cea2aj, line 49: can we update these to be `result.snippet?.value`? That way if the snippet isn't a property of the result, the code won't crash by trying to access `snippet.value`

cea2aj (approved):
LGTM!

PR #385, opened 2023-07-17
Added example cards for HTML and Markdown Rendering

saahithjanapati:
Added example cards to render converter HTML and Markdown strings in search results

J=BACK-2395

T=manual
Manually ran test-site with each of the example cards, searching for entities that had the associated data type

### DIFF
--- a/test-site/src/components/HTMLExampleCard.tsx
+++ b/test-site/src/components/HTMLExampleCard.tsx
@@ -16,13 +16,15 @@ const htmlFieldName = 'c_richTextV2Data';
 interface CustomRawDataType {
   name: string,
   description: string,
-  [htmlFieldName]: any
+  [htmlFieldName]: { html: string }
 }
 
 function renderHTMLContent(HTMLContent: string | undefined) {
   if ( HTMLContent )
   {
-    return <div className="reset-style" dangerouslySetInnerHTML={{ __html: HTMLContent }} />;
+    const createDangerousHTMLObject = (htmlContent: string) => ({ __html: htmlContent });
+    const dangerouslySetHTML = createDangerousHTMLObject(HTMLContent);
+    return <div className="reset-style" dangerouslySetInnerHTML={dangerouslySetHTML} />;
   }
   return null;
 }
--- a/test-site/src/components/HTMLExampleCard.tsx
+++ b/test-site/src/components/HTMLExampleCard.tsx
@@ -1,0 +1,43 @@
+import { CardProps, useCardAnalyticsCallback, useCardFeedbackCallback } from '@yext/search-ui-react';
+import { useCallback } from 'react';
+import '../styles/resetStyles.css'
+
+const builtInCssClasses = {
+    container: 'flex flex-col justify-between border border-gray-200 rounded-lg mb-4 p-4 shadow-sm',
+    header: 'flex text-neutral-dark',
+    title: 'text-lg font-medium',
+    thumbsFeedbackContainer: 'flex justify-end mt-4 text-sm text-gray-500 font-medium',
+};
+
+// change to the field name that contains html string
+const htmlFieldName = "c_richTextV2Data";
+interface CustomRawDataType {
+    name: string,
+    description: string
+    [htmlFieldName]: any
+}
+
+function renderRichTextDescription(richTextHTML: string | undefined) {
+    if ( richTextHTML )
+    {
+        return <div className={"reset-style"} dangerouslySetInnerHTML={{ __html: richTextHTML }} />;
+    }
+    return null;
+}
+
+export function HTMLExampleCard(props: CardProps<CustomRawDataType>): JSX.Element {
+    const { result } = props;
+    const onClickTitle = useCardAnalyticsCallback(result, 'TITLE_CLICK');
+    const cardFeedbackCallback = useCardFeedbackCallback(result);
+    const onClick = useCallback(() => {
+        cardFeedbackCallback('THUMBS_UP');
+    }, [cardFeedbackCallback]);
+    return (
+        <div className={builtInCssClasses.container}>
+            <p className={builtInCssClasses.header}>HTML Card</p>
+            <button onClick={onClickTitle} className={builtInCssClasses.title}>{result.rawData.name}</button>
+            {renderRichTextDescription(result.rawData?.[htmlFieldName]?.html)}
+            <button onClick={onClick} className={builtInCssClasses.thumbsFeedbackContainer}>Feedback</button>
+        </div>
+    );
+}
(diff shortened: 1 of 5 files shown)


test-site/src/components/HTMLExampleCard.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+// change to the field name that contains html string
+const htmlFieldName = "c_richTextV2Data";
+interface CustomRawDataType {
+    name: string,
+    description: string
+    [htmlFieldName]: any
+}
+
+function renderRichTextDescription(richTextHTML: string | undefined) {
```
EmilyZhang777, line 20: I'm not sure what "Description" means in here. Would it make more sense to call it `renderRichTextContent`?

test-site/src/components/HTMLExampleCard.tsx
```diff
@@ -0,0 +1,43 @@
...
+    header: 'flex text-neutral-dark',
+    title: 'text-lg font-medium',
+    thumbsFeedbackContainer: 'flex justify-end mt-4 text-sm text-gray-500 font-medium',
+};
+
+// change to the field name that contains html string
+const htmlFieldName = "c_richTextV2Data";
+interface CustomRawDataType {
+    name: string,
+    description: string
```
EmilyZhang777, line 16: this doesn't seem to be used?

test-site/src/components/MarkdownExampleCard.tsx
```diff
@@ -0,0 +1,42 @@
...
+    title: 'text-lg font-medium',
+    thumbsFeedbackContainer: 'flex justify-end mt-4 text-sm text-gray-500 font-medium',
+};
+
+
+const markdownFieldName  = "c_markdownData";
+
+interface CustomRawDataType {
+    name: string,
+    description: string,
```
EmilyZhang777, line 20: same here

test-site/src/components/HTMLExampleCard.tsx
```diff
@@ -0,0 +1,43 @@
...
+    container: 'flex flex-col justify-between border border-gray-200 rounded-lg mb-4 p-4 shadow-sm',
+    header: 'flex text-neutral-dark',
+    title: 'text-lg font-medium',
+    thumbsFeedbackContainer: 'flex justify-end mt-4 text-sm text-gray-500 font-medium',
+};
+
+// change to the field name that contains html string
+const htmlFieldName = "c_richTextV2Data";
+interface CustomRawDataType {
+    name: string,
```
EmilyZhang777, line 15: maybe call it `title`?

EmilyZhang777 (commented):
Is this what Baigel wants? I feel like there's still something we can encapsulate and handle on our end. I'm imaging we can have a <HTMLCard/> component and all a user needs to do is passing in a CustomRawDataType (or call it `HTMLFieldData` maybe?). But definitely talk to Baigel and see what he thinks.

test-site/src/components/MarkdownExampleCard.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+const builtInCssClasses = {
+    container: 'flex flex-col justify-between border border-gray-200 rounded-lg mb-4 p-4 shadow-sm',
+    header: 'flex text-neutral-dark',
+    title: 'text-lg font-medium',
+    thumbsFeedbackContainer: 'flex justify-end mt-4 text-sm text-gray-500 font-medium',
+};
+
+
+const markdownFieldName  = "c_markdownData";
```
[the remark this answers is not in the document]
saahithjanapati: added comment

test-site/src/components/HTMLExampleCard.tsx
saahithjanapati: Changed this to "renderHTMLContent" because it can be used for either Lexical or Markdown that has been converted to HTML

test-site/src/components/HTMLExampleCard.tsx
saahithjanapati: This is used to define the type of prop used in the exported function. Without it, typescript seems to give errors when we try to access the user-specified field name

test-site/src/components/MarkdownExampleCard.tsx
saahithjanapati: (same comment as above): This is used to define the type of prop used in the exported function. Without it, typescript seems to give errors when we try to access the user-specified field name

test-site/src/components/HTMLExampleCard.tsx
saahithjanapati: Do you mean changing "htmlFieldName" to title? I think htmlFieldName might be more flexible because HTML can be returned for any field that has converted RichTextV2/markdown.

saahithjanapati:
Regarding your comment about integrating the cards together, this is something we can do. However, there were a few reasons we decided against doing this last week, after consulting with Baigel and Tom (from Slapshot):

- Awkward to Use: The HTML Card can already be used to render both RichTextV2 and Markdown that has been converted to HTML. We can also integrate the Markdown renderer into this same card, but it will be a bit awkward since we won't be able to include a Lexical Editor into the card (because Lexical is currently in beta and we don't have a fixed version to use yet).

- Not Backfire's Responsibility: Tom was saying this card probably doesn't belong in the search-ui-react repo because the problem of rendering these different text formats is not a search-specific issue. He was saying that this card type of card would better fit in the react-components repository and should probably be created after the RichTextV2 conversion was stable (i.e. we would have a fixed version of lexical editor to point to).

benmcginnis:
Lgtm

cea2aj:
It looks like there are some linting errors on the cards, can you address those?

test-site/src/components/HTMLExampleCard.tsx
```diff
@@ -16,13 +16,15 @@ const htmlFieldName = 'c_richTextV2Data';
...
+  [htmlFieldName]: { html: string }
 }
 
 function renderHTMLContent(HTMLContent: string | undefined) {
   if ( HTMLContent )
   {
-    return <div className="reset-style" dangerouslySetInnerHTML={{ __html: HTMLContent }} />;
+    const createDangerousHTMLObject = (htmlContent: string) => ({ __html: htmlContent });
+    const dangerouslySetHTML = createDangerousHTMLObject(HTMLContent);
+    return <div className="reset-style" dangerouslySetInnerHTML={dangerouslySetHTML} />;
```
cea2aj, line 27: Fixing the linting error seems to have complicated this a bit. Can we instead create the dangerous html object in the HTML example card using the useMemo hook and pass that object into the renderHTML content function? While we're here, can we also avoid using the parameter name HTMLContent since that looks like a type rather than a variable? I think we should stick to `htmlContent` or `html`

cea2aj (approved):
LGTM!

PR #387, opened 2023-08-08
Make lib ESM compatible

mkilpatrick:


mkilpatrick:
This is currently breaking the storybook tests because it needs to be updated to the latest version 7 for ESM support.

mkilpatrick:
Ran into issues upgrading Storybook. Going to put up another PR that is much simpler.

PR #400, opened 2023-09-25
PagesJS compatibility

cea2aj:
Introduce tsup and improve the package.json file in order to support PagesJS.

This works in both dev and prod mode of pagesJS. Currently the storybook tests are failing, however we will address that in a future PR.

This changes updates our esm files to created in a bundle in the dist folder. The esm files now end with .mjs rather than .js which may be a breaking change for some build tools. I struggled to get `type: module` to work in many situations, whereas using .mjs files worked better across the board. Regardless, these changes caused a lot of different issues, so I think this should go in the next major version

J=BACK-2545
TEST=manual

Publish an alpha version and test it with @yext/pages@1.0.0-rc.3. Add several search components to the page and confirm they work as expected.

Currently the storybook tests are failing and need to be fixed.


package.json
```diff
@@ -142,7 +143,7 @@
     "@reach/auto-id": "^0.18.0",
     "@restart/ui": "^1.0.1",
     "@tailwindcss/forms": "^0.5.0",
-    "@yext/analytics": "^0.6.4",
+    "@yext/analytics": "^0.5.0",
```
benmcginnis, line 146: should we be downgrading?

src/components/Dropdown/Dropdown.tsx
```diff
@@ -1,12 +1,16 @@
 import { createElement, isValidElement, KeyboardEvent, PropsWithChildren, ReactNode, useEffect, useMemo, useRef, useState } from 'react';
 import { DropdownContext, DropdownContextType } from './DropdownContext';
 import { InputContext, InputContextType } from './InputContext';
-import useRootClose from '@restart/ui/useRootClose';
+import useRootClosePkg from '@restart/ui/useRootClose';
+const useRootClose = typeof useRootClosePkg === 'function' ? useRootClosePkg : useRootClosePkg['default'];
```
benmcginnis, line 5: yikes we have to do this in our source code to make vite happy? cc @mkilpatrick

src/components/ThumbsFeedback.tsx
```diff
@@ -1,8 +1,11 @@
 import { useSearchState } from '@yext/search-headless-react';
 import { useCallback, useState } from 'react';
 import { ThumbIcon } from '../icons/ThumbIcon';
-import useLayoutEffect from 'use-isomorphic-layout-effect';
 import { useComposedCssClasses } from '../hooks/useComposedCssClasses';
+import useIsomorphicLayoutEffect from "use-isomorphic-layout-effect";
+const useLayoutEffect = typeof useIsomorphicLayoutEffect === 'function'
+  ? useIsomorphicLayoutEffect
```
benmcginnis, line 7: why would it be different?

package.json
cea2aj: upon further investigation, downgrading doesn't seem necessary. while were here, i'll bump to the latest v0.6.5

src/components/Dropdown/Dropdown.tsx
cea2aj: This was the only way I was able to make typescript, vite, and deno happy at the same time. Potentially there is a config option or plugin that could handle this during the build step, but I wasn't able to find one. tsup has a cjsInterop flag but that doesn't fix this issue

src/components/ThumbsFeedback.tsx
cea2aj: when vite is running this esm code (pagesjs dev and node), it imports the package as cjs and the export becomes an object with the "default" property, which is different than how the import works in other environments

package.json
```diff
@@ -1,12 +1,11 @@
...
-  "version": "1.3.3",
+  "version": "1.4.0-alpha.19",
   "description": "A library of React Components for powering Yext Search integrations",
   "author": "slapshot@yext.com",
   "license": "BSD-3-Clause",
-  "main": "./lib/esm/index.js",
-  "types": "./lib/esm/index.d.ts",
-  "type": "module",
+  "main": "./dist/index.js",
+  "types": "./dist/index.d.ts",
```
mkilpatrick, line 8: You removed `"type":"module"`

package.json
```diff
@@ -19,20 +18,21 @@
...
     "src",
-    "THIRD-PARTY-NOTICES"
+    "THIRD-PARTY-NOTICES",
+    "dist"
   ],
   "scripts": {
-    "build": "rm -rf lib/** && npm run build:js && npm run build:css && npm run api-extractor && npm run generate-docs && npm run generate-notices",
-    "build:js": "tsc -p tsconfig.esm.json && tsc -p tsconfig.cjs.json",
+    "build": "rm -rf lib/** && rm -rf dist/** && npm run build:js && npm run build:css && npm run api-extractor && npm run generate-docs && npm run generate-notices",
+    "build:js": "tsc -p tsconfig.esm.json && tsc -p tsconfig.cjs.json && tsup src/index.ts --dts --sourcemap --format esm,cjs",
```
mkilpatrick, line 35: I don't think you need both tsc and tsup

src/components/ThumbsFeedback.tsx
```diff
@@ -1,8 +1,11 @@
 import { useSearchState } from '@yext/search-headless-react';
 import { useCallback, useState } from 'react';
 import { ThumbIcon } from '../icons/ThumbIcon';
-import useLayoutEffect from 'use-isomorphic-layout-effect';
 import { useComposedCssClasses } from '../hooks/useComposedCssClasses';
+import useIsomorphicLayoutEffect from "use-isomorphic-layout-effect";
```
mkilpatrick, line 5: I recommend opening an issue with this library. It'd be very easy for them to fix their ESM outputs and would avoid you having to do this bologna.

https://publint.dev/use-isomorphic-layout-effect@1.1.2

src/hooks/useRemovableFilters.ts
```diff
@@ -1,5 +1,5 @@
 import { FieldValueFilter, useSearchState, useSearchActions, DisplayableFacet, DisplayableFacetOption, SearchActions, Matcher } from '@yext/search-headless-react';
-import { isEqual } from 'lodash';
+import lodash from 'lodash';
```
mkilpatrick, line 2: I'd use lodash-es instead.

src/components/ThumbsFeedback.tsx
benmcginnis: We could also just fork the library and fix it in our fork, and submit a PR

benmcginnis: Would that be the same for the rest of the bologna?

mkilpatrick: It would mean you should be able to use named imports properly instead of a package-level default import and getting the properties off of it. And also not have to do these weird checks. At least I'd hope not. It might depend on their actual implementation of what they're exporting.

package.json
mkilpatrick: What do you mean here? People import your package like `@yext/search-ui-react/lib/...`?

You're outputting to /dist now instead of /lib.

cea2aj: I suppose we don't advertise it for this repo specifically, but for but search-core we do advertise the different import paths: https://github.com/yext/search-core#explanation-of-builds

It's sometimes useful to be able to import a specific module type rather than relying on build tools to pick it automatically

cea2aj: You don't have to import the full file path however

src/hooks/useRemovableFilters.ts
cea2aj: I attempted to use lodash-es, but I was running into some strange errors with jest, however regular lodash worked fine so I stuck with that

benmcginnis (approved):

test-site/package-lock.json
```diff
@@ -20343,22 +21121,37 @@
       "requires": {}
     },
     "@csstools/selector-specificity": {
-      "version": "2.0.2",
-      "resolved": "https://registry.npmjs.org/@csstools/selector-specificity/-/selector-specificity-2.0.2.tgz",
-      "integrity": "sha512-IkpVW/ehM1hWKln4fCA3NzJU8KwD+kIOvPZA4cqxoJHtE21CCzjyp+Kxbu0i5I4tBNOlXPL9mjwnWlL0VEG4Fg==",
+      "version": "2.2.0",
```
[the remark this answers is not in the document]
cea2aj: I updated the package-lock which should resolve this

benmcginnis (approved):

PR #407, opened 2023-11-29
Take snapshots without Percy

nmanu1:
Update the visual regression snapshots to use the Storybook add-on `jest-image-snapshot` rather than Percy. We can now remove Percy from the repo entirely. The workflow now checks snapshots and commits any that have changed. Also, the play functions for stories are fixed so async actions are properly awaited. 

The loading state for the `Geolocation` component includes animation, so the snapshot doesn't match perfectly each time. To account for this, the threshold for failure for this story is increased to a 0.005% difference in the image rather than an exact match. We also use this threshold for `MapboxMap` stories because the map canvas can have slight changes.

The `MapboxMap` component takes some time to load. I wasn't able to find a good way to detect the correct time to take a snapshot, so I added a 7500ms wait for these stories to give the map time to load before taking the snapshot. Adding more robust detection logic will be part of another [item](https://yexttest.atlassian.net/browse/BACK-2843). This item will also address the snapshot for the loading story for `LocationBias`. Right now, we skip taking a snapshot for this story because the snapshot is blank. This is likely because geolocation permissions are disabled for the test runner, so the request for the location is automatically denied.

J=BACK-2832
TEST=auto


.storybook/snapshots/test-runner.ts
```diff
@@ -0,0 +1,37 @@
...
+    if (isMapboxMapStory) {
+      await page.waitForTimeout(7500);
+    }
+
+    const image = await page.screenshot();
+    expect(image).toMatchImageSnapshot({
+      customSnapshotsDir,
+      customSnapshotIdentifier: context.id,
+      failureThreshold: context.id === 'geolocation--loading' || isMapboxMapStory
+        ? 0.005
```
EmilyZhang777, line 30: out of curiosity, does this mean the difference threshold between two snapshots?

EmilyZhang777 (approved):
LGTM! Glad we could finally fix this

.storybook/snapshots/test-runner.ts
nmanu1: exactly, this means that if the image that's currently taken is more than 0.005% pixels different from the image saved in the repo, then it'll fail and trigger a snapshot update

PR #421, opened 2024-01-16
feat(custom): add onDropdownInputChange optional prop to FilterSearch

Jeffrey-Rhoads17:
expose a callback prop to be called everytime after input changes in the FilterSearch component. This enables greater customization for our Locators that are built using FilterSearch primarily

J=BACK-2765
TEST=manual


src/components/FilterSearch.tsx
```diff
@@ -78,6 +78,8 @@ export interface FilterSearchProps {
   searchOnSelect?: boolean,
   /** A function which is called when a filter is selected. */
   onSelect?: (params: OnSelectParams) => void,
+  /** A function which is called when the input's value changes. This does not replace executeFilterSearch, but gets called just before it */
+  onDropdownInputChange?: (value: string) => void,
```
benmcginnis, line 82: I think it would be helpful to elaborate on why one would use this callback over `onSelect` which superficially sounds very similar.

test-site/src/pages/PeoplePage.tsx
```diff
@@ -52,6 +52,18 @@ export function PeoplePage() {
             searchFields={filterSearchFields}
             searchOnSelect={true}
             label='Filters'
+            onDropdownInputChange={(value: string) => {
+              if (value !== "") return;
+              const fieldIds = filterSearchFields.map(field => field.fieldApiName);
+              const filtersToKeep: SelectableStaticFilter[] = [];
+              searchActions.state.filters.static?.forEach((staticFilter) => {
```
benmcginnis, line 59: so this use case is demonstrating removing static filters from the static filter state when they're used in filter search?

It might be helpful as documentation to explain a bit the intention behind this feature.

benmcginnis (commented):
@nmanu1 if you could take a look as well, I'd appreciate it!

src/components/FilterSearch.tsx
nmanu1, line 82: it might be confusing to users that `onSelect` replaces the select behavior but `onDropdownInputChange` is added onto the existing input change behavior. what do you think about changing `onDropdownInputChange` to replace the behavior like `onSelect` does? that would make it more consistent, and gave even more flexibility to users in case they don't want to trigger the `executeFilterSearch` at all

src/components/FilterSearch.tsx
```diff
@@ -267,6 +270,11 @@ export function FilterSearch({
     }
   }, [executeFilterSearch]);
 
+  const handleInputChange = onDropdownInputChange ? (value: string) => {
```
nmanu1, line 273: can we use `useCallback` here to memoize this handler?

nmanu1 (commented):
you'll want to merge this PR into `develop` rather than `main`

src/components/FilterSearch.tsx
```diff
@@ -55,6 +55,22 @@ export interface OnSelectParams {
...
 
+/**
+ * The parameters that are passed into {@link FilterSearchProps.onDropdownInputChange}.
+ *
+ * @public
+ */
+export interface OnDropdownInputChangeProps {
+  /** The new input element's new value after the change */
+  value: string,
+  searchFields: Omit<SearchParameterField, 'fetchEntities'>[],
```
nmanu1, line 66: is there a reason to have the `searchFields` passed in here? the user passes them to the `FilterSearch` component to begin with, so they should already have that object created

src/components/FilterSearch.tsx
```diff
@@ -78,6 +94,11 @@ export interface FilterSearchProps {
   searchOnSelect?: boolean,
   /** A function which is called when a filter is selected. */
   onSelect?: (params: OnSelectParams) => void,
+  /** A function which is called everytime the input element's value changes.
+   * Helpful for removing static filters from the static filter state if a user clears the search input completely.
+   * Replaces the default behavior of calling executeFilterSearch whenever the input changes.
```
nmanu1, line 99: I think these docs are a little too specific and reference implementation details that the user doesn't necessarily need to know about (e.g. that we call `executeFilterSearch`). Maybe it could be something like this instead: `A function which is called when the input element's value changes. Replaces the default behavior.` The specific example you gave could be added to the HH documentation or a starter repo

src/components/FilterSearch.tsx
```diff
@@ -55,6 +55,22 @@ export interface OnSelectParams {
   executeFilterSearch: (query?: string) => Promise<FilterSearchResponse | undefined>
 }
 
+/**
+ * The parameters that are passed into {@link FilterSearchProps.onDropdownInputChange}.
+ *
+ * @public
+ */
+export interface OnDropdownInputChangeProps {
+  /** The new input element's new value after the change */
```
nmanu1, line 64: nit: `The input element's new value after the change.`

benmcginnis (approved):

.storybook/snapshots/__snapshots__/filtersearch--dropdown-sectioned.png
nmanu1, line 1: looks like there was a visual regression here

src/components/FilterSearch.tsx
```diff
@@ -224,6 +246,21 @@ export function FilterSearch({
     matchingFieldIds
   ]);
 
+  const handleInputChange = useCallback((value) => {
+    if (onDropdownInputChange) {
+      onDropdownInputChange({
```
nmanu1, line 251: nit: I think stylistically we prefer a ternary operator here

tests/components/FilterSearch.test.tsx
```diff
@@ -668,6 +668,13 @@ describe('search without section labels', () => {
     userEvent.keyboard('{enter}');
     expect(inputNode).toHaveValue('first name 1');
   });
+  
+  it('when an onDropdownInputChange prop is specified, it gets called each time after the input changes', async () => {
+    const mockedOnDropdownInputChange = jest.fn();
+    renderFilterSearch({ searchFields: searchFieldsProp, onDropdownInputChange: mockedOnDropdownInputChange});
+    userEvent.type(screen.getByRole('textbox'), 'a');
+    await waitFor(() => expect(mockedOnDropdownInputChange).toHaveBeenCalledTimes(1));
```
nmanu1, line 676: can we also check that `executeFilterSearch` was not called here?

src/components/FilterSearch.tsx
Jeffrey-Rhoads17: The reason I included `searchFields` as a prop for the `onDropdownInputChange` function is so the user can write one function to pass into `onDropdownInputChange` for multiple `FilterSearch` components that may be on one single page. If I don't pass `searchFields`, the user would need to write a unique function for each `FilterSearch` component that uses the proper `searchFields` that belong to that `FilterSearch` instance.

tests/components/FilterSearch.test.tsx
```diff
@@ -668,6 +668,17 @@ describe('search without section labels', () => {
     userEvent.keyboard('{enter}');
     expect(inputNode).toHaveValue('first name 1');
   });
+  
+  it('when an onDropdownInputChange prop is specified, it gets called each time after the input changes and executeFilterSearch does not', async () => {
+    const mockedOnDropdownInputChange = jest.fn();
+    const executeFilterSearch = jest
+      .spyOn(SearchHeadless.prototype, 'executeFilterSearch')
+      .mockResolvedValue(unlabeledFilterSearchResponse);
```
nmanu1, line 676: nit: we shouldn't need the `mockResolvedValue` since this function isn't being called

tests/components/FilterSearch.test.tsx
```diff
@@ -668,6 +668,17 @@ describe('search without section labels', () => {
...
+    renderFilterSearch({ searchFields: searchFieldsProp, onDropdownInputChange: mockedOnDropdownInputChange});
+    userEvent.type(screen.getByRole('textbox'), 'a');
+    await waitFor(() => expect(mockedOnDropdownInputChange).toHaveBeenCalledTimes(1));
+    await waitFor(() => expect(executeFilterSearch).toHaveBeenCalledTimes(0));
```
nmanu1, line 680: nit: this `waitFor` can probably be removed since we're already waiting for the other function to have executed before this check

tests/components/FilterSearch.test.tsx
```diff
@@ -668,6 +668,16 @@ describe('search without section labels', () => {
...
   });
+  
+  it('when an onDropdownInputChange prop is specified, it gets called each time after the input changes and executeFilterSearch does not', async () => {
+    const mockedOnDropdownInputChange = jest.fn();
+    const executeFilterSearch = jest
+      .spyOn(SearchHeadless.prototype, 'executeFilterSearch');
+    renderFilterSearch({ searchFields: searchFieldsProp, onDropdownInputChange: mockedOnDropdownInputChange});
+    userEvent.type(screen.getByRole('textbox'), 'a');
+    await waitFor(() => expect(mockedOnDropdownInputChange).toHaveBeenCalledTimes(1));
+    () => expect(executeFilterSearch).toHaveBeenCalledTimes(0);
```
nmanu1, line 679: since this is in the form of a callback, I don't think it'll actually execute the check

tests/components/FilterSearch.stories.tsx
```diff
@@ -48,7 +48,7 @@ export default meta;
 export const Primary: StoryFn<FilterSearchProps> = (args) => {
   return (
     <SearchHeadlessContext.Provider value={generateMockedHeadless(mockedHeadlessState)}>
-      <FilterSearch {...args} searchFields={searchFields} />
+      <FilterSearch {...args} searchFields={searchFields} onDropdownInputChange={undefined} />
```
nmanu1, line 51: nit: can you add this to the `args` object above instead of specifying it inline? it would be similar to how we have "Filter" as the default value for the `label` prop on line 43. that way if we add any stories in the future, we won't have to remember to add this

nmanu1 (approved):

PR #422, opened 2024-01-31
Resolve Vulnerabilities

EmilyZhang777:
This PR addresses the following vulnerabilities under test-site:
- bump webpack version to 5.90.0
- remove real live api key

J=VULN-37163,VULN-38428
TEST=auto

Ran `npm run test`. Spun up the test site and confirmed the pages are working.


benmcginnis (approved):

test-site/sample.env
```diff
@@ -0,0 +1 @@
+REACT_APP_LIVE_API_KEY=[[Your Live API Key]]
```
nmanu1, line 1: nit: can you also add a similar entry for `REACT_APP_MAPBOX_API_KEY`? looks like we forgot to add it previously for the test site

nmanu1 (commented):
besides the nit, lgtm!

nmanu1 (commented):
if we're making a new patch version for these fixes, we should probably merge into a hotfix branch rather than develop. but, personally I think we might be fine without because these are just changes to our test site

test-site/sample.env
```diff
@@ -0,0 +1,2 @@
...
+REACT_APP_MAPBOX_API_KEY=[[Your Live API Key]]
```
nmanu1, line 2: nit: this should probably read `Your Mapbox API Key` instead of live api key

PR #423, opened 2024-02-01
Support React 18

jesuyedavid:
The following changes were made to support React 18:
- Updated `react` and `react-dom` dev dependency versions from 17.0.2 to 18.2.0
- Updated peer dependencies to allow React 18
- Updated `react-collapsed` from v3.6.0 to v4.1.2 which supports React 18
- Copied `@reach/auto-id`'s `useId` hook  with minor modifications from https://github.com/reach/reach-ui/blob/dev/packages/auto-id/src/reach-auto-id.ts. Because the peer dependencies of that library don't allow React 18, we weren't able to install it directly anymore. This `useId` hook uses React 18's  `useId` hook if it's present and otherwise falls back to custom logic for React 16 and 17.
- Added a new `renderPin` prop to `MapboxMap` to allows React 18 users to render custom JSX into the DOM node for the marker element. This gets around the issue of having to call `createRoot` from `react-dom/client` in `MapboxMap`. This was causing issues on React 16/17 sites because Vite and Webpack were unable to determine that the dynamic import of `react-dom/client` wouldn't be reachable, and tried to resolve the package unsuccessfully. Rather than force existing consumers of the repo to update their bundler config to exclude `react-dom/client`, we provide this optional prop to React 18 users if they want to use React 18 features in their custom pin component, or if they don't want to get the console warning about using `ReactDOM.render` if they use the `PinComponent` prop.

To support testing in React 18:
- We are now fetching TextEncoder from utils when setting up the test environment because the new react-dom does not have it
- `@testing-library/react-hooks` doesn't support React 18 and it's functionality has now been added to `@testing-library/react` in v14. Because of this, `@testing-library/react` was upgraded to v14.2.1, which only supports React 18, and `@testing-library/react-hooks` was removed from the dev dependencies. The GitHub workflow for testing React 16 and 17 was updated to manually downgrade `@testing-library/react` to v12 and install `@testing-library/react-hooks`. Custom logic is added so that if `@testing-library/react-hooks` is installed, it's `renderHook` method is used (i.e. React 16/17), and otherwise the `renderHook` method from `@testing-library/react` is used (React 18).

J=BACK-2911, BACK-2923
TEST=auto, manual

Updated the test-site to use the new React 18 app initialization and tested all updated components manually to ensure they're not broken. Did a local npm pack of the repo and successfully used it in Pages repos with React 17 and 18.

Some updates were also made to clean up our tests:
- Upgraded `@testing-library/user-event` from v13.5.0 to v14.5.2, which makes async calls, so we no longer need `waitFor`. As part of this change, we started getting some new console errors when trying to navigate in tests by clicking links. This is because navigation is not defined in the jest jsdom environment. A util function was added to mock the console to remove these errors.
- Updated the GitHub test workflow for React 16 and 17 to manually install optional swc libraries because the post install script doesn't seem to run properly after we install specific versions of react-related packages. More info at https://github.com/swc-project/swc/issues/5616#issuecomment-1651214641


benmcginnis:
Seems like we need to fix all of the workflows

yext-search-ui-react-1.4.1.tgz
benmcginnis, line 1: I think this file was committed by accident

.github/workflows/run-tests.yml
```diff
@@ -6,18 +6,28 @@ name: Run Tests
...
+    uses: yext/slapshot-reusable-workflows/.github/workflows/run_tests.yml@v1
+    with:
+      node_matrix: '["16.x", "18.x"]'
+    secrets:
+      MAPBOX_API_KEY: ${{ secrets.MAPBOX_API_KEY }}
   call_run_tests-react-17:
     uses: yext/slapshot-reusable-workflows/.github/workflows/run_tests.yml@v1
     with:
+      build_script: |
+        npm i -D react@17 react-dom@17
```
nmanu1, line 19: you'll have to do the `@testing-library/react-hooks` install in this step so there aren't peer dependency conflicts (e.g. something like `npm i -D react@17 react-dom@17 @testing-library/react-hooks`). also you should choose a specific major version of `@testing-library/react-hooks` to pin to so that we don't get any breaking changes if they release a new major version of the package

package-lock.json
nmanu1, line 1: can you revert the package-lock changes and try running npm install again? it looks like the whole lockfile was regenerated here so we're seeing the semgrep issues from package updates that we don't need

.github/workflows/run-tests.yml
```diff
@@ -6,18 +6,47 @@ name: Run Tests
 on: [push, pull_request]
 
 jobs:
+  call_run_tests-react-18:
+    uses: yext/slapshot-reusable-workflows/.github/workflows/run_tests.yml@v1
+    with:
+      build_script: |
+        npx playwright install
```
nmanu1, line 13: do you know what change is making us have to install the playwright browsers now when we didn't need to before?

package.json
```diff
@@ -72,12 +72,11 @@
...
     "@storybook/react": "^7.4.5",
     "@storybook/react-webpack5": "^7.4.5",
-    "@storybook/test-runner": "^0.15.2",
+    "@storybook/test-runner": "^0.16.0",
     "@storybook/testing-library": "^0.2.1",
-    "@testing-library/jest-dom": "^5.16.1",
-    "@testing-library/react": "^12.1.3",
-    "@testing-library/react-hooks": "^8.0.0",
-    "@testing-library/user-event": "^13.5.0",
+    "@testing-library/jest-dom": "^5.12.0",
```
nmanu1, line 77: why are we pinning to a lower jest-dom version?

src/components/Dropdown/Dropdown.tsx
```diff
@@ -17,7 +17,10 @@ import { ScreenReader } from '../ScreenReader';
 import { recursivelyMapChildren } from '../utils/recursivelyMapChildren';
 import { DropdownItem, DropdownItemProps, DropdownItemWithIndex } from './DropdownItem';
 import useIsomorphicLayoutEffect from 'use-isomorphic-layout-effect';
-import { useId } from '@reach/auto-id';
+import { TextEncoder} from 'util';
+import { useId } from "react-id-generator";
+
+global.TextEncoder = TextEncoder;
```
nmanu1, line 23: why do we need this? I'm not seeing it being used in the component

src/components/Filters/CheckboxOption.tsx
```diff
@@ -137,7 +138,7 @@ export function CheckboxOption(props: CheckboxOptionProps): JSX.Element | null {
           onChange={handleChange}
           disabled={isOptionsDisabled}
         />
-        <label className={labelClasses} htmlFor={optionId}>{labelText}</label>
+        <label className={labelClasses} htmlFor={optionId}>{String(labelText)}</label>
```
nmanu1, line 141: do we need this change?

src/components/Filters/CheckboxOption.tsx
```diff
@@ -78,7 +79,7 @@ export function CheckboxOption(props: CheckboxOptionProps): JSX.Element | null {
     resultsCount
   } = props;
   const cssClasses = useComposedCssClasses(builtInCssClasses, props.customCssClasses);
-  const optionId = useId();
+  const [optionId] = useId();
```
nmanu1, line 82: out of curiosity, what else does this `useId` method return besides the id?

src/components/Filters/FilterGroupContext.ts
```diff
@@ -1,5 +1,36 @@
-import { createContext, useContext } from 'react';
-import { UseCollapseOutput } from 'react-collapsed/dist/types';
+import { createContext, useContext, CSSProperties, ReactNode, TransitionEvent, MouseEvent } from 'react';
+
+
+declare type AriaBoolean = boolean | 'true' | 'false';
+
+export interface GetTogglePropsOutput {
```
nmanu1, line 6: why are we redefining all the react-collapsed types? can't we get them from the library?

nmanu1 (commented):
have you spun up the test-site to make sure it's working properly? I thought we'd need to change some things in there because I imagine we're using ReactDOM.render

tests/tsconfig.json
```diff
@@ -3,6 +3,7 @@
   "include": [
     "components",
     "hooks",
-    "ssr"
+    "ssr",
+    "./jest-setup.ts",
```
nmanu1, line 7: which file is this?

tests/components/Geolocation.test.tsx
```diff
@@ -244,9 +281,10 @@ describe('default click handler', () => {
   });
 });
 
-function clickUpdateLocation() {
+async function clickUpdateLocation() {
+  const user = userEvent.setup();
   const updateLocationButton = screen.getByRole('button');
-  userEvent.click(updateLocationButton);
+  await user.type(updateLocationButton, '{enter}');
```
nmanu1, line 287: is there a reason we can't use click anymore?

src/components/Filters/FilterGroupContext.ts
jesuyedavid: on the latest version of react-collapse, they removed the /dist/types folder that had all the types laid out. so i had to redefine them to get the types.

tests/components/AppliedFilters.test.tsx
```diff
@@ -1,5 +1,5 @@
-import { render, screen } from '@testing-library/react';
-import userEvent from '@testing-library/user-event';
+import { render, screen, waitFor } from '@testing-library/react';
+import { userEvent } from '@testing-library/user-event';
```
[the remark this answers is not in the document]
jesuyedavid: i had tried that and it didn't work that's why i settled for setting userEvent first. found this solution somewhere online.

tests/tsconfig.json
jesuyedavid: may have been mistakenly committed. removing.

.github/workflows/run-tests.yml
jesuyedavid: not sure. does github workflow usually have it installed somewhere? if so, maybe a version got upgraded.

src/components/Dropdown/Dropdown.tsx
jesuyedavid: when i ran test, i was getting TextEncoder undefined. seems like this was a fix other people used.

src/components/Filters/CheckboxOption.tsx
jesuyedavid: looks like it returns a string array

.github/workflows/run-tests.yml
nmanu1: github workflows don't install it automatically, so my guess is that one of the storybook-related package upgrades you made included some change related to this. can you track that down in their release notes to see if we can avoid this somehow?

src/components/Filters/CheckboxOption.tsx
nmanu1: what would the other strings in the array be used for? are we sure we want the first one? also, are we sure the array won't be empty?

src/components/Filters/FilterGroupContext.ts
nmanu1: they don't export the types from anywhere anymore? I thought their release notes indicated they had added better types.

if they don't define them anywhere, then can we derive the types from `useCollapse` rather than hard-coding them ourselves? hard-coding the types could be problematic if they ever change them in the future, because we'd have to keep checking and remain in sync with the library

tests/components/AppliedFilters.test.tsx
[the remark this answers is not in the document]
nmanu1: tried it locally and it seems to work for me. what error are you getting?

src/components/Filters/CheckboxOption.tsx
nmanu1: I see. since the error is for NumberRangeValue, what does it look like in the UI if we coerce labelText of type NumberRangeValue to be a string? if it looks ok, maybe we can call toString when defining labelText. if it doesn't, we should either omit the display name or come up with an alternative that looks better

tests/components/Dropdown.test.tsx
```diff
@@ -1,11 +1,20 @@
...
+import  {userEvent} from '@testing-library/user-event';
 import { Dropdown, DropdownProps } from '../../src/components/Dropdown/Dropdown';
 import { DropdownInput } from '../../src/components/Dropdown/DropdownInput';
 import { DropdownMenu } from '../../src/components/Dropdown/DropdownMenu';
 import { DropdownItem } from '../../src/components/Dropdown/DropdownItem';
+import { TextEncoder} from 'util';
+
+//due to react-dom not having TextEncoder, we need to mock it
+//before importing testSSR
+global.TextEncoder = TextEncoder;
```
nmanu1, line 12: can we move this into `setup-env.ts`? that way we won't need to do this in multiple test files

package.json
```diff
@@ -153,19 +152,21 @@
...
     "mapbox-gl": "^2.9.2",
     "prop-types": "^15.8.1",
-    "react-collapsed": "3.6.0",
+    "react-collapsed": "^4.1.2",
+    "react-id-generator": "^3.0.2",
     "recent-searches": "^1.0.5",
     "tailwind-merge": "^1.3.0",
-    "use-isomorphic-layout-effect": "^1.1.2"
+    "use-isomorphic-layout-effect": "^1.1.2",
+    "util": "^0.12.5"
```
nmanu1, line 169: since this is just used in our tests, can we move this to devDependencies?

jesuyedavid: done

tests/components/Dropdown.test.tsx
jesuyedavid: done

tests/components/AppliedFilters.test.tsx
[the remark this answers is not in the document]
jesuyedavid: works for me now! i wonder if it's cause i updated user-event after

src/components/Filters/CheckboxOption.tsx
jesuyedavid: i took a look at the code and it seems that it always returns one element in the array if no value is being passed into useId() as we are currently doing so i think we should be fine here.

src/components/Filters/CheckboxOption.tsx
jesuyedavid: i did `toString` and it seems to work fine in the test site

.github/workflows/run-tests.yml
jesuyedavid: looks like  storybook@^7.4.5 and @storybook/test-runner@^0.15.2  that we are using for the react17 and 16 test runners are the same versions we have been using before this change. it may be difficult to figure out which dependency exactly is not installing playwright.

test-site/package.json
```diff
@@ -8,15 +8,14 @@
...
-    "react-dom": "file:../node_modules/react-dom",
     "react-router-dom": "^6.2.2",
     "uuid": "^9.0.0"
   },
   "devDependencies": {
     "@types/mapbox-gl": "^2.7.6",
     "@types/node": "^16.11.26",
-    "@types/react": "^17.0.42",
-    "@types/react-dom": "^17.0.14",
+    "@types/react": "^17.0.75",
```
nmanu1, line 17: should these types for react and react-dom be updated to v18?

.github/workflows/run-tests.yml
nmanu1: using storybook@^7.4.5 is different from storybook@7.4.5. by pinning to the major, we're likely using the same version for react 16, 17, and 18 tests (v7.6.13) assuming that's the latest storybook version. you could try pinning to v7.4.5 by installing `storybook@7.4.5` and removing `npx playwright install` to see if it works.

is the storybook version bump needed at all? we're still pinned to the same minor so I'd be surprised if the change was significant enough to block react 18 support

tests/__setup__/setup-env.ts
```diff
@@ -1,5 +1,11 @@
 import '@testing-library/jest-dom';
 import { server } from './server';
+import { TextEncoder} from 'util';
+
+//due to react-dom not having TextEncoder, we need to mock it
+//before importing testSSR
```
nmanu1, line 6: nit: we can probably remove this comment since it's not relevant here

tests/components/AppliedFilters.test.tsx
nmanu1, line 2: I think we can just use the default export of this library instead of the named export here and in other test files

tests/components/Dropdown.test.tsx
```diff
@@ -101,19 +102,19 @@ describe('Dropdown', () => {
...
+    await userEvent.click(inputNode);
     const itemNode = screen.getByText('item1');
 
-    userEvent.keyboard('{Tab}');
+    await userEvent.keyboard('{Tab}');
     expect(itemNode.className).toContain('FocusedItem1');
     expect(inputNode).toHaveValue('item1');
 
-    userEvent.keyboard('{Shift}{Tab}');
+    await userEvent.keyboard('{Shift>}{Tab}{/Shift}');
```
nmanu1, line 112: why is this change needed?

package.json
```diff
@@ -104,7 +104,8 @@
     "tailwindcss": "^3.0.23",
     "ts-jest": "^29.1.1",
     "tsup": "^8.0.1",
-    "typescript": "^5.2.2"
+    "typescript": "^5.2.2",
+    "util": "^0.12.5"
```
nmanu1, line 108: doesn't look like the package-lock was updated after this change

src/components/Filters/CheckboxOption.tsx
```diff
@@ -118,7 +118,7 @@ export function CheckboxOption(props: CheckboxOptionProps): JSX.Element | null {
 
   const isSelected = existingStoredFilter ? existingStoredFilter.selected : false;
 
-  const labelText = resultsCount ? `${displayName} (${resultsCount})` : displayName;
+  const labelText = (resultsCount ? `${displayName} (${resultsCount})` : displayName).toString();
```
nmanu1, line 121: we don't need to call toString on this whole statement. the first part of the ternary already returns a template string. also, if displayName was of type NumberRangeValue, wouldn't this display `[Object object]` or something instead of a useful label?

package.json
```diff
@@ -153,16 +153,17 @@
     "restoreMocks": true
   },
   "dependencies": {
-    "@reach/auto-id": "^0.18.0",
     "@restart/ui": "^1.0.1",
     "@tailwindcss/forms": "^0.5.0",
+    "@types/react": "^18.2.55",
```
nmanu1, line 158: out of curiosity, why didn't we need this package before?

tests/__fixtures__/data/filtercontext.ts
```diff
@@ -26,22 +45,8 @@ export const filterGroupContextValue: FilterGroupContextType = {
...
-    disabled: false,
-    type: 'button',
-    role: 'role',
-    id: 'id',
-    'aria-controls': 'element',
-    'aria-expanded': 'true',
-    tabIndex: 0,
-    onClick: () => null
-  }),
+  getCollapseProps: getDefaultCollapseProps(),
```
nmanu1, line 48: I thought `getCollapseProps` was a function. why do we assign it to the returned object here? shouldn't it be `getCollapseProps: getDefaultCollapseProps`, and similarly for `getToggleProps`?

nmanu1 (commented):
looks like some of the visual tests are still failing with the error "Bindings not found" (even though the check itself might be marked as passing)

also, I'm seeing some console errors in the unit tests that I don't think used to be there. can you make sure they're expected and mock them if needed?

tests/components/FieldValueDirectAnswer.test.tsx
```diff
@@ -13,16 +13,22 @@ import React from 'react';
...
     render(<FieldValueDirectAnswer
       result={fieldValueDAResult}
       viewDetailsClickHandler={viewDetailsClickHandler}/>
     );
-    userEvent.click(screen.getByRole('link', { name: 'View Details' }));
-    expect(viewDetailsClickHandler).toHaveBeenCalledTimes(1);
+    const viewDetailsLink = screen.getByRole('link', { name: 'View Details' });
+    await userEvent.click(viewDetailsLink);
+
+    waitFor(() => {
```
nmanu1, line 29: now that we're awaiting all the user events, can we remove all the waitFor's in our tests?

tests/components/Geolocation.test.tsx
```diff
@@ -220,6 +220,43 @@ describe('default click handler', () => {
     });
   });
 
+  it('replaces existing location filters with a new location filter in static filters state', async () => {
+    mockAnswersState(mockedStateWithFilters);
+    const actions = spyOnActions();
+    render(<Geolocation />);
+    clickUpdateLocation();
```
nmanu1, line 227: this should be awaited. I think there are other usages in this file and in LocationBias.test.tsx that need to be updated too

package.json
jesuyedavid: done

package.json
jesuyedavid: removing

tests/__fixtures__/data/filtercontext.ts
jesuyedavid: we resulted to using FilterGroupProvider instead

tests/components/FieldValueDirectAnswer.test.tsx
jesuyedavid: done.

tests/components/Geolocation.test.tsx
jesuyedavid: done

tests/components/Dropdown.test.tsx
jesuyedavid: the keyboard wasn't triggering properly so i switched to this format

tests/components/RangeInput.stories.tsx
```diff
@@ -18,9 +18,12 @@ const meta: Meta<typeof RangeInput> = {
   },
   decorators: [(Story) => (
     <SearchHeadlessContext.Provider value={generateMockedHeadless()}>
-      <FilterGroupContext.Provider value={filterGroupContextValue}>
+      <FilterGroupProvider
+          fieldId={'123'}
+          defaultExpanded={true}
```
nmanu1, line 23: I think defaultExpanded is optional. can we remove this prop unless we need the behavior to be different from the default?

tests/ssr/utils.tsx
```diff
@@ -29,5 +29,8 @@ export function testSSR(App: ReactElement) {
 
   // hydrate a container whose HTML contents were rendered by ReactDOMServer
   render(App, { container, hydrate: true });
-  expect(unexpectedErrorCount).toEqual(0);
+  
+  await waitFor(() => {
```
nmanu1, line 33: we should remove this waitFor, especially since we didn't need it before

src/components/Dropdown/Dropdown.tsx
```diff
@@ -66,8 +67,17 @@ export function Dropdown(props: PropsWithChildren<DropdownProps>): JSX.Element {
     alwaysSelectOption = false
   } = props;
 
+  //reset the id when the component is re-rendered
```
nmanu1, line 70: we should pull this logic out into a custom hook. in that hook, we can first check if we have `useId` available through react 18 and use that before falling back to this logic if it's not present

test-site/src/index.tsx
```diff
@@ -1,12 +1,8 @@
...
 
-ReactDOM.render(
-  <React.StrictMode>
-    <App />
-  </React.StrictMode>,
-  document.getElementById('root')
-);
+const container = document.getElementById('root');
+const root = createRoot(container!);
+root.render(<App/>);
```
nmanu1, line 8: is there a reason we removed strict mode?

yext-search-ui-react-1.5.0.tgz
nmanu1, line 1: we can delete this file

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@yext/search-ui-react",
-  "version": "1.4.1",
+  "version": "1.5.0",
```
nmanu1, line 3: since we're publishing an alpha, can we bump this to `1.5.0-alpha.423` instead?

nmanu1 (commented):
the PR should point to develop rather than main

package.json
nmanu1: actually, we should revert this change and bump the package version in a separate pr since we have other changes in develop

test-site/package-lock.json
```diff
@@ -33,10 +33,9 @@
     },
     "..": {
       "name": "@yext/search-ui-react",
-      "version": "1.4.1",
+      "version": "1.5.0",
```
nmanu1, line 36: looks like the package lockfiles haven't been updated

nmanu1 (commented):
looks like we use ReactDOM.render in the MapboxMap. we'll need to update that use the local react version's rendering approach

src/components/MapboxMap.tsx
```diff
@@ -184,3 +186,31 @@ function getDefaultCoordinate<T>(result: Result<T>): Coordinate | undefined {
   }
   return yextDisplayCoordinate;
 }
+
+function renderPinComponentByReactVersion(
```
nmanu1, line 190:
```suggestion
const renderReact = (jsxElement: JSX.Element, container: Element) => {
  try {
    import('react-dom/client').then(({ createRoot }) => {
      const root = createRoot(container);
      root.render(jsxElement);
    });
  } catch {
    ReactDOM.render(jsxElement, container);
  }
}
```
this way the logic is generic and reusable and should hopefully get the tests passing. we could even pull this function out into a separate file. the method signature is also almost the same as react 16/17's ReactDOM.render, so it'll be a trivial change to switch over to using it in the code

tsconfig.json
```diff
@@ -10,7 +10,8 @@
     "declarationMap": true,
     "sourceMap": true,
     "jsx": "react",
-    "target": "es2017"
+    "target": "esnext",
+    "module": "esnext",
```
nmanu1, line 14: I think we want to avoid using `esnext` because this is a production project and es**next** implies that it'll have the experimental changes for esm that would be added in the next stable version. we should use a stable version that supports dynamic imports instead.

also, since we're now changing the tsconfig module to be specific to esm rather than cjs, we want to be extra sure to test in a cjs repo to make sure tsup is still able to correctly produce the cjs bundle

nmanu1 (approved):

PR #433, opened 2024-03-27
Update CODEOWNERS to Watson

EmilyZhang777:


vijay267 (approved):

vijay267 (changes requested):
The comment seems to say the Watson group doesn't exist? Any idea why that's the case?

vijay267 (approved):
As we talked about on slack, we'll try to see if one works by merging it and if not, try to fix it.

PR #432, opened 2024-03-14
Allow custom afterDropdownInputFocus on FilterSearch

EmilyZhang777:
This change adds a param so that callers can pass in a custom afterDropdownInputFocus method for FilterSearch's input box. This request was brought up to allow analytics event logging upon clicking on the input box.

This change also bump the version to 1.6.0-beta.432.

J=TECHOPS-11352
TEST=auto, manual

Ran `npm run test`. Made sure `test-site` still works without a custom afterDropdownInputFocus method.
Used my local `search-ui-react` in the client's repo. Verified clicking into the input box triggers an analytics event and selecting an option from the dropdown properly hides it.

https://drive.google.com/file/d/1-kIYMU5TQAB2CzLxhSFpfbIJFqWhRJAR/view?usp=sharing

PR #439, opened 2024-05-15
Version 1.6.0

sahilvaidya:
### Features
- Added a custom `afterDropdownInputFocus` prop to the `FilterSearch` component that is called after focus enters the input. This can be used, for example, to fire analytics events when a user clicks into the `FilterSearch` input. (#432)


nmanu1 (commented):
the prop isn't only for analytics events, so I think we can word it a bit more generically, something like:

Added a custom `afterDropdownInputFocus` prop to the `FilterSearch` component that is called after focus enters the input. This can be used, for example, to fire analytics events when a user clicks into the `FilterSearch` input.

sahilvaidya:
Adjusted wording, thanks

nmanu1 (approved):

PR #441, opened 2024-06-12
add generativeDirectAnswer support

anguyen-yext2:
J=CLIP-1226
TEST=auto,manual

added unit and visual regression test, ran and saw tests passed
added GDADisplay to test-site App and verified that the appropriate gda content shows up and the UI looks as expected
image (1)


src/components/GenerativeDirectAnswerDisplay.tsx
```diff
@@ -0,0 +1,137 @@
...
+  generativeDirectAnswerCitationTitle: 'font-bold',
+  generativeDirectAnswerCitationSnippet: 'line-clamp-2 text-ellipsis break-words',
+};
+
+/**
+ * Props for {@link GenerativeDirectAnswerDisplay}.
+ *
+ * @public
+ */
+export interface GenerativeDirectAnswerDisplayProps<T> {
```
yen-tt, line 45: what is the generic type for?

src/components/GenerativeDirectAnswerDisplay.tsx
```diff
@@ -0,0 +1,137 @@
...
+    } else if (verticalResults) {
+      return verticalResults.results;
+    }
+  }, [universalResults, verticalResults]);
+
+  const searchActions = useSearchActions();
+  const gdaResponse = useSearchState(state => state?.generativeDirectAnswer?.response);
+  const isLoading = useSearchState(state => state?.generativeDirectAnswer?.isLoading);
+  
+  React.useEffect(() => {
```
yen-tt, line 82: instead of using `useEffect`, can you use headless's `addListener` function to listen in changes for the results state instead?

src/components/GenerativeDirectAnswerDisplay.tsx
```diff
@@ -0,0 +1,137 @@
...
+ *
+ * @param props - {@link GenerativeDirectAnswerDisplayProps}
+ * @returns A React element for the generative direct answer display, or null if there is no generated answer
+ */
+export function GenerativeDirectAnswerDisplay<T>({
+  customCssClasses
+}: GenerativeDirectAnswerDisplayProps<T>): JSX.Element | null {
+  const cssClasses = useComposedCssClasses(builtInCssClasses, customCssClasses);
+
+  const isVertical = useSearchState(state => state.meta.searchType === SearchTypeEnum.Vertical);
```
yen-tt, line 63: I think checking either isVertical or isUniversal should be sufficient

src/components/GenerativeDirectAnswerDisplay.tsx
```diff
@@ -0,0 +1,137 @@
...
+interface GenerativeDirectAnswerProps<T> extends GenerativeDirectAnswerDisplayProps<T> {
+  searchResults: Result[] | undefined,
+  gdaResponse: GenerativeDirectAnswerResponse,
+  cssClasses: GenerativeDirectAnswerDisplayCssClasses
+}
+
+/**
+ * Renders the generative direct answer text and citations.
+ */
+function renderGenerativeDirectAnswer<T>(props: GenerativeDirectAnswerProps<T>): JSX.Element {
```
yen-tt, line 108: nit: can we make this and below React function components instead

src/components/GenerativeDirectAnswerDisplay.tsx
```diff
@@ -0,0 +1,137 @@
...
+import React from 'react';
+
+/**
+ * The CSS class interface used for {@link GenerativeDirectAnswerDisplay}.
+ *
+ * @public
+ */
+export interface GenerativeDirectAnswerDisplayCssClasses {
+  generativeDirectAnswerTitle?: string,
+  generativeDirectAnswerText: string,
```
yen-tt, line 20: this and below should be optional?

src/components/GenerativeDirectAnswerDisplay.tsx
```diff
@@ -0,0 +1,137 @@
...
+
+/**
+ * Displays the AI generated answer of a generative direct answer.
+ *
+ * @public
+ *
+ * @param props - {@link GenerativeDirectAnswerDisplayProps}
+ * @returns A React element for the generative direct answer display, or null if there is no generated answer
+ */
+export function GenerativeDirectAnswerDisplay<T>({
```
yen-tt, line 58: I believe most, if not all, BlahDisplay components are meant to be internal. the public exported component would just be GenerativeDirectAnswer. see index.ts

yen-tt (commented):
UI wise, I think we can follow the figma mocks just a bit closer. I think it would be better to use `Sources` instead of `Citations`, and we should probably display the count. And how does it look when there are a lot of sources? would it scroll or keep stacking?


For the example in the image, does the result just not have markdown? the format seems a little off...could be from the conversions we discussed on Search side that dropped the markdown.

src/components/GenerativeDirectAnswerDisplay.tsx
```diff
@@ -0,0 +1,137 @@
...
+ *
+ * @public
+ */
+export interface GenerativeDirectAnswerDisplayCssClasses {
+  generativeDirectAnswerTitle?: string,
+  generativeDirectAnswerText: string,
+  generativeDirectAnswerDivider: string,
+  generativeDirectAnswerContainer?: string,
+  generativeDirectAnswerCitations: string,
+  generativeDirectAnswerCitation?: string,
```
yen-tt, line 24: It may not be necessary to include `generativeDirectAnswer` prefix since the component is already named that, can you check the patterns from other existing components to see the naming format? I think only some have the component name prefix

src/components/GenerativeDirectAnswerDisplay.tsx
anguyen-yext2: I have renamed this to GenerativeDirectAnswerProps instead. It's the props to GDA component. Currently, its fields include the css classes, answer header, and citations header so user can customize the UI.

src/components/GenerativeDirectAnswerDisplay.tsx
anguyen-yext2: I have tried the `addListener` and unsure of why its callback function is triggered even when there is no state change. I'm going to try digging into search-headless to see if I can find the cause. I'll update this as soon as I figure out what's wrong with `addListener`

src/components/GenerativeDirectAnswerDisplay.tsx
anguyen-yext2: Agree. I'll remove it when I switch from `useEffect` to `addListener`

src/components/GenerativeDirectAnswerDisplay.tsx
anguyen-yext2: yes they should be optional

anguyen-yext2:
This is what it looks like when there are a lot of citations (you need to scroll horizontally to see more). 
image

I want to note one thing I couldn't achieve using tailwinds - I wanted to either apply a `overflow-wrap: anywhere;` or `word-break: break-word`; rule to accomplish pic1. However, I think in Tailwinds I can only do a `break-all` (`word-break: break-all;`), which looks odd because it goes to new line in mid-word (pic2). But if I don't apply any break word class, then it looks odd when there is a long word because the citation div just expands larger than others (pic3).

pic1: image
pic2: 
image
pic3: 
image

src/components/GenerativeDirectAnswerDisplay.tsx
yen-tt: I still don't follow what is the generic type (<T>) is for here.
Screenshot 2024-06-13 at 6 43 13 PM
Unless I missed it somewhere, in your latest update, the type T is not use in any of the fields, so I believe you can remove this throughout GDA component

src/components/GenerativeDirectAnswer.tsx
```diff
@@ -0,0 +1,176 @@
...
+    if (!searchId || !searchTerm) {
+      return;
+    } 
+    if (isVertical) {
+      if (!verticalResults) {
+        return;
+      }
+    } else if (!universalResults) {
+      return;
+    }
```
yen-tt, line 97: can this be `if (!searchResults) return?`

src/components/GenerativeDirectAnswer.tsx
```diff
@@ -0,0 +1,176 @@
...
+  const searchResults: Result[] | undefined = React.useMemo(() => {
+    if (universalResults) {
+      return universalResults.verticals?.flatMap(v => v.results);
+    } else if (verticalResults) {
+      return verticalResults.results;
+    }
+  }, [universalResults, verticalResults]);
+
+  const searchActions = useSearchActions();
+  const gdaResponse = useSearchState(state => state?.generativeDirectAnswer?.response);
```
yen-tt, line 84: the first optional chain is not needed since state _should_ always be defined (here and below)

src/components/GenerativeDirectAnswer.tsx
```diff
@@ -0,0 +1,176 @@
...
+    executeGenerativeDirectAnswer(searchActions);
+  }, [isVertical, universalResults, verticalResults, searchId]);
+
+  if (isLoading || !gdaResponse || gdaResponse.resultStatus !== 'SUCCESS') {
+    return null;
+  }  
+
+  return (
+    <div className={cssClasses.container}>
+      {Answer({ gdaResponse, cssClasses, answerHeader })}
```
yen-tt, line 107: can you use component format here and for Citations? (e.g. `<Answer .../>`)

src/components/GenerativeDirectAnswer.tsx
```diff
@@ -0,0 +1,176 @@
...
+      <div className={cssClasses.divider} />
+      {Citations({ gdaResponse, cssClasses, citationsHeader, searchResults })}
+    </div>
+  );
+}
+
+interface AnswerProps<T> extends GenerativeDirectAnswerProps<T> {
+  gdaResponse: GenerativeDirectAnswerResponse,
+  cssClasses: GenerativeDirectAnswerCssClasses,
+  answerHeader: string | JSX.Element | undefined,
```
yen-tt, line 117: nit: you can use `answerHeader?:` instead of `| undefined` (here and below)

src/components/GenerativeDirectAnswer.tsx
```diff
@@ -0,0 +1,176 @@
...
+ */
+function Citations<T>(props: CitationsProps<T>) {
+  const { 
+    gdaResponse, 
+    cssClasses,
+    citationsHeader = `Sources (${gdaResponse.citations.length})`,
+    searchResults
+  } = props;
+  return <>
+    {gdaResponse.citations.length > 0 && searchResults && <>
```
yen-tt, line 155: nit: pull this check out into an if statement and return null -- easier to read

src/components/GenerativeDirectAnswer.tsx
```diff
@@ -0,0 +1,176 @@
...
+  const searchActions = useSearchActions();
+  const gdaResponse = useSearchState(state => state?.generativeDirectAnswer?.response);
+  const isLoading = useSearchState(state => state?.generativeDirectAnswer?.isLoading);
+  
+  React.useEffect(() => {
+    if (!searchId || !searchTerm) {
+      return;
+    } 
+    if (isVertical) {
+      if (!verticalResults) {
```
EmilyZhang777, line 92: nit, can be combined. The next `else if` would need to be changed to `(!isVertical && !universalResults)`

src/components/GenerativeDirectAnswer.tsx
```diff
@@ -0,0 +1,176 @@
...
+}
+
+/**
+ * The answer section of the Generative Direct Answer.
+ */
+function Answer<T>(props: AnswerProps<T>) {
+  const { 
+    gdaResponse, 
+    cssClasses,
+    answerHeader = 'AI Generated Answer',
```
EmilyZhang777, line 127: do we need to translate this?

src/components/GenerativeDirectAnswer.tsx
```diff
@@ -0,0 +1,167 @@
...
+
+const builtInCssClasses: Readonly<GenerativeDirectAnswerCssClasses> = {
+  container: 'p-6 border border-gray-200 rounded-lg shadow-sm',
+  header: 'text-xl',
+  answerText: 'mt-4',
+  divider: 'border-b border-gray-200 w-full pb-6 mb-6',
+  citationsContainer: 'mt-4 flex overflow-x-auto gap-4',
+  citation: 'p-4 border border-gray-200 rounded-lg shadow-sm bg-slate-100 flex flex-col grow-0 shrink-0 basis-64 text-sm text-neutral',
+  citationTitle: 'font-bold',
+  citationSnippet: 'line-clamp-2 text-ellipsis break-words'
```
yen-tt, line 36: per discussion offline, we should update the overflow-x css for the parent div to achieve the desired result

src/components/GenerativeDirectAnswer.tsx
```diff
@@ -0,0 +1,167 @@
...
+
+  const searchActions = useSearchActions();
+  const gdaResponse = useSearchState(state => state.generativeDirectAnswer?.response);
+  const isLoading = useSearchState(state => state.generativeDirectAnswer?.isLoading);
+  
+  React.useEffect(() => {
+    if (!searchResults?.length) {
+      return;
+    }
+    executeGenerativeDirectAnswer(searchActions);
```
yen-tt, line 87: if there's an error for the request, the old gda results would still be saved and rendered right? I think we should clear the result on failure

anguyen-yext2: would you want me to update search-headless so that the generative direct answer state is cleared if there is an error executing gda? 
When adding the gda endpoint, I was copying the flow of the other endpoints like executeUniversalQuery, in which the universal state was not cleared upon error. I tried forcing executeUniversalQuery to return `Promise.reject()`, but the old universal result was still rendered on the test page.

yen-tt: I think updating in headless make sense. If it handles state update on success, it should handle on failure as well.

src/components/GenerativeDirectAnswerDisplay.tsx
yen-tt: this may be a bug for Search to investigate

tests/components/GenerativeDirectAnswer.test.tsx
```diff
@@ -0,0 +1,53 @@
+import { render, screen } from '@testing-library/react';
+
```
yen-tt, line 2: you may also want to add some storybook stories (some of which would be viewable by users of SDK on this [page](https://storybook.hitchhikers.yext.com/?path=/docs/searchbar--docs)).

anguyen-yext2: will do in a followup PR

yen-tt (approved):

PR #444, opened 2024-06-26
support clickable citation card with link

anguyen-yext2:
J=CLIP-1332
TEST=auto,manual

ran `npm run test` and tested manually on test-site
https://github.com/yext/search-ui-react/assets/143001514/e9fb1baf-8423-4e9c-acf5-ea54f88edd54


test-site/src/pages/UniversalPage.tsx
```diff
@@ -92,7 +87,13 @@ export default function UniversalPage(): JSX.Element {
         customCssClasses={customSearchBarCss}
       />
       <SpellCheck />
-      <GenerativeDirectAnswer answerHeader='A custom answer header' CitationCard={CustomCitationCard}/>
+      <GenerativeDirectAnswer 
+        answerHeader='A custom answer header' 
+        getCitationLink={searchResult => searchResult.rawData.link}
```
yen-tt, line 92: mm not sure if I'm on board with the need for these new props. I think we can update the default Citation to automatically provide a clickable title if `searchResult.rawData.link` is present.
Other customization can defer to passing a custom Citation card. Otherwise, I feel like we will be making the default Citation way too bloated in more and more props, and that defeats the purpose of supporting custom card. What do you think?

anguyen-yext2: I can ask if the PMs is ok with that. I agree with you, which is why we only had support for the custom citation card in the beginning. However, the PM really wanted an easier approach to customize the citation card so that's why I added these changes.

yen-tt (approved):

PR #456, opened 2024-08-12
Avoid map key conflict for document results

sahilvaidya:
We currently can't have VerticalResults for document verticals since two results from the same entity will cause a map key conflict. To resolve this, we are making the key the result ID and the result index which should always be unique.

J=WAT-4282
TEST=manual,auto

Created a test document search config on the test-site which previously failed to return VerticalResults properly due to the key error now returns the results and updates on a new search


Fondryext (approved):
LGTM! glad it was simple

PR #459, opened 2024-08-29
Function Vertical Support via search-core/headless upgrade

sahilvaidya:
This change upgrades the version for search-core and search-headless to catch the changes which add support for function verticals. It also adds a function vertical page to the test-site for easier testing. An image of that page is attached.

J=WAT-3474
TEST=auto,manual

Tested via test-site, saw responses returned were upper cased as expected by the plugin, and the source was FUNCTION_VERTICAL.

Screenshot 2024-08-29 at 4 03 26 PM


Fondryext (approved):
LGTM!

vijay267 (approved):

PR #457, opened 2024-08-12
Version 1.6.2

sahilvaidya:
**Fixes**
- Enables Document Verticals to use VerticalResult by adjusting VerticalResult map key to combine the result id and index avoiding conflicts #456

PR #462, opened 2024-09-19
v1.6.3

k-gerner:
Add CloudChoice functionality


Fondryext (commented):
Sorry for not noticing this earlier, but I realized we need to up the version to 1.6.3 in the `package.json` files and re-run `npm i` to update `package-lock` as well. I think you can do that directly in the hotfix branch instead of making another dev branch.
Also, this can wait until you go to merge, but fyi that for pulling into main I'd pref a commit message formatted similarly to:
https://github.com/yext/search-ui-react/pull/457

vijay267 (approved):

PR #475, opened 2024-11-13
Update rollup transitive package

k-gerner:
Address vulnerability in older versions of rollup.

Added the following to `package.json`:
```
"resolutions": {
    "rollup": "4.22.4"
  }
```

Then ran 
`npx npm-force-resolutions`
`npm install`

Then removed `resolutions` from `package.json`.

Ran tests and spun up test site to confirm build was working.


vijay267 (commented):
Why'd you remove the change in package.json itself?

PR #480, opened 2024-11-22
search-ui-react: Resolve Vulnerabilities

mkouzel-yext:
Regenerates test-site/package-lock.json to include the safe dependencies.

J=VULN-39418, VULN-39419
TEST=none

Saw that babel imports in package-lock.json and test-site/package-lock.json were for versions at or above the safe dependency.


mkouzel-yext:
Yes, I just incremented the package version in `./package.json`. If the only change to `package-lock.json` was its package version then that's fine - to be clear I basically just incremented the package version and ran `rm -rf node_modules package-lock.json && npm i` in `.` and `./test-site`.

Fondryext:
Since we're already doing this change, could you see if it would be easy to include https://yexttest.atlassian.net/browse/VULN-38477 as part of this? To reduce the number of patch versions. Thanks!

mkouzel-yext:
I looked through `test-site/package.json` and it now has a safe @babel/traverse version, so that other ticket should be fixed by this PR as well. I expect the issue was that the package-lock file hadn't been regenerated in `test-site`.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@yext/search-ui-react",
-  "version": "1.6.7",
+  "version": "1.6.8",
   "description": "A library of React Components for powering Yext Search integrations",
   "author": "slapshot@yext.com",
```
vijay267, line 5: Maybe should be a separate change, but shouldn't this say watson@yext.com?

vijay267:
I'm also a little confused by this change? Shouldn't there be some other changes upgrading the dependency causing a vulnerability like Jacob said? The test site is just a test site to be able to test search-ui-react so I'd expect the main repo's package.json to change more.

Also separately not sure why but seems like some of the github checks are failing.

package.json
Fondryext: I figure we might as well do it here, since otherwise it'd be yet another patch version lol.

Fondryext:
> I'm also a little confused by this change? Shouldn't there be some other changes upgrading the dependency causing a vulnerability like Jacob said? The test site is just a test site to be able to test search-ui-react so I'd expect the main repo's package.json to change more.
> 
> Also separately not sure why but seems like some of the github checks are failing.

To be clear @vijay267 , what happened was the normal package, the normal package-lock, and the test site package, all got updated in the previous version, but not the test site package lock. So that's what this change is.

vijay267:
Ah got it. That makes sense then. Thanks for the explanation Jacob!

package.json
vijay267: True

mkouzel-yext: Added in [new commit](https://github.com/yext/search-ui-react/pull/480/commits/ae9dcbef20f78a9045160407d5b7d682cc792b28)

vijay267 (approved):

Issue #481, opened 2025-01-09
Unable to Fully Override Default Classes in FilterSearch Component


hiroki-takezawa-vivojpn:
Hi team,

I hope this message finds you well. I would like to kindly request your assistance in addressing an issue related to the `FilterSearch` component.

In the `FilterSearch` component, when using `customCssClasses.inputElement` to specify custom classes, the default `builtInClasses` (e.g., `text-sm`, `bg-white`, `outline-none`, etc.) are always merged into the final class list. This behavior makes it impossible to fully override the styles.

The current implementation of `useComposedCssClasses` forces a merge between `builtInClasses` and `customCssClasses`, resulting in unintended styles being applied to the `inputElement`. This issue hinders achieving a fully custom design for the input element.

---

# Steps to Reproduce

1. Use the `FilterSearch` component with custom classes:
   ```tsx
   <FilterSearch
     customCssClasses={{
       inputElement: "search-input Locator-input js-locator-input",
     }}
     ...
   />
   ```
2. Inspect the generated HTML. You will see that `builtInClasses` are still merged into the final class list:
   ```html
   <input class="text-sm bg-white outline-none h-9 w-full p-2 rounded-md border border-gray-300 focus:border-primary text-neutral-dark placeholder:text-neutral search-input Locator-input js-locator-input">
   ```

---

# Expected Behavior

When `customCssClasses.inputElement` is specified:
1. It should completely replace the default `builtInClasses`.
2. An option to disable merging should be provided (e.g., `disableDefaultClasses`).

---

# Actual Behavior

Even when specifying `customCssClasses.inputElement`, the `builtInClasses` are forcibly merged. This results in unintended styles, as shown below:
```html
<input class="text-sm bg-white outline-none h-9 w-full p-2 rounded-md border border-gray-300 focus:border-primary text-neutral-dark placeholder:text-neutral search-input Locator-input js-locator-input">
```

---

# Proposed Solution

- Add an option to `useComposedCssClasses` to allow full override of `customCssClasses`:
  ```javascript
  function useComposedCssClasses(builtInClasses, customClasses, overwrite = false) {
    return useMemo(() => {
      if (overwrite && customClasses) {
        return customClasses;
      }
      // Existing merge logic...
    }, [builtInClasses, customClasses, overwrite]);
  }
  ```
- Alternatively, add a `disableDefaultClasses` option to the `FilterSearch` component to disable default styles.

---

# Additional Context

[...]

PR #482, opened 2025-01-10
Allow disabling built-in CSS classes for FilterSearch component

k-gerner:
### Add a new flag to FilterSearchProps named disableBuiltInClasses.
To address https://github.com/yext/search-ui-react/issues/481
If this property is true, and there are customCssClasses set, then
we will completely remove the builtin CSS classes from the
component.

_TEST=manual_
Spun up test site and set the property on a component. Saw the
classes get replaced as expected.


Fondryext (commented):
Don't forget to increment the version. Maybe a new minor version? Could be argued this is kinda like a bugfix though, so I could be fine with a patch. I see an older pull request with Max that hasn't been shipped. May be worth syncing with him real quick to see what the state of that is.

k-gerner:
I was thinking that I'd just roll up Max's change at the same time and keep the `v1.6.7` since that change hadn't been shipped yet, but I think you're actually right that it's more of a minor version change here. I'll update the version to `v1.7.0`. 

(The parent branch of this one will have the wrong name though but that's fine)

Fondryext (approved):

PR #484, opened 2025-01-29
search-ui-react: Resolve Vulnerabilities

mkouzel-yext:
Upgrades dependency versions to resolve potential vulnerabilities.


src/components/SearchBar.tsx
```diff
@@ -386,7 +386,7 @@ export function SearchBar({
     entityPreviewsCount
   );
   const activeClassName = classNames('relative z-10 bg-white border rounded-3xl border-gray-200 w-full overflow-hidden', {
-    ['shadow-lg' ?? '']: hasItems
+    ['shadow-lg']: hasItems
```
k-gerner, line 389: Why is this impacted?

package.json
```diff
@@ -36,7 +36,7 @@
   ],
   "scripts": {
     "build": "rm -rf lib/** && rm -rf dist/** && npm run build:js && npm run build:css && npm run api-extractor && npm run generate-docs && npm run generate-notices",
-    "build:js": "tsup src/index.ts --dts --sourcemap --format esm,cjs && npm run create:lib-esm && npm run create:lib-commonjs",
+    "build:js": "NODE_OPTIONS='--max-old-space-size=16384' tsup src/index.ts --dts --sourcemap --format esm,cjs && npm run create:lib-esm && npm run create:lib-commonjs",
```
k-gerner, line 39: Where's this coming from?

mkouzel-yext:
The reason for both of these changes is to try to get the automatic checks passing again. After making the package upgrades that resolving our vulnerabilities require, the majority of our automatic tests failed due to build errors. I'll reply with the individual reasons for each, but this is the overarching goal.

src/components/SearchBar.tsx
mkouzel-yext: Was causing `tsup src/index.ts --dts --sourcemap --format esm,cjs` to fail because `?? ''` is never used. Ultimately preventing `npm run build` from succeeding.

package.json
mkouzel-yext: The `tsup` build command was failing due to OOM error in its worker process. This isn't a permanent fix because as the repo grows the heap size may need to be bumped again, but the root of the problem lies with `tsup` itself (https://github.com/egoist/tsup/issues/920).

PR #485, opened 2025-02-10
Mark all icons as hidden for screen readers.

chrissnyder2337:
Because the SVGs are decorative, it does not need alt text, but the browser and accessibility tools need to know that they should not worry about it. 

See https://www.unimelb.edu.au/accessibility/techniques/accessible-svgs#:~:text=In%20order%20to%20hide%20an,to%20display%3Anone%20in%20CSS.


Fondryext (approved):
Thanks!

PR #486, opened 2025-02-10
Add analytics event for Generative Direct Answer card

anguyen-yext2:
J=WAT-4600,WAT-4604
TEST=auto,manual

added and ran unit test
tested manually on test-site:
- clicked on a citation and saw the citation click event logged in snowflake
- hardcoded a link in the gda text, clicked on the link, saw cta click event logged in snowflake


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@yext/search-ui-react",
-  "version": "1.7.0-beta.10",
+  "version": "1.8.0-beta",
```
k-gerner, line 3: I'm not familiar with the version naming but should this be `1.8.0-beta.1`?

src/components/GenerativeDirectAnswer.tsx
```diff
@@ -211,13 +219,36 @@ export interface CitationProps {
...
     </a>
   );
+}
+
+/**
+ * Payload for events fired on a generative direct answer card.
+ *
+ * @public
+ */
+export interface GenerativeDirectAnswerData {
```
k-gerner, line 244: nit: This is more like the data for a specific citation right? Can we change the name of the interface to reflect that more?

src/hooks/useCardAnalytics.ts
```diff
@@ -94,6 +107,9 @@ export function useCardAnalytics<T>(): (
     if (isDirectAnswer(result)) {
       directAnswer = true;
       entityId = result.relatedResult.id;
+    } else if (isGenerativeDirectAnswer(result)) {
```
k-gerner, line 110: should this also have `generativeDirectAnswer = true` ?

package-lock.json
```diff
@@ -82,6 +82,34 @@
         "react-dom": "^16.14 || ^17 || ^18"
       }
     },
+    "../analytics": {
```
mkouzel-yext, line 85: This seems off, and looks like it's related to the test failures. Why was this generated?

package.json
anguyen-yext2: I think there is no strict rule. I was following what was most recently done, which begins with x.x.x-beta and next up would be x.x.x-beta.2.
https://github.com/yext/search-core/pull/266#discussion_r1649387983

src/components/GenerativeDirectAnswer.tsx
anguyen-yext2: Yes, it is currently citation specific, but I kept it generic because I think we would add more to this interface if we support other analytics events for GDA that is not citation specific, for example the answer text or the citations list. 

Alternatively, I could rename this to CitationClickData and if we support other analytics events for GDA in the future, we need to define a separate type and add it to CardAnalyticsDataType. Would you prefer that?

src/hooks/useCardAnalytics.ts
anguyen-yext2: analytics event with feedbackType currently does not have the `generativeDirectAnswer` field yet, but yes we would need to set this to true if we support something like thumbs up event for GDA. For now, I modified this portion of the code to pass type check.

package-lock.json
anguyen-yext2: ohh good catch! Yes it was generated, so I didn't notice it. 
Running `npm i` alone didn't work. I ended up having to revert the `package-lock.json` file then run `npm i`.

mkouzel-yext (approved):

src/components/GenerativeDirectAnswer.tsx
k-gerner: I lean towards having a separate type, just because if we did add more events in the future, we would be leaving these fields blank anyways (e.g. thumbs up feedback, doesn't relate to entity or link). However, ultimately I think following whatever pattern is set by the original DirectAnswer makes the most sense. Does that interface support multiple events?

Separately, are we supporting rich text clicks here? E.g. if the generated answer has a link in the snippet, is that being tracked?

anguyen-yext2: ok I renamed to CitationClickData. 

I can't follow the pattern of DirectAnswer because the data stored in state (`useSearchState(state => state.directAnswer.result)`) has all the info to pass to the analytics event payload, so we don't have to define any new type. However, the data stored in GDA state (`useSearchState(state => state.generativeDirectAnswer?.response)`) is not sufficient to build the citation click event payload:
1. It has a list of citations but does not have an indication of which citation is being clicked
2. We don't know the destination url of the clicked citation
Therefore, I have to define a new type for citation click event data and also add a new hook, `useReportCitationClick`, instead of applying the same logic as DirectAnswer.

> Separately, are we supporting rich text clicks here? E.g. if the generated answer has a link in the snippet, is that being tracked?
Let me ask Baigel about this.

anguyen-yext2: Updated to handle link click from generated answer text

Fondryext (approved):
LGTM! (sorry again about the merge conflict)

Fondryext (approved):

PR #503, opened 2025-03-26
Update MapboxMap with VLE props for using inside an iframe

k-gerner:
In this PR, I have added two props to the `MapboxMap` component: `iframeWindow` and `allowUpdates`. These props were added to enable usage of this component in the Visual Layout Editor. Because we are using the component in an iframe, we need to use the iframe's `mapboxgl` library to properly reference the window. Additionally, the `allowUpdates` property is used so that we can dynamically change properties / styles of the map through the VLE. Currently, we only allow changing the "style" of the map (such as satellite view, dark mode, etc), but more can be added as necessary. With this property set to `true`, we will listen to updates to the mapbox options, and update when applicable. 

Separately, I have added a navigation button section, which contains `+` and `-` buttons for zooming in and out of the map.


Fondryext:
Could you add a little more context to the commit message and perhaps change the title? Going through Slack I think I understand the motivation here but it isn't as clear in the PR proper.

k-gerner:
> Could you add a little more context to the commit message and perhaps change the title? Going through Slack I think I understand the motivation here but it isn't as clear in the PR proper.

Updated!

Fondryext (approved):

PR #508, opened 2025-04-07
ksearch: upgrade to Events API

Fondryext:
This PR upgrades our analytics calls to the next major version, also called the Events API. As part of this work, analytics calls have changed shape, and some deprecated properties have been dropped.
J=WAT-4651
TEST=auto,manual

Updated auto tests. Ran test site locally and saw events all the way to snowflake.


src/hooks/useCardAnalytics.ts
```diff
@@ -77,16 +70,23 @@ export function useCardAnalytics<T>(): (
...
+      return;
+    }
     analytics?.report({
-      type: eventType,
-      entityId: entityId,
-      searcher: verticalKey ? 'VERTICAL' : 'UNIVERSAL',
-      queryId,
-      verticalKey: verticalKey || '',
-      url,
-      fieldName,
```
Fondryext, line 87: we drop 'searcher' and 'fieldName'

src/hooks/usePaginationAnalytics.ts
```diff
@@ -22,15 +21,20 @@ export function usePaginationAnalytics(): (
       console.error('Unable to report a pagination event. Missing field: verticalKey.');
       return;
     }
-    analytics.report({
-      type: 'PAGINATE',
-      queryId: queryId,
-      verticalKey: verticalKey,
-      newPage,
-      currentPage,
-      totalPageCount
```
Fondryext, line 31: We drop all three of these page-related values.

src/hooks/usePaginationAnalytics.ts
```diff
@@ -1,16 +1,15 @@
...
-import { useSearchState } from '@yext/search-headless-react';
+import {useAnalytics} from './useAnalytics';
+import {useSearchState} from '@yext/search-headless-react';
 
-export function usePaginationAnalytics(): (
-  newPage: number,
-  currentPage: number,
-  totalPageCount: number
-) => void {
+export function usePaginationAnalytics(): () => void {
```
Fondryext, line 4: This function changes signature

src/hooks/useSearchBarAnalytics.ts
```diff
@@ -1,44 +1,62 @@
...
+    const queryId = useSearchState(state => state.query.queryId);
+    const searchId = useSearchState(state => state.meta.uuid);
+    const locale = useSearchState(state => state.meta.locale);
+    const experienceKey = useSearchState(state => state.meta.experienceKey);
 
-  const reportAutocompleteEvent = (suggestedSearchText: string) => {
-    analytics?.report({
-      type: 'AUTO_COMPLETE_SELECTION',
-      ...(queryId && { queryId }),
-      suggestedSearchText
```
Fondryext, line 18: We drop suggestedSearchText

src/components/sections/SectionHeader.tsx
```diff
@@ -90,10 +93,19 @@ export function SectionHeader(props: SectionHeaderProps): JSX.Element {
       console.error('Unable to report a vertical view all event. Missing field: queryId.');
       return;
     }
+    if (!experienceKey) {
```
k-gerner, line 96: Do we need a check for `locale` or `searchId`? Or are those automatically set

src/hooks/useCardAnalytics.ts
```diff
@@ -77,16 +70,23 @@ export function useCardAnalytics<T>(): (
...
-      directAnswer,
-      generativeDirectAnswer
+      action: eventType,
+      destinationUrl: url,
+      entity: entityId,
+      locale,
+      search: {
+        searchId,
+        queryId,
+        ...verticalKey &&  {verticalKey: verticalKey},
```
k-gerner, line 85: Previously we used an empty string for `verticalKey`, so just wanted to confirm that excluding it is now the correct way to handle universal?

src/hooks/useCardAnalytics.ts
```diff
@@ -77,16 +70,23 @@ export function useCardAnalytics<T>(): (
...
+        isDirectAnswer: directAnswer,
+        isGenerativeDirectAnswer: generativeDirectAnswer,
```
k-gerner, line 87: As very minor cleanup we could rename `directAnswer` to `isDirectAnswer` etc to match the format of the rest of the `search` object here

src/hooks/useCardAnalytics.ts
```diff
@@ -77,16 +70,23 @@ export function useCardAnalytics<T>(): (
       console.error('Unable to report a CTA event. Missing field: queryId.');
```
k-gerner, line 70: Interesting that we require query id here but not search id. Is that because query id comes from `state.query` rather than `state.meta`?

src/hooks/useSearchBarAnalytics.ts
k-gerner: That's interesting, seems like it'd be a pretty relevant piece of data for this

src/hooks/useSearchBarAnalytics.ts
```diff
@@ -1,44 +1,62 @@
...
-      : reportSearchClearEvent();
-  };
-  return reportAnalyticsEvent;
+    const reportSearchClearEvent = () => {
+        if (!queryId) {
+            console.error('Unable to report a search clear event. Missing field: queryId.');
+            return;
+        }
+        if (!experienceKey) {
+            console.error('Unable to report a vertical view all event. Missing field: experienceKey.');
```
k-gerner, line 38: Should be `search clear event`. 
(Same issue in a few of the other files, but won't spam those comments so just leaving this one)

src/components/sections/SectionHeader.tsx
Fondryext: locale can be unset in the event payload so passing it possible undefined is ok. For searchId we could also add error handling if the searchId is not present.

src/hooks/useCardAnalytics.ts
Fondryext: Are you asking me why I didn't change the earlier behavior?

src/hooks/useSearchBarAnalytics.ts
Fondryext: How do you mean? Interesting that analytics dropped it from their new system, or that I followed their example?

src/hooks/useSearchBarAnalytics.ts
Fondryext: good catch, thanks!

src/hooks/useCardAnalytics.ts
Fondryext: good call, changed back

src/components/sections/SectionHeader.tsx
k-gerner: Hm as I think about it, our analytics are all based off query id now right? Probably not necessary to require the search id if we don't see that changing

src/hooks/useSearchBarAnalytics.ts
k-gerner: Interesting that analytics dropped it. We should ofc follow whatever their system supports / expects

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@yext/search-ui-react",
-  "version": "1.8.2",
+  "version": "1.8.3",
```
k-gerner, line 3: My debounce change is 1.8.3 so this will have to change - I assume to 2.0.0 since this has breaking changes

src/components/sections/SectionHeader.tsx
k-gerner: Got it, so it seems necessary, but also that it should be automatically set elsewhere correct? Therefore not requiring any extra check here?

package.json
Fondryext: Yes, this change is a draft.

src/components/sections/SectionHeader.tsx
Fondryext: I agree a check would be nice, and I could add it. Could you elaborate the second comment? What should be set, the searchId? Isn't that pulled from the meta state? I'm not sure we have any guarantees about that.

k-gerner: Sorry maybe I was misunderstanding your prior comment. The way I read it was that the search id was being set as the query id by default somewhere up the chain in the SDK somewhere. Feel free to disregard. I agree adding a check is good then

src/hooks/useAnalytics.ts
```diff
@@ -1,13 +1,13 @@
...
-export const AnalyticsContext = createContext<AnalyticsService | null>(null);
+export const AnalyticsContext = createContext<AnalyticsEventService | null>(null);
 
 /**
  * Returns a service that can be used to report analytics events.
  *
  * @public
  */
-export function useAnalytics(): AnalyticsService | null {
+export function useAnalytics(): AnalyticsEventService | null {
```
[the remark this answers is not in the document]
k-gerner: I think this would be a good idea given that we want to abstract away the analytics code like you said

src/components/SearchBar.tsx
```diff
@@ -243,7 +243,7 @@ export function SearchBar({
       executeQuery();
     }
     if (typeof index === 'number' && index >= 0 && !itemData?.isEntityPreview) {
-      reportAnalyticsEvent('AUTO_COMPLETE_SELECTION', value);
```
vijay267, line 246: Sorry what was the value we were submitting before but no longer need to?

src/hooks/useCardAnalytics.ts
```diff
@@ -98,40 +100,52 @@ export function useCardAnalytics<T>(): (
...
-      type: feedbackType,
-      entityId,
-      searcher: verticalKey ? 'VERTICAL' : 'UNIVERSAL',
-      queryId,
-      verticalKey: verticalKey || '',
-      directAnswer
+      action: feedbackType,
+      entity: entityId,
+      locale,
+        searchId,
```
vijay267, line 128: Hm is the different indenting between this line and the line before intentional

src/hooks/usePaginationAnalytics.ts
vijay267: Hm so we're just no longer sending that data with the event after this change? Or is that information picked up elsewhere?

src/models/SearchEventPayload.ts
```diff
@@ -0,0 +1,94 @@
...
+    versionNumber?: number;
+    /** The identifier of the search experience. */
+    experienceKey: string;
+    /** Whether or not the event occurred on a generative direct answer card */
+    isGenerativeDirectAnswer?: boolean;
+    /** The query entered by the user. */
+    searchTerm?: string;
+}
+
+//todo - how many of these do we want to delete?
```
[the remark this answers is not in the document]
vijay267: Is there a reason we want to prune these down? At the very least, I imagine we want things like map pin, auto complete selection, the paginates, chat impression, search clear, etc.

src/hooks/useAnalytics.ts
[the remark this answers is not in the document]
Fondryext: ended up doing this

src/models/SearchAnalyticsEventServiceImpl.ts
```diff
@@ -0,0 +1,47 @@
...
+        const eventPayload: EventPayload = {
+            action: payload.action,
+            destinationUrl: payload.destinationUrl,
+            entity: payload.entity,
+            locale: payload.locale,
+            search: {
+                searchId: payload.searchId,
+                queryId: payload.queryId,
+                verticalKey: payload.verticalKey,
+                isDirectAnswer: payload.isDirectAnswer,
```
vijay267: nit - Not sure if the order matters (if it does, then don't worry about this), but should we put isDirectAnswer at the end right before isGenerativeDirectAnswer to group similar properties together? I'd also have experience key right before vertical key.

src/models/SearchEventPayload.ts
```diff
@@ -0,0 +1,79 @@
...
+    action?: SearchAction;
+    /** The URL of the page the event is directing the visitor to. */
+    destinationUrl?: string;
+    /** The Yext entity to which the event corresponds. If passed as a string, the value is
+     * the mutable, customer-settable entity ID for the entity associated with the event.
+     * If passed as a number, it is the immutable entity ID (UID) set by the system. This is an internal ID.
+     */
+    entity?: string | number;
+    /** The locale of the user who generated the event.
+     */
```
vijay267: nit - Feels like the comment shouldn't wrap and should just be on 1 line.

tests/components/SearchBar.test.tsx
```diff
@@ -370,8 +372,12 @@ describe('SearchBar', () => {
       expect(await screen.findByRole('textbox')).toHaveDisplayValue('query suggestion');
       expect(mockedReport).toHaveBeenCalledTimes(1);
       expect(mockedReport).toHaveBeenCalledWith({
-        type: 'AUTO_COMPLETE_SELECTION',
-        suggestedSearchText: 'query suggestion'
+        action: 'AUTO_COMPLETE_SELECTION',
```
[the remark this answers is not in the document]
vijay267: Hm... I see because it's an autocomplete event there's no search. When you say fake a search id, are you suggesting the frontend just makes a unique id or something?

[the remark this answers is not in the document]
vijay267: Slack convo link - https://yext.slack.com/archives/GTTQTJH9B/p1767904911666969

vijay267 (commented):
Largely looks good! Some minor questions / pieces of feedback and the question around the searchId + autocomplete selection event.

src/models/SearchAnalyticsEventServiceImpl.ts
```diff
@@ -0,0 +1,47 @@
...
+    public async report(payload: SearchEventPayload): Promise<string> {
+        const eventPayload: EventPayload = {
+            action: payload.action,
+            destinationUrl: payload.destinationUrl,
+            entity: payload.entity,
+            locale: payload.locale,
+            search: {
+                searchId: payload.searchId,
+                queryId: payload.queryId,
+                verticalKey: payload.verticalKey,
```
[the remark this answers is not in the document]
Fondryext: verticalKey can be empty. Is that what you meant?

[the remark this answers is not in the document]
vijay267: Ah that also works. I just wanted to make sure events events for universal searches would still work.

tests/components/SearchBar.test.tsx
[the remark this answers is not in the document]
vijay267: For folks who don't have access to slack, it turns out this is handled fine (not having a searchId) so nothing extra is needed.

vijay267 (dismissed):
Looks good to me! Looks like some checks were failing though, so let's double check those.

k-gerner (approved):
Reviewing the typescript fixes only, since the previous changes were already approved by both me and Deepak

PR #511, opened 2025-04-11
v1.8.3: Add debouncing to filter search and autocomplete

k-gerner:
This adds debouncing to the autocomplete and filter search calls. Currently, this is set to 200ms. This means that we will not fire autocomplete or filter search requests until 200ms has elapsed since the last keystroke, in order to save on unnecessary queries while the user is still actively typing. This number was chosen based on the average typing speed being roughly 4-5 characters per second.

The relevant tests were modified by adding intentional pauses during execution of some tests, in order to allow the debouncing logic to have time to execute

J=[WAT-4708](https://yexttest.atlassian.net/browse/WAT-4708?atlOrigin=eyJpIjoiOWY5NmU0YzdmYmVhNGNjZDk5YWViNTE2MjBkZmQ2ZWYiLCJwIjoiaiJ9)


Fondryext:
I forget, has code coverage been broken?

k-gerner:
> I forget, has code coverage been broken?

Nope. Reran the test and it passed. Unfortunately it may be flaky sometimes due to the async nature of the debouncing. Maybe this could be fixed by increasing the wait time in tests by a good amount? Though it would make the tests take longer

Fondryext (approved):

PR #514, opened 2025-04-21
Deduplicate generative answer sources

mkouzel-yext:
If an entity is present in search results more than once (e.g. returned by two different verticals) and is cited by the GDA, it appeared in the sources more than one. This change dedups the search results while determining the sources to prevent this.

J=TECHOPS-13971
TEST=manual


src/components/GenerativeDirectAnswer.tsx
```diff
@@ -195,7 +195,17 @@ function Citations(props: CitationsProps) {
         return false;
       }
       return gdaResponse.citations.includes(uid);
-    })
+    }).filter((result, index, self) => {
```
k-gerner, line 198: Similar to the comment on the Gerrit CR, can we condense this by using e.g. a secondary `Set` that tracks which uids we haven't seen yet?

src/components/GenerativeDirectAnswer.tsx
```diff
@@ -195,7 +195,17 @@ function Citations(props: CitationsProps) {
...
+      // If an entity is returned by multiple different verticals, it will be present in
```
k-gerner, line 199: Can we add a test case for this as well

k-gerner (changes requested):
I believe we need to update the version in `package.json` right?

src/components/GenerativeDirectAnswer.tsx
mkouzel-yext: Updated

src/components/GenerativeDirectAnswer.tsx
mkouzel-yext: Added

mkouzel-yext:
> I believe we need to update the version in `package.json` right?

Yes, incremented

Fondryext:
> > I believe we need to update the version in `package.json` right?
> 
> Yes, incremented

FYI there's a flow set up which makes an increment version PR when hotifx branches are made, in this case: https://github.com/yext/search-ui-react/pull/513 . Since you've done the increment here just be sure to abandon that PR

mkouzel-yext:
Ahhh I was wondering why that PR got created

PR #522, opened 2025-05-22
Replace pinColor with markerOptionsOverride

anguyen-yext2:
This allows user to specify options other than the pin color as well. Updated test-site to specify a pin color and a pin size based on pin selection instead of passing a PinComponent. Confirmed it work

image


src/components/MapboxMap.tsx
```diff
@@ -104,13 +104,13 @@ export interface MapboxMapProps<T> {
   /** A function that scrolls to the search result corresponding to the selected pin. */
   scrollToResult?: (result: Result<T> | undefined) => void,
   /**
-   * The color that default map markers should be, in the form of a HEX color code.
-   * By default, the standard Mapbox pin color is used, which is light blue (#3FB1CE).
+   * The options to apply to the map markers based whether it is selected.
+   * By default, the standard Mapbox pin is used regardless of the pin selection.
```
k-gerner, line 108: What is meant by `is used regardless of the pin selection`?

src/components/MapboxMap.tsx
```diff
@@ -219,7 +219,7 @@ export function MapboxMap<T>({
         if (markerLocation) {
           const { latitude, longitude } = markerLocation;
           const el = document.createElement('div');
-          const markerOptions: mapboxgl.MarkerOptions = {};
+          let markerOptions: mapboxgl.MarkerOptions = {};
```
k-gerner, line 222: I think maybe this should be `mapboxInstance` rather than `mapboxgl`. See line 147

k-gerner: Actually I'm not sure. I see this was configured this way prior.

src/components/MapboxMap.tsx
```diff
@@ -237,12 +237,16 @@ export function MapboxMap<T>({
...
+          } else if (markerOptionsOverride) {
+            markerOptions = {
+              ...markerOptions,
+              ...markerOptionsOverride(selectedResult === result)
+            }
           }
           const marker = new mapboxInstance.Marker(markerOptions)
             .setLngLat({ lat: latitude, lng: longitude })
             .addTo(mapbox);
+          marker.getElement().addEventListener('click', () => handlePinClick(result));
```
k-gerner, line 249: will this conflict with the `onClick` we set on line 234 in the case of when `PinComponent` is present?

test-site/src/pages/LocationsPage.tsx
```diff
@@ -73,4 +102,31 @@ export function LocationsPage() {
       </div>
     </div>
   );
+
+  /**
+   * uncomment this to test side by side view of search results and interactive map pins
+   * note that scroll to results do not fully work due to extra height in the StandardCard padding
```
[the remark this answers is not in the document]
k-gerner: Got it. I think that's fine, but would you mind clarifying that this doesn't work specifically for the test site in this comment

src/components/MapboxMap.tsx
anguyen-yext2: I meant the default styling will be applied regardless of whether the pin is clicked or not. Let me just remove that so it's less confusing.

src/components/MapboxMap.tsx
anguyen-yext2: i think it should be `mapboxgl` bc it's just a type definition

test-site/src/pages/LocationsPage.tsx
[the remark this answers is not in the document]
anguyen-yext2: updated

k-gerner (approved):
LGTM assuming required tests pass

vijay267 (dismissed):

vijay267 (approved):

PR #526, opened 2025-06-09
Cleanup various event listeners.

anguyen-yext2:
Also make sure localizeMap() is executed when map style changes as well.

TEST=manual
verified map is localized upon style changes as well


mkouzel-yext (approved):

k-gerner (changes requested):
Version should now be 1.8.13

PR #548, opened 2025-07-11
Fix CurrentLocationIcon prop names

mkouzel-yext:
Incorrect names were throwing warnings.

Rebuilt test site with newly named props and warnings were no longer thrown.


k-gerner (commented):
I assume after this change, there were 0 warnings?

k-gerner (approved):

k-gerner (approved):

PR #554, opened 2025-08-04
Debounce generative direct answer calls

mkouzel-yext:
The effect hook that triggers a GDA call has two dependencies: the search results and the search ID. These both change when a search is made. In React 18, these updates are batched together so only one call to GDA gets made when the search is updated. However, in React 17, no batching is performed and updates are handled in order, meaning a call to GDA is made when the search results are updated and when the search ID is updated, even if the same search triggers both updates. This change tracks the search results and only executes a new GDA call if the search results have changed.

J=WAT-4861
TEST=manual

Created a local test site that runs on React 17. Made a local tarball with my changes to search-ui-react using npm pack and pointed the local test site's import of search-ui-react to the tarball. Confirmed that only one GDA call is made per-seasrch.

Also spun up the search-ui-react test-site (which runs on React 18) and confirmed that GDA calls are still made once per search as expected.

### DIFF
--- a/src/components/GenerativeDirectAnswer.tsx
+++ b/src/components/GenerativeDirectAnswer.tsx
@@ -11,7 +11,7 @@ import { useCardAnalytics } from '../hooks/useCardAnalytics';
 import { DefaultRawDataType } from '../models/index';
 import { executeGenerativeDirectAnswer } from '../utils/search-operations';
 import { Markdown, MarkdownCssClasses } from './Markdown';
-import React, { useMemo } from 'react';
+import React, { useMemo, useState } from 'react';
 
 /**
  * The CSS class interface used for {@link GenerativeDirectAnswer}.
@@ -92,17 +92,19 @@ export function GenerativeDirectAnswer({
     }
   }, [isUniversal, universalResults, verticalResults]);
 
+  const [lastExecutedSearchResults, setLastExecutedSearchResults] = useState(searchResults);
   const searchActions = useSearchActions();
   const gdaResponse = useSearchState(state => state.generativeDirectAnswer?.response);
   const isLoading = useSearchState(state => state.generativeDirectAnswer?.isLoading);
   const handleClickEvent = useReportClickEvent();
 
   React.useEffect(() => {
-    if (!searchResults?.length || !searchId) {
+    if (!searchResults?.length || !searchId || searchResults === lastExecutedSearchResults) {
       return;
     }
     executeGenerativeDirectAnswer(searchActions);
-  }, [searchResults, searchId]);
+    setLastExecutedSearchResults(searchResults);
+  }, [searchResults, searchId, lastExecutedSearchResults]);
 
   if (!searchResults?.length || isLoading || !gdaResponse || gdaResponse.resultStatus !== 'SUCCESS') {
     return null;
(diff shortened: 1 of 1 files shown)


k-gerner (changes requested):
I wonder if we could use `src/hooks/useDebouncedFunction.ts` instead?

Also, do we want to update the package version in this change or does an automated GH workflow action do that?

mkouzel-yext:
> I wonder if we could use src/hooks/useDebouncedFunction.ts instead?

We _could_, as the delta between the searchResults object getting updated and the searchId object getting updated is likely minimal. However, explicitly processing just one of the updates strikes me as better practice. I think that my usage of "debounce" here is incorrect and misleading - I meant "deduplicate". 

We absolutely want to bump the patch version, good call. Updating.

src/components/GenerativeDirectAnswer.tsx
```diff
@@ -92,17 +92,19 @@ export function GenerativeDirectAnswer({
     }
   }, [isUniversal, universalResults, verticalResults]);
 
+  const [lastExecutedSearchResults, setLastExecutedSearchResults] = useState(searchResults);
```
k-gerner, line 95: it's possible i'm not interpreting this code correctly, but if we set the last executed results to searchResults initially, doesn't that mean the first search won't fire a GDA? Or does this get loaded before there are any results?

src/components/GenerativeDirectAnswer.tsx
```diff
@@ -92,17 +92,19 @@ export function GenerativeDirectAnswer({
...
 
   React.useEffect(() => {
-    if (!searchResults?.length || !searchId) {
+    if (!searchResults?.length || !searchId || searchResults === lastExecutedSearchResults) {
       return;
     }
     executeGenerativeDirectAnswer(searchActions);
-  }, [searchResults, searchId]);
+    setLastExecutedSearchResults(searchResults);
+  }, [searchResults, searchId, lastExecutedSearchResults]);
```
k-gerner, line 107: since we only update `lastExecutedSearchResults` inside the `useEffect`, do we need it in the dep array?

src/components/GenerativeDirectAnswer.tsx
k-gerner: Got it. I think we should set the initial state to null/undefined then, since that's the behavior we're expecting

src/components/GenerativeDirectAnswer.tsx
mkouzel-yext: My understanding of best practice is to include all external variables in the dependencies array (the alpha linter always throws an error for it). You're right that we only update it within the effect hook, though.

src/components/GenerativeDirectAnswer.tsx
mkouzel-yext: Good call, updating

src/components/GenerativeDirectAnswer.tsx
k-gerner: Sounds good

src/components/GenerativeDirectAnswer.tsx
```diff
@@ -92,17 +92,20 @@ export function GenerativeDirectAnswer({
...
```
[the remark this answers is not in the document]
k-gerner: image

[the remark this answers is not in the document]
mkouzel-yext: Interesting, `useRef` is definitely the way to go. Updated, thanks!

k-gerner (approved):

PR #566, opened 2025-10-28
v1.10.1

Fondryext:
feat: upgrade playwright version to avoid vuln
This PR ups the playwright version used to avoid a vuln

J=VULN-40423
TEST=compile


anguyen-yext2:
> > LGTM, but can you revert the version bump in `package.json` and `package-lock.json`? [thread](https://yext.slack.com/archives/GTTQTJH9B/p1760457504426429) for context. Sorry I haven't updated the wikis yet because we haven't completed the oidc epic
> 
> yeah will do! Does this mean we should at some point delete all the automated workflows that make the version bump prs as well?

ooooh good catch! I didn't notice that workflow because not all repos were using it and  i have never prefixed my branch name with `hotfix/` or `release/`. We should remove the `version_update.yml` workflow in the repos that contain `publish.yml` workflow then. If you don't mind can you do it for this one while you're at it? I'll create an item to do it for the `search-headless` and `search-headless-react`

PR #595, opened 2025-12-29
fix: upgrade storybook version

mkouzel-yext:
A vulnerability exists in storybook 8.5.0 that bundles the .env file in the static output during the storybook build. This vulnerability is fixed in version 8.6.15.


.github/workflows/run-tests.yml
```diff
@@ -11,6 +11,7 @@ jobs:
     with:
       build_script: |
         npx playwright install
+        npm run build
```
Fondryext: nice catch thanks

mkouzel-yext: ofc, np!

Fondryext (approved):

PR #610, opened 2026-01-13
chore: add linting configuration and hook

k-gerner:
Adds a linting configuration so that we will not auto-fail the corresponding test during PRs. Also fixes all the linting errors found by this new linting.

Additionally, adds a `.husky` configuration which will run the linter on every commit, and not let you commit if there is a lint failure.

Additionally, adjusts the dropdown behavior when `Tab` key is pressed, to better follow accessibility standards (mentioned in issue https://github.com/yext/search-ui-react/issues/611). Now the `Tab` button will exit the dropdown rather than cycling through the options.

J=[WAT-3497](https://yext.atlassian.net/browse/WAT-3497?atlOrigin=eyJpIjoiZDIwNjFlODhmYTMxNDQxZDg1NGEzYzRiYmE4ZGQwMGIiLCJwIjoiaiJ9)

### DIFF
--- a/src/components/sections/index.ts
+++ b/src/components/sections/index.ts
@@ -6,4 +6,4 @@ export type {
   StandardSectionProps
 } from './StandardSection';
 
-export { type SectionHeaderCssClasses } from './SectionHeader';
+export type { SectionHeaderCssClasses } from './SectionHeader';


src/components/sections/index.ts
```diff
@@ -6,4 +6,4 @@ export type {
   StandardSectionProps
 } from './StandardSection';
 
-export {type  SectionHeaderCssClasses } from './SectionHeader';
```
Fondryext: silly question and I know this was me - but should the `type` be outside the brackets?

src/components/AppliedFiltersDisplay.tsx
```diff
@@ -5,7 +5,7 @@ import { useClearFiltersCallback } from '../hooks/useClearFiltersCallback';
 import { FieldValueFilter, useSearchActions } from '@yext/search-headless-react';
 import { isDuplicateFieldValueFilter } from '../utils/filterutils';
 import { executeSearch } from '../utils';
-import React from 'react';
+import React, { useCallback, useMemo } from 'react';
```
Fondryext: should these all be in brackets?

Fondryext (dismissed):

src/components/sections/index.ts
k-gerner: It's equivalent, but I'll update it to `type {...` to be consistent 🙂

Fondryext (approved):

PR #640, opened 2026-02-27
fix: suppress i18next console support notice

chrissnyder2337:
Add showSupportNotice: false to i18next init config to prevent the Locize support notice from appearing in the console for all consumers of this library.

Closes #639


chrissnyder2337:
This should not be the approach that is taken. 

If i18next is updated to 26.0 or higher, the notice is removed.

See:

- https://github.com/i18next/i18next/blob/master/CHANGELOG.md#2600
- https://www.locize.com/blog/i18next-support-notice/
- https://github.com/i18next/i18next/issues/2385#issuecomment-4147565544

PR #665, opened 2026-04-22
fix: up node version in tests to 20

Fondryext:
We recently upped the node versions in the reusable workflows to 20, so do the same here


k-gerner (commented):
Looks like a new `npm i` is needed?

Fondryext:
> Looks like a new `npm i` is needed?

Yes, something is wrong. I have deleted and remade a couple times now. Still working on it

k-gerner (approved):

PR #682, opened 2026-06-24
chore: move the signpost next to GDA header and remove default label

anguyen-yext2:
Currently, the default GDA signpost seems redundant because the default GDA header is 'AI Generated Answer'. To address this, we're moving the signpost to the header row and removing the redundant label by default.

J=WAT-5664
TEST=manual,auto

https://github.com/user-attachments/assets/7f8707f2-0305-411a-8007-acfed988d2d1


mkouzel-yext:
nit: can we make the default button behind the signpost icon square when the label is empty?

anguyen-yext2:
> nit: can we make the default button behind the signpost icon square when the label is empty?

updated!

mkouzel-yext (approved):
LGTM!

PR #693, opened 2026-07-30
fix: give facet options search input an accessible name by default (#692)

### LINKED ISSUE #692: Accessibility bug: facet options-search input only gets an accessible name when the visible label is enabled (`showOptionsSearchInputLabel`)
## Summary
 
The options-search input rendered by `FilterGroup` (used by `StandardFacet`, `NumericalFacet`, and `StaticFilters`) only receives an associated `<label>` when `showOptionsSearchInputLabel` is enabled — but enabling it also displays the label visually. One prop controls both the semantic association and the visual presentation.
 
With the prop off (the default), the input's accessible name falls back to the shared placeholder `"Search here..."` — identical for every searchable facet on the page, and gone once the user types. The [HTML spec notes](https://html.spec.whatwg.org/multipage/input.html#the-placeholder-attribute) placeholders should not substitute for labels.
 
**Consumers who need a persistent, distinguishable accessible name without adding a visible label row to an existing design must compose two separate API surfaces manually (see workaround below).**
 
This is an enhancement request, not a claim that the component can't be configured accessibly. Is the coupling intentional, and would you consider providing programmatic naming independently of visible-label display?
 
## Affected Versions
 
- `@yext/search-ui-react`: 3.2.1 (verified against current `main`)
- `@yext/search-headless-react`: 2.7.1
## Root Cause
 
`FilterGroup.tsx` only generates the label text when the visible-display prop is on:
 
```tsx
const searchInputLabel = showOptionsSearchInputLabel
  ? t('filterGroupSearchInputLabel', { title })  // "Search {{title}} Options"
  : undefined;
```
 
`SearchInput.tsx` renders the associated `<label>` only when that text is provided, and has no other naming mechanism (no `aria-label` prop):
 
```tsx
{label && (
  <label id={labelId} htmlFor={inputId} className={labelClassName}>
    {label}
  </label>
)}
<input id={inputId} placeholder={placeholder ?? t('searchHere')} ... />
```
 
## Steps to Reproduce
 
1. Render a facet whose option count exceeds `showMoreLimit`, so the options-search input appears:
```tsx
   <StandardFacet fieldId="products" showMoreLimit={1} />
```
 
2. Inspect the input: no `<label>` exists; the accessible name is the generic placeholder.
3. Add `showOptionsSearchInputLabel`: an associated label ("Search Products Options") appears — but visibly.
No configuration produces the associated label without the visible display, other than the CSS workaround below.
 
## Current Workaround
 
```tsx
<StandardFacet
  fieldId="products"
  showMoreLimit={1}
  showOptionsSearchInputLabel
  customCssClasses={{ searchInputLabel: 'sr-only' }}
/>
```
 
This works — the label stays in the DOM, associated via `htmlFor`, available to assistive technology, and visually hidden. But it requires consumers to know that semantic naming and visual presentation must be composed manually.
 
### Workaround unavailable with auto-rendered `<Facets>`
 
When `<Facets>` places facets automatically, the workaround is not possible. `FacetsProps` does not expose `showOptionsSearchInputLabel`, and auto-rendered facets are built in `Facets.tsx` with only `fieldId`, `label`, and merged `customCssClasses` — so the label can never be enabled and there is nothing for a visually-hidden class to act on. (`FacetsCssClasses` extends `FilterGroupCssClasses`, so `searchInputLabel` styling *can* be passed globally, but there is no way to enable the label globally.)
 
The only recourse is to declare a child override per facet:
 
```tsx
<Facets>
  <StandardFacet fieldId="products" showOptionsSearchInputLabel customCssClasses={{ searchInputLabel: 'sr-only' }} />
  {/* ...one per searchable facet */}
</Facets>
```
 
which requires knowing every searchable facet's `fieldId` ahead of time and defeats the purpose of automatic rendering. Proposed solution 1 below would resolve this case automatically.
 
## Proposed Solutions
 
Possible directions rather than a prescribed implementation:
 
1. **Visually-hidden label by default.** Always generate the label text; use `showOptionsSearchInputLabel` only to control visibility (e.g., render with `sr-only` when false). Backward-compatible with the existing prop, and matches the [W3C WAI hidden-label pattern](https://www.w3.org/WAI/tutorials/forms/labels/#hiding-the-label-element). Also fixes auto-rendered `<Facets>`.
2. **Dedicated accessible-name prop**, e.g. an `aria-label` passthrough on `SearchInput`.
3. **Expose `showOptionsSearchInputLabel` on `FacetsProps`** so auto-rendered facets can opt in globally.
4. **Document the workaround** above as the supported pattern for nonvisual naming.

## Related Issues/PRs

- #638 introduced `showOptionsSearchInputLabel`
- #573, #611, #623 addressed separate facet/dropdown accessibility issues

chrissnyder2337:
Fixes #692.

## Problem

The options-search input rendered by `FilterGroup` (used by `StandardFacet`, `NumericalFacet`, and `StaticFilters`) only received an associated `<label>` when `showOptionsSearchInputLabel` was enabled, and enabling it also displayed the label visually. One prop controlled both the semantic association and the visual presentation.

With the prop off (the default), the input's accessible name fell back to the shared `"Search here..."` placeholder: identical for every searchable facet on the page, and gone once the user typed.

## Approach

This implements proposed solution 1 from the issue, plus solution 3.

`FilterGroup` now always generates the label text and always associates it with the input via `htmlFor`/`id`. `showOptionsSearchInputLabel` only controls whether the label is visible, through a new `visuallyHiddenLabel` prop on `Filters.SearchInput` that renders it with `sr-only`.

Every searchable facet therefore gets a persistent, distinguishable accessible name ("Search Products Options", "Search Price Options") with no visible label row and no consumer configuration. This also resolves the auto-rendered `<Facets>` case from the issue, where the CSS workaround was not available at all.

## Changes

- `Filters.SearchInput`: new `visuallyHiddenLabel` prop. When set, the label renders with `sr-only`, or with `labelClassName` if the consumer supplies one, so consumers whose stylesheet does not define an unprefixed `sr-only` can hide it another way.
- `FilterGroup`: label text is always generated; `showOptionsSearchInputLabel` maps to `visuallyHiddenLabel={!showOptionsSearchInputLabel}`.
- `FacetsProps.showOptionsSearchInputLabel` (solution 3): shows the label across auto-rendered facets, overridable per child facet. Facet props passed as `undefined` now count as unspecified, so they inherit the `Facets` value rather than erasing it.
- `SearchI18nextProvider`: added `filterGroupSearchInputLabel` to the `translationKeys` union. The key already existed in all 31 locale files, but overriding it through `translationOverrides` was a type error.

## Layout

The hidden label renders *after* the input. `CollapsibleSection` defaults to `space-y-3`, whose compiled rule is `.space-y-3>:not([hidden])~:not([hidden]){margin-top:.75rem}`. A label in front of the input would make the input a non-first child and give it a 0.75rem top margin it did not previously have, even though an `sr-only` label (`position: absolute`) contributes no height. Keeping the input first leaves the default rendering pixel-identical to v3.2.1. `htmlFor` preserves the association regardless of order, and a test pins the sibling order.

## Backwards compatibility

No breaking changes. `showOptionsSearchInputLabel` still defaults to `false` and still shows the label when `true`. The workaround documented in the issue (`showOptionsSearchInputLabel` plus `customCssClasses={{ searchInputLabel: 'sr-only' }}`) keeps working.


k-gerner:
Addressed in https://github.com/yext/search-ui-react/pull/699

